PR #28, opened 2025-08-20
HotFix for TPS: Removed SSE Path, Made Stream-only for Accuracy.

aceangel3k:
I removed the SSE path of the token per second calculation. It doesn't seem to show up even though Cerebras has support. Made it also count the stream-only.


mpfaffenberger:
hey really quick - shouldn't there be some deletions here? Didn't you move some code around in this p/r? If not, can you make sure there aren't two versions of things?

mpfaffenberger:
Ohhhh I see you got merge conflicts

aceangel3k:
Ahh, the codebase has changed considerably. Probably don't need this anymore right?

PR #31, opened 2025-08-25
Versioning for 0.0.122

aceangel3k:
I added git like file versioning (response/file undo/redo) to the TUI and in interactive mode. You can merge it or get ideas off of it if you were working on something similar.

Thanks,
Angel


aceangel3k:
Hah. The reason why there's some extra stuff is because I was using this version for some of my projects. It's very handy being able to revert and compare to certain revisions without commiting to git (like ac

PR #39, opened 2025-09-28
Handle missing model configs with friendly fallback path

diegonix:
## Summary
- add a helper that picks a safe fallback model whenever the configured model name is missing
- emit clear console guidance listing available models instead of crashing
- surface a final actionable error when no model can be loaded at all

## Testing
- manual: pointed the agent at an invalid model name and observed fallback logging + graceful recovery


mpfaffenberger (approved):
Approved

PR #45, opened 2025-10-05
feat: Adding edit file permission feature

cgycorey:


cgycorey:
I have done a few more changes, mainly around on the prompts for rejects, and tried to make the logic more self-contained under the plugin framework.

PR #52, opened 2025-10-15
TUI: Interactive /model picker + border fix

jackdevs000:
Implements a /model command that opens a modal window, allowing users to select from a list of live models directly in the interface.

Replace emoji icons with single width glyphs use "▶" and "■" instead of "▶️" and "⏹️".
Switch the input field border from “round” to “solid”.
These two changes eliminate width/rounding artifacts that caused the jagged edges.

New /model Selector Modal
image

Before (Old Jagged Edges issue)
image

After (Fixed Edges)
image


mpfaffenberger (approved):
LGTM

PR #54, opened 2025-10-16
Feat/history

RonStew2019:
Adds /history command to display message history within current autosave session

### **Usage:**
```bash
/history [N] [-v|--verbose]
```

### **Features:**
- **Default**: Shows last 10 messages with truncation for long content
- **Line count**: `N` limits displayed messages (e.g., `/history 5`)
- **Verbose mode**: `-v` displays full content without truncation

## Normal:
image

## Verbose:
image

### DIFF
--- a/code_puppy/command_line/command_handler.py
+++ b/code_puppy/command_line/command_handler.py
@@ -787,6 +787,8 @@ def handle_command(command: str):
         )
         return True
 
+
+
     if command.startswith("/history"):
         from code_puppy.command_line.history_command import handle_history_command
         return handle_history_command(command)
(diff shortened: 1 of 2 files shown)


code_puppy/command_line/command_handler.py
```diff
@@ -786,6 +787,472 @@ def handle_command(command: str):
         )
         return True
 
+    if command.startswith("/history"):
```
mpfaffenberger: Hey Ronan - do you think you could refactor this nested functionality into its own thing, and then make this change in `command_handler.py` a 2-liner?

PR #56, opened 2025-10-17
Fix: Improved Large File Handling in `edit_file` Tool

cdakotabrown:
### Problem
The `edit_file` tool was failing when creating files with large amounts of content (typically >5,000 lines), producing unhelpful error messages that left users confused about what went wrong.

### Solution
This PR adds proper size validation and helpful guidance for working with large files:

- **Clear size limits**: Recommended max of 100KB, hard limit of 500KB
- **Helpful warnings**: Users get notified when approaching size limits
- **Better error messages**: Clear, actionable guidance instead of cryptic errors
- **Improved reliability**: Chunked writing for better handling of large content

### What Changed
- Added size validation with helpful error messages
- Implemented chunked writing for files over 50KB
- Updated documentation with size guidelines and best practices
- Enhanced error handling to catch issues early

### User Impact
✅ **Before**: Cryptic errors when files were too large  
✅ **After**: Clear guidance on size limits and how to work with large content

Users working with large files now get helpful suggestions to split content into smaller, more manageable files—which aligns with our code quality guidelines anyway! 🐶

### Compaction Safety ✅
**Does this break compactions?** No! This change is completely safe:

- **Compactions work in-memory**: The message history compaction system (`message_history_processor`, `summarize_messages`) processes messages as in-memory `List[ModelMessage]` objects and never writes to files
- **No file I/O in compactions**: The summarization agent uses `agent.run()` to generate text summaries, not file operations
- **Size limits only affect `edit_file` tool**: These limits only apply when the agent explicitly uses the `edit_file` tool to create/modify files on disk
- **No internal systems affected**: Session storage, message processing, and all other internal systems don't use `_write_to_file()`

The 500KB hard limit (≈15,000 lines of code) and 100KB recommended limit (≈3,000 lines) are reasonable for generated code files and align with our <600 lines per file guideline.

### DIFF
--- a/code_puppy/tools/file_modifications.py
+++ b/code_puppy/tools/file_modifications.py
@@ -6,6 +6,12 @@ Key guarantees
 2. **Full traceback logging** for unexpected errors via `_log_error`.
 3. Helper functions stay print-free and return a `diff` key, while agent-tool wrappers handle
    all console output.
+
+Size Limits
+-----------
+- **Recommended max content size**: 100KB (~3000-5000 lines)
+- **Practical limit**: 500KB (hard limit for XML parameter encoding)
+- Files larger than recommended size should be split or created via shell commands
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


mpfaffenberger (changes requested):
Hey man - unfortunately this failure occurs before the edit_file tool ever sees the payload. It's a server side failure :(... 

One thing that could potentially help is kind if similar to what you've done - the modifications to the docstring. Although 3000-5000 lines is WAY too much. I would suggest we edit the docstring (which automatically goes into the system prompt) to suggest no more than 400-ish lines at once. 

Models may still choose to ignore this. 

It's more of a problem in Vertex AI than it is for other providers. Again, the error is happening server side. 

It's when the output token count exceeds the max_output. We cannot turn up max_output arbitrarily either because that could cause compaction and other context errors.

PR #61, opened 2025-10-21
fix: attachment parser crash from visual dividers and improve comment detection

diegonix:
Fix attachment parser crash from visual dividers and improve comment detection

Prevent OSError: File name too long crashes when CLI output containing screen drawing characters is misinterpreted as file paths. The parser was incorrectly processing sequences like "/"
followed by repeated Unicode box drawing characters as valid filesystem paths.

Key changes:
- Add comprehensive Unicode box drawing character detection: ─ │ ┌ ┐ └ ┘ ├ ┤ ┬ ┴ ┼ ═ ║ ╚ ╝ ╔ ╗ ╠ ╣ ╦ ╩ ╬ ━ ┃ ┏ ┓ ┗ ┛ ┣ ┫ ┳ ┻ ╋
- Add line separator and decorative characters: - = — – • · ○ ● ■ □ ░ ▒ ▓ █ ▄ ▌ ▐ ▀ ╭ ╮ ╯ ╰
- Implement 50% threshold rule: exclude tokens where >50% of non-slash characters are dividers
- Enhance comment detection with whitespace support for #, #, //, /*, ;, --, !, %, rem, REM, Note:
- Add OSError protection around unhandled path.exists() call at line 270
- Preserve all valid path functionality including Unicode characters in filenames

This fix prevents crashes from CLI output containing visual dividers while maintaining full compatibility with legitimate file paths that may contain Unicode characters.

### DIFF
--- a/code_puppy/command_line/attachments.py
+++ b/code_puppy/command_line/attachments.py
@@ -4,8 +4,10 @@ from __future__ import annotations
 
 import mimetypes
 import os
+import re
 import shlex
 from dataclasses import dataclass
+from functools import lru_cache
 from pathlib import Path
 from typing import Iterable, List, Sequence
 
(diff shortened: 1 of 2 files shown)


code_puppy/command_line/attachments.py
```diff
@@ -68,8 +68,83 @@ def _is_probable_path(token: str) -> bool:
     # Reject absurdly long tokens before any processing to avoid OS errors
     if len(token) > MAX_PATH_LENGTH:
         return False
-    if token.startswith("#"):
+
+    # Reject screen drawing/divider characters
+    divider_chars = {
```
mpfaffenberger: There has to be a better way to do this.

diegonix: Thanks for the review! I agree the long hard‑coded list of “divider” characters wasn’t ideal. I’ve refactored it to a small, maintainable heuristic:
* Detect visual separators by **Unicode block ranges** (Box Drawing U+2500–U257F, Block Elements U+2580–U259F, Geometric Shapes U+25A0–U25FF) plus a short extra set for common CLI/Markdown dividers, and keep the **>50% ratio** rule.
* Treat long runs like `====`/`----` as dividers via a tiny regex.
* Consolidate comment detection into one regex (`^\s*(#|//|/\*|;|--|!|%)`, case‑insensitive).
* Fix the `OSError` guard so the **only** `exists()/is_file()` check lives inside a `try/except`, preventing ENAMETOOLONG crashes.
This removes the giant char set, improves readability, and broadens coverage without risking false positives on legitimate Unicode filenames. I’ve also added tests covering these cases. Happy to tweak thresholds if you prefer.
(Unicode references for the blocks are in the PR description.)

PR #69, opened 2025-10-26
browser: defer Camoufox imports and add Playwright fallback to avoid browserforge downloads at import-time

IgriegaL:


mpfaffenberger (approved):
LGTM - thank you for doing this!

PR #82, opened 2025-11-02
TUI Changes

janfeddersen-wq:


code_puppy/config.py
```diff
@@ -998,6 +998,126 @@ def finalize_autosave_session() -> str:
...
+
+def load_api_keys_to_environment():
+    """Load all API keys from .env and puppy.cfg into environment variables.
+
+    Priority order:
+    1. .env file (highest priority) - if present in current directory
+    2. puppy.cfg - fallback if not in .env
+    3. Existing environment variables - preserved if already set
+
+    This should be called on startup to ensure API keys are available.
```
mpfaffenberger, line 1082: I think this should also get called whenever a model is loaded

PR #87, opened 2025-11-03
small change to remove hanging

cgycorey:
Just want to fix the hanging terminal when user press ctrl + c during file permission approval.


mpfaffenberger (commented):
Will test this today or tomorrow, sorry for the delay

PR #97, opened 2025-11-07
Fix UnicodeEncodeError on Windows when printing emojis

gs131016:


mpfaffenberger:
Tested this branch on windows and I'm still getting the unicode error

gs131016:
Intersting. It works fine for me after this fix.

gs131016:
So, bascially I updated the puppy version (0.0.270) today and tried to execute **code-puppy -i**, which worked fine. When I tried **code-puppy -w**, I received the same error again.
image
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f436' in position 0: character maps to <undefined>.

The fix suggested in this PR didn't work for me either for this update. It was working fine for that last 2 updates.
On further analyzing this, the issue was resolved after I executed the below powershell command to update the registry settings and without making any changes to main.py
            [Environment]::SetEnvironmentVariable("PYTHONUTF8", "1", "User")

janfeddersen-wq:
```
import sys
import os
import io

# Windows Unicode/Emoji fix
if sys.platform == "win32":
    os.environ["PYTHONUTF8"] = "1"
    try:
        import ctypes
        kernel32 = ctypes.windll.kernel32
        kernel32.SetConsoleOutputCP(65001)
        kernel32.SetConsoleCP(65001)
    except Exception:
        pass
    if hasattr(sys.stdout, 'buffer') and not isinstance(sys.stdout, io.TextIOWrapper):
        sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8', errors='replace', line_buffering=True)
    if hasattr(sys.stderr, 'buffer') and not isinstance(sys.stderr, io.TextIOWrapper):
        sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8', errors='replace', line_buffering=True)
        
``` 

should be doing this ... the prob the env var alone doesnt change the CP on the the thing ... this isnt a good way and hacky :D but i had to do similar because windows is STUPIDLY setting the shell type depending on CMD / Powershell / OS version ...  and initializing the ConsoleCP 


The above provided shell var works fine in WSL but not clean windwos

Issue #109, opened 2025-11-16
Feature request: Add shell mode


cgycorey:
Add a shell mode to puppy. Just like major coding agent CLI. Instead of going to another terminal to execute a trivial shell command that you don't need LLM to execute, you instead switch on this mode and execute it in our cli's input. And this input will not send to LLM but just to execute the shell command directly.

PR #118, opened 2025-11-23
feature/shell

jasimpson-yobo:
Added ability to pause code puppy to use shell commands 

/shell to switch to a shell 
feature_shell


mpfaffenberger:
@jasimpson-yobo ty for the contrib - please make sure you run the linters `ruff check --fix` and `ruff format .`, you can install `lefthook` to ensure these are run when you commit.

mpfaffenberger:
closes #109

PR #181, opened 2026-02-16
feat: add direct session resume by name

ccrios83:
## Summary

Allow users to resume a session directly by name instead of always opening the interactive picker:

```bash
/resume                              # Opens picker (unchanged)
/resume auto_session_20260214_175543 # Loads directly (NEW)
```

## Problem

Users kept getting told they could use `/resume <session_name>` but it didn't actually work - it just opened the picker regardless.

## Changes

| File | Changes |
|------|---------|
| `session_commands.py` | Parse optional session_name argument, return direct load marker |
| `cli_runner.py` | Handle `__AUTOSAVE_LOAD_DIRECT__` marker with session lookup |

## User Experience

**Success:**
```
✅ Resumed: auto_session_20260214_175543
📊 47 messages (12,345 tokens)
```

**Error:**
```
❌ Session not found: invalid_session_name
💡 Use /resume to open the session picker.
```

## Testing

- All existing session tests pass ✅
- Manual testing confirms both picker and direct load work

---

*Ported from Walmart internal implementation*

## Summary by CodeRabbit

* **Improvements**
  * Session loading now supports direct load by session name in addition to interactive selection.
  * Enhanced error handling for session loading with helpful error messages and recovery guidance.

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -653,8 +653,14 @@ async def interactive_mode(message_renderer, initial_command: str = None) -> Non
                         from code_puppy.messaging import emit_error, emit_success
                         from code_puppy.session_storage import load_session
 
-                        base_dir = Path(AUTOSAVE_DIR)
-                        session_path = base_dir / f"{session_name}.pkl"
+                        base_dir = Path(AUTOSAVE_DIR).resolve()
+                        session_path = (base_dir / f"{session_name}.pkl").resolve()
+
+                        # Security: Prevent path traversal attacks
+                        if base_dir not in session_path.parents and session_path.parent != base_dir:
+                            emit_error(f"❌ Invalid session name: {session_name}")
+                            emit_info("💡 Use /resume to open the session picker.")
+                            continue
 
                         # Check if session exists
                         if not session_path.exists():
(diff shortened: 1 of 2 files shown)


nhicks00 (changes requested):
## Review: Request Changes

Thanks for adding direct session resume! Great feature. Found a few issues that need addressing:

### 🐛 Critical: `emit_info` Not Imported

In `cli_runner.py`, the direct load error path uses `emit_info` but only imports `emit_error` and `emit_success`:

```python
from code_puppy.messaging import emit_error, emit_success  # ❌ Missing emit_info!

emit_error(f"❌ Invalid session name: {session_name}")
emit_info("💡 Use /resume to open the session picker.")  # 💥 NameError!
```

This will crash when a session is not found or path traversal is detected.

**Fix:** Add `emit_info` to the import statement.

---

### ⚠️ UX Inconsistency: Missing `display_resumed_history`

The interactive picker path calls `display_resumed_history(history)` after loading, but the direct load path skips this. Users get different experiences depending on how they resume.

**Suggestion:** Add the same `display_resumed_history(history)` call to the direct load path for consistency.

---

### 💡 Nice to Have: DRY Refactor

The session loading logic is duplicated between both paths:
- `load_session()`
- `set_message_history()`
- `set_current_autosave_from_session_name()`
- Token calculation
- Success message

Consider extracting to a shared helper like `_load_and_resume_session(session_name)`.

---

### ✅ What Looks Good

- Path traversal protection 👍
- Backward compatibility (no args = picker)
- Updated help text and docstrings
- All tests pass

Let me know if you have questions!

PR #193, opened 2026-02-20
fix: handle DBOS cancel/shutdown race condition gracefully

mpfaffenberger:
When DBOS tries to persist state via asyncio.to_thread() after the event loop's default ThreadPoolExecutor has been shut down, a RuntimeError is raised with 'cannot schedule new futures after shutdown'. This adds a dedicated except* RuntimeError handler that catches this specific case and logs an info message instead of crashing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during agent shutdown to manage executor race conditions more gracefully and prevent workflow cancellation.

* **Chores**
  * Added infrastructure support for synthetic status plugin functionality.

### DIFF
--- a/code_puppy/agents/base_agent.py
+++ b/code_puppy/agents/base_agent.py
@@ -1957,21 +1957,36 @@ class BaseAgent(ABC):
                 if get_use_dbos():
                     await DBOS.cancel_workflow_async(group_id)
             except* RuntimeError as rt_error:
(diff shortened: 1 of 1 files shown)


wkramme:
@mpfaffenberger I have a PR (195) to fix what the rabbit didn't like in my original commit

PR #201, opened 2026-02-24
Perf/reduce redundant syscalls

SugarFreeDoNoSo:
It was taking longer than it had any right to, so I took a look under the hood. This trims redundant filesystem syscalls (scandir/stat batching), caches ripgrep lookup, and removes a few avoidable checks.

## Summary by CodeRabbit

* **Refactor**
  * Faster, more efficient directory and file scanning and deduplication for snappier browsing and searches.
  * Centralized and cached external search-tool lookup to speed repeated search operations.
* **Bug Fixes / Reliability**
  * More robust per-entry error handling to reduce failures during listing and reading.
  * Ensures config directories are created with secure permissions when needed.
  * Log rotation now behaves consistently even when log files are missing.

### DIFF
--- a/code_puppy/command_line/utils.py
+++ b/code_puppy/command_line/utils.py
@@ -5,21 +5,16 @@ from rich.table import Table
 
 
 def list_directory(path: str = None) -> Tuple[List[str], List[str]]:
-    """
-    Returns (dirs, files) for the specified path, splitting out directories and files.
-    """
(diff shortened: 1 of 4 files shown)


SugarFreeDoNoSo:
Not critical, just a micro-optimization. With multiple agents running in parallel on the same filesystem, contention was causing the same ops to slow down dramatically, so I tidied it up.

Issue #205, opened 2026-02-25
fix: /cd command does not update LLM system prompt or project AGENT.md rules


nhicks00:
### User-reported symptom
After running `/cd <new-directory>`, the working directory changes in the terminal (the prompt display updates), but the **path sent in the LLM system prompt remains stale** — the model still believes it is in the original directory.

A secondary symptom: project-local `AGENT.md` rules from the new directory are not picked up after `/cd`.

### Root cause: two un-invalidated caches

**Bug 1 — Stale LLM instructions (`_code_generation_agent`)**

`reload_code_generation_agent()` bakes the system prompt (which in some deployments includes `os.getcwd()`) into a `PydanticAgent(instructions=...)` instance stored as `self._code_generation_agent`. In `run_with_mcp()` the agent is reused if it already exists, so the instructions never refresh after `/cd`.

**Bug 2 — Stale project AGENT.md rules (`_puppy_rules` cache)**

`load_puppy_rules()` uses a one-shot cache:
```python
if self._puppy_rules is not None:
    return self._puppy_rules  # never re-reads after first load
```
After `/cd` to a new project, the old project's `AGENT.md` is forever stuck in cache; the new project's rules are silently ignored.

### What is NOT broken
- Terminal CWD display — uses `os.getcwd()` fresh on every render ✅
- File tools (`list_files`, `read_file`, etc.) — `os.path.abspath()` against live CWD ✅
- Shell command runner — subprocess inherits live CWD ✅
- Shell safety cache — keyed by `(command, cwd)`, different CWD = cache miss ✅
- Agent registry — `_discover_agents()` always clears and rebuilds ✅

### Fix

1. **`base_agent.py` — `reload_code_generation_agent()`**: Add `self._puppy_rules = None` at the top so every reload reads fresh rules from disk.

2. **`core_commands.py` — `handle_cd_command()`**: After `os.chdir(target)` succeeds, call `get_current_agent().reload_code_generation_agent()` to rebuild the `PydanticAgent` with an up-to-date system prompt and fresh project rules.

### Affected files
- `code_puppy/command_line/core_commands.py`
- `code_puppy/agents/base_agent.py`

PR #207, opened 2026-02-26
Fix/cd command invalidate agent cache

tsternweis:
When using `/cd` to change directories, the cached pydantic agent (`_code_generation_agent`) retains stale instructions containing the old `os.getcwd()` value. The model continues operating as if the user is in the previous directory, suggesting wrong file paths and misunderstanding project context.

**Root cause:** `handle_cd_command()` calls `os.chdir()` but never invalidates the cached agent. Other config-changing commands (`/model`, `/set`, `/reasoning`, `/verbosity`) already call `reload_code_generation_agent()` after changes, but `/cd` was missing this pattern.

**Affected scenarios:** `/resume` + `/cd` workflow, working across multiple projects in one session, any directory change mid-conversation.

### Fix (1 file, +8 lines)

**`/core_commands.py`**
After `os.chdir()` succeeds, sets `agent._code_generation_agent = None` to invalidate the cache. Triggers lazy rebuild on next user message via `run_with_mcp()` which checks: 
```python
self._code_generation_agent or self.reload_code_generation_agent()
```

**Fixes:** Agent using wrong working directory in system prompt after `/cd`, causing incorrect file path suggestions and project context confusion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures the code-generation context is refreshed when changing directories so generated code reflects the current workspace. This prevents stale context from affecting results and surfaces an error if the refresh fails, improving reliability when navigating between projects.

### DIFF
--- a/code_puppy/command_line/core_commands.py
+++ b/code_puppy/command_line/core_commands.py
@@ -80,9 +80,13 @@ def handle_cd_command(command: str) -> bool:
             #Invalidate cached agent so system prompt refreshed with new CWD
             from code_puppy.agents.agent_manager import get_current_agent
 
-            agent = get_current_agent()
-            if agent is not None:
-                agent._code_generation_agent = None
+            try:
+                agent = get_current_agent()
+                if agent is not None:
+                    agent._code_generation_agent = None
+            except Exception as e:
+                emit_error(f"Error invalidating agent cache: {e}")
+                pass
 
             emit_success(f"Changed directory to: {target}")
         else:


nhicks00:
Hey @Apollex-byte — thanks for jumping on this! 🙌

I also started a branch for this same fix (#205) — looks like we landed on the same approach, which is great. Mine just goes a bit further by also clearing the `_puppy_rules` cache (so project-local AGENT.md rules refresh too) and includes tests, so I'm going to open a PR from that branch. Going to close this one in favor of it, but appreciate the contribution!

nhicks00:
Closing in favor of the more comprehensive fix coming in from #205 branch. Thanks again @Apollex-byte!

PR #215, opened 2026-02-28
Feat/scheduler daily at

mattnico:
## What

Adds a new `daily_at` schedule type that fires a scheduled task at one or more
specific wall-clock times each day (e.g. `09:00`, or `09:00,17:30`).

## Why

The existing schedule types (`interval`, `hourly`, `daily`, `cron`) don't let
users say _"run this at 9am"_. The `daily_at` type fills that gap with a simple
`HH:MM` format — no cron syntax required.

### `code_puppy/scheduler/daemon.py`
- **`parse_daily_at_times(schedule_value)`** — parses a comma-separated string
  of `HH:MM` times into `(hour, minute)` tuples. Invalid entries are skipped
  with a warning rather than crashing the daemon.
- **`should_run_task()`** — new `daily_at` branch fires if any target time has
  passed today and `last_run` is before that target. Restart-safe: if the daemon
  was down at `09:00` and restarts at `09:15`, it still fires.

### `code_puppy/scheduler/config.py`
- Updated `schedule_type` field comment to document `daily_at` as a valid value.

### `code_puppy/plugins/scheduler/scheduler_wizard.py`
- Replaced the old `"Daily"` menu option with `"Daily at specific time(s)..."`.
- Collects time(s) via a `TextInputMenu`, validates each `HH:MM` entry, strips
  invalid ones with a warning, and aborts if none are valid.
- Summary display renders `"daily at 09:00,17:30"` instead of the raw internal
  type/value string.

### `tests/scheduler/test_daily_at.py` _(new, 31 tests)_
Covers `parse_daily_at_times()` and all branches of `should_run_task()` for
the `daily_at` type. All wall-clock comparisons use a pinned `NOW` constant so
tests are never flaky.

| Group | What's covered |
|---|---|
| Parser | Single/multiple times, whitespace, midnight, 23:59, invalid formats, empty string, warns on bad input |
| Basic fire/no-fire | Never-run fires after target; future target doesn't fire; fires exactly at boundary; doesn't fire 1s before |
| `last_run` logic | Already ran today → no re-fire; ran yesterday → fires; restart-safe catch-up; ran before target same day → fires |
| Multiple times | First due + second future; first serviced + second future; second becomes due; all future; both past |
| Edge cases | Disabled task; all-invalid schedule; mix valid/invalid; midnight target |

### `tests/plugins/test_scheduler_wizard.py` _(6 new tests + 1 fix)_

| Test | What it guards |
|---|---|
| `test_daily_at_single_time` | Happy path — one `HH:MM` → `schedule_type="daily_at"` |
| `test_daily_at_multiple_times` | Comma list is preserved verbatim in `schedule_value` |
| `test_daily_at_cancel_time_input` | `None` from `TextInputMenu` → wizard returns `None` |
| `test_daily_at_all_invalid_times` | All-bad input → `None` |
| `test_daily_at_strips_invalid_times` | Bad entries dropped, valid ones kept |
| `test_daily_at_summary_display` | `"daily at HH:MM"` appears in stdout; raw `"daily_at"` string does not leak |
| `test_wizard_code_puppy_first` _(fix)_ | Was referencing `"Daily"` which was removed from `schedule_map`; updated to `"Every hour"` |

## Summary by CodeRabbit

* **New Features**
  * Added "Daily at specific time(s)..." scheduling to enter one or more validated HH:MM times.

* **Enhancements**
  * Renamed display option to "Daily (every 24h)"; summaries now show "daily at HH:MM" and scheduling better handles missed daily runs.

* **Documentation**
  * Config docs updated to list "daily_at" as a valid schedule type.

* **Tests**
  * Expanded unit and wizard tests for parsing, scheduling behavior, summaries, validation, and edge cases.

### DIFF
--- a/code_puppy/plugins/scheduler/scheduler_wizard.py
+++ b/code_puppy/plugins/scheduler/scheduler_wizard.py
@@ -246,6 +246,7 @@ def create_task_wizard() -> Optional[dict]:
 
     if schedule_choice == "Daily at specific time(s)...":
         print("\n  Enter one or more 24-hour times separated by commas.")
+        print("  (24-hour clock: 1:00 PM = 13:00, 5:30 PM = 17:30, midnight = 00:00)")
         print("  Examples: 09:00   or   09:00,13:00,17:30\n")
         time_input = TextInputMenu(
             "Time(s) (HH:MM, comma-separated)", default="09:00"
--- a/code_puppy/plugins/scheduler/scheduler_wizard.py
+++ b/code_puppy/plugins/scheduler/scheduler_wizard.py
@@ -217,6 +217,7 @@ def create_task_wizard() -> Optional[dict]:
             "Every hour",
             "Every 2 hours",
             "Every 6 hours",
+            "Daily (every 24h)",
             "Daily at specific time(s)...",
             "Custom interval...",
         ],
@@ -226,6 +227,7 @@ def create_task_wizard() -> Optional[dict]:
             "Run once per hour",
             "Run 12 times per day",
             "Run 4 times per day",
+            "Run once per day, 24h after last run",
             "Run at exact wall-clock times, e.g. 09:00 or 09:00,17:30",
             "Specify custom interval like 45m, 3h, 2d",
         ],
@@ -242,6 +244,7 @@ def create_task_wizard() -> Optional[dict]:
         "Every hour": ("hourly", "1h"),
         "Every 2 hours": ("interval", "2h"),
         "Every 6 hours": ("interval", "6h"),
+        "Daily (every 24h)": ("daily", "24h"),
     }
 
     if schedule_choice == "Daily at specific time(s)...":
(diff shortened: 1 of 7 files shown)


nhicks00 (changes requested):
- The old "Daily" option (run every 24h) got removed from the wizard — was that intentional? Existing daily tasks still work in the daemon but users can't create new ones through the wizard anymore. Should both
  options coexist, or is this a deliberate replacement?
  - daily_at introduces wall-clock time semantics into a scheduler that's entirely timezone-naive (datetime.now() without tz=). For intervals that's fine, but "run at 09:00" is inherently timezone-dependent. Have
  you thought about at least dropping a code comment on the daily_at branch in should_run_task() acknowledging the naive-datetime limitation? Not asking you to solve TZ support here, just document it.



  - parse_daily_at_times docstring says "Invalid entries are silently skipped" but the implementation prints a [Scheduler] Warning: — update the docstring to match. (CodeRabbit flagged this too, +1 on their inline
  comment.)
  - Normalize times with strftime("%H:%M") after parsing and deduplicate — 9:00,09:00 would store duplicates in the config as-is. Won't cause double-firing but it's messy in persisted state. (Same as CodeRabbit's
  suggestion on the wizard.)



  - Both CI failures (ruff format + pytest) are pre-existing on main — unrelated files, not introduced by this PR.

Issue #213, opened 2026-02-27
Allow disabling resume history display and clear old history on re-resume


nhicks00:
Currently the resume history display (shown on `/resume`, `/autosave_load`, `/load_context`) has two limitations:

1. The minimum `resume_message_count` is 1 — there's no way to disable it entirely
2. If you resume multiple sessions in a row, old history stays on screen and new history stacks below it

This change:
- Allows `resume_message_count=0` to disable the display
- Automatically clears previously displayed history when resuming a new session (using ANSI cursor save/restore)

PR #235, opened 2026-03-14
fix: allow resume_message_count=0 to disable resume history display

RyanLHicks:
get_resume_message_count() enforced a minimum of 1, making it impossible to fully disable the resume history display.

Changed the lower bound from 1 to 0 so users can set resume_message_count=0` to suppress the display entirely, as requested in #213.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted resume message count minimum to allow 0 (disabled) while keeping the 100 maximum.
  * Prevented rendering resumed history when the requested message count is zero or negative, avoiding unnecessary display.


mpfaffenberger (approved):
Seems fine.

PR #240, opened 2026-03-16
fix: /cd command now reloads agent context including AGENTS.md

wkramme:
## Problem
The `/cd` command changes the working directory but the agent context was not being refreshed, causing:
- AGENTS.md files in new directory to be ignored  
- Stale working directory in system prompts
- Plugin callbacks not re-firing with new context

## Root Causes
1. **Cached `_puppy_rules`** - AGENTS.md content is cached in `base_agent.py` and never re-read
2. **No agent reload on `/cd`** - Directory changes but agent context stays stale  
3. **Plugin callbacks don't re-fire** - Any prompt additions based on cwd become stale

## Changes
- Modified `/cd` command to clear `_puppy_rules` cache and reload agent
- Added graceful error handling (directory change succeeds even if reload fails)
- Added two comprehensive tests to verify reload behavior

## Impact
When users `/cd` to a new directory, the agent now:
- ✅ Re-reads AGENTS.md from new directory
- ✅ Updates working directory in system prompt  
- ✅ Re-fires all plugin callbacks (`on_load_prompt`)
- ✅ Refreshes all context-dependent prompt components

## Testing
All 7 `/cd` command tests pass, including 2 new tests:
- `test_cd_reloads_agent_context`
- `test_cd_handles_agent_reload_failure_gracefully`

## Summary by CodeRabbit

* **Bug Fixes**
  * `/cd` now triggers an automatic agent context reload after changing directories.
  * Directory changes remain successful even if agent reload fails; a non-fatal error is emitted on failure and an informational message is shown on successful reload. Emission behavior changed from a prior warning to a non-fatal error on reload failure.

* **Tests**
  * Added tests verifying reload on `/cd` and graceful handling when reload raises an exception; updated expectations to match the new error emission behavior.

### DIFF
--- a/code_puppy/command_line/core_commands.py
+++ b/code_puppy/command_line/core_commands.py
@@ -85,9 +85,8 @@ def handle_cd_command(command: str) -> bool:
                 
                 current_agent = get_current_agent()
                 if current_agent:
-                    # Clear cached puppy rules so AGENTS.md is re-read from new directory
-                    current_agent._puppy_rules = None
-                    # Reload agent to rebuild system prompt with new working directory
+                    # reload_code_generation_agent() invalidates cached rules
+                    # and rebuilds prompt/context from the new cwd
                     current_agent.reload_code_generation_agent()
                     emit_info("Agent context updated for new directory")
             except Exception as e:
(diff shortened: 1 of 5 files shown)


wkramme:
Dupe with #291

PR #250, opened 2026-03-22
fix: ripgrep fallback for list_files

### LINKED ISSUE #36: [BUG] File listing doesn't work when ripgrep is not installed
Running MacOS Sequoia 15.7
Python 3.13.5
Code Puppy 0.0.170

If ripgrep is not installed through Homebrew, Code Puppy is not able to list any files.

Once I installed ripgrep through Homebrew, Code Puppy started working.

RyanLHicks:
Closes #36

Ripgrep fallback for list_files
When ripgrep is not installed, `_list_files` returned a hard error blocking all file operations. Changed to a graceful fallback to non-recursive `os.listdir` with a warning instead of failing.

## Release Notes

* **Bug Fixes**
  * File listing now gracefully falls back to non-recursive mode when ripgrep is unavailable, instead of failing.
  * Token estimation for file operations now uses improved heuristics for more accurate counting.

* **Tests**
  * Added regression tests for file listing fallback behavior.
  * Updated existing tests to reflect improved error handling.


mpfaffenberger:
Noting that we should have already merged the token estimation changes - thanks for that contribution!

Can you explain what are the conditions where `rg` doesn't get installed? `rg` is a _hard_ dependency, it's specified in our `pyproject.toml`.

RyanLHicks:
No problem, glad to help! Yes that's my bad I forgot update the title from this PR, but everything is updated now for the ripgrep. 

This addresses Issue #36. The specific scenario: a user running pyenv with a named virtualenv where rg is installed inside the env but pyenv shims can't resolve it via shutil. which, the venv fallback could also miss it depending on how pyenv sets sys.executable. The user had to install ripgrep separately to unblock themselves. This change means file listing degrades to non-recursive instead of returning a hard error in that situation. 

If it's one off issue it's probably fine, but something I just noticed from someone else. If it's not a serious issue in your eyes, I'll close this PR.

PR #253, opened 2026-04-01
feat(skills): auto-activate skills by fuzzy-matching user prompt

Prithvi1994:
## Problem

Skills in Code Puppy are injected into the system prompt as an XML list of names + descriptions, with guidance text telling the LLM to call `activate_skill()` when a match is detected.

In practice, skills rarely activate unless the user explicitly says "skill" in their prompt. The LLM doesn't proactively match task intent against skill descriptions — it needs a very direct signal.

This is a real usability gap: a user asking *"help me create a pull request"* won't benefit from an installed `github-pr-workflow` skill, even though it's a perfect match.

## Root Cause

Agents are activated **deterministically** — `invoke_agent` is an explicit tool call. Skills rely entirely on **LLM discretion** — passive XML in the prompt with a vague instruction to "match". There's no pre-run matching layer.

## Solution

This PR adds a new plugin `auto_skill_activator` that hooks into `get_model_system_prompt` and:

1. Scores every installed skill's `name + description + tags` against the user's prompt using `rapidfuzz.token_set_ratio` (already a project dependency)
2. Skills scoring **≥ 65** (configurable via `AUTO_ACTIVATE_THRESHOLD`) have their full `SKILL.md` content auto-injected into the system prompt **before the agent runs**
3. Caps at **3 skills max** (`MAX_AUTO_ACTIVATE`) to protect context window
4. Always sets `handled=False` so `claude-code`, antigravity, and other model handlers still fire
5. Degrades gracefully on any error — never crashes the agent run

## Why This Approach

- **Pure plugin** — zero changes to core (`base_agent.py`, `model_utils.py`, `callbacks.py`)
- **Follows AGENTS.md** — uses existing callback architecture exactly as documented
- **Uses existing dependency** — `rapidfuzz` is already in `pyproject.toml`
- **Non-breaking** — users with no skills installed see zero behavior change

## Files Changed

| File | Purpose |
|------|---------|
| `code_puppy/plugins/auto_skill_activator/__init__.py` | Plugin package |
| `code_puppy/plugins/auto_skill_activator/register_callbacks.py` | Core logic (158 lines) |
| `tests/plugins/test_auto_skill_activator.py` | 16 test cases covering activation, ranking, caps, error resilience |

## Test Coverage

- ✅ Matching skill is injected with full content
- ✅ Unrelated prompt returns None (no-op)  
- ✅ Empty/whitespace prompt returns None
- ✅ Globally disabled skills respected
- ✅ Per-skill disabled list respected
- ✅ Skills without SKILL.md skipped
- ✅ Multiple skills ranked by score, capped at MAX_AUTO_ACTIVATE
- ✅ `handled=False` always set
- ✅ User prompt preserved in result
- ✅ Exception in discovery returns None gracefully

## Summary by CodeRabbit

* **New Features**
  * Auto Skill Activator: scores discovered skills against the user prompt and auto-injects the most relevant SKILL.md content into the system prompt, honoring disabled/missing skills and a maximum-injection cap; preserves the original user prompt and degrades gracefully on errors.
* **Tests**
  * Comprehensive test suite for LLM and fuzzy scoring, injection behavior, handling of disabled/missing skills, selection caps, state reset after history compaction, and robust error handling.

### DIFF
--- a/code_puppy/plugins/auto_skill_activator/__init__.py
+++ b/code_puppy/plugins/auto_skill_activator/__init__.py
@@ -1,0 +1,1 @@
+"""Auto Skill Activator plugin."""
--- a/code_puppy/plugins/auto_skill_activator/register_callbacks.py
+++ b/code_puppy/plugins/auto_skill_activator/register_callbacks.py
@@ -1,0 +1,158 @@
+"""Auto Skill Activator plugin.
+
+Problem this solves:
+    Skills are injected into the system prompt as XML, but the LLM only calls
+    activate_skill() if it independently decides the skill matches the task.
+    In practice this rarely happens unless the user says "use skill" explicitly.
+
+    This plugin fixes that by scoring the user prompt against all installed
+    skill descriptions at prompt-assembly time, and auto-injecting the full
+    SKILL.md content for any strong matches — no LLM decision required.
+
+How it works:
+    Hooks into `get_model_system_prompt` (fires during prompt assembly).
+    Scores each skill using rapidfuzz token_set_ratio (already a dependency).
+    Any skill scoring >= AUTO_ACTIVATE_THRESHOLD gets its full SKILL.md content
+    appended to the system prompt directly before the agent runs.
+
+This mirrors how agents work: explicit deterministic activation vs passive hope.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, List, Optional
+
+from code_puppy.callbacks import register_callback
+
+logger = logging.getLogger(__name__)
+
+# Fuzzy match threshold (0-100).
+# 65 is permissive enough to catch synonyms/paraphrasing.
+AUTO_ACTIVATE_THRESHOLD = 65
+
+# Max skills to auto-activate per run (avoid flooding context window)
+MAX_AUTO_ACTIVATE = 3
+
+
+def _score_prompt_against_skill(user_prompt: str, skill_text: str) -> float:
+    """Score how well a user prompt matches a skill using fuzzy token matching.
+
+    Uses rapidfuzz.fuzz.token_set_ratio which handles:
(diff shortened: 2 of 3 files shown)


mpfaffenberger:
I like the idea, but I think it would be much more powerful to leverage a smaller model like Haiku, one of the GPT Nanos, or even an open source one like Gemma 4 / GLM Flash to choose which skill to inject - if _any_. 

We would also need the background model to note if a compaction occurs at any given time to re-inject the skill if necessary. 

Then we could further generalize this concept beyond skill injection. Background steering agents.

Prithvi1994:
Great feedback — I'm on board with the direction. Here's my plan:

1. **Swap fuzzy matching for a small background model** — Replace `rapidfuzz.token_set_ratio` with a lightweight LLM call (Haiku / GPT-4o-mini / Gemma 4 IT / GLM Flash) to semantically evaluate which skills to inject (if any). This gives us actual understanding of intent vs. token overlap, which is the right call.

2. **Re-inject on compaction** — Add a hook so the background steering model re-evaluates and re-injects the relevant skill content after any context compaction event, keeping the skill context alive across the full session.

3. **Generalize to background steering agent** — Once the above is solid, extend the pattern beyond skill injection into a broader "background steering" concept — a lightweight model that runs alongside the main agent to handle skill selection, context management, and other steering decisions.

I'll push an updated commit with items 1 and 2 first, then follow up with the generalization. Thanks for the direction!

Prithvi1994:
@mpfaffenberger , Please review the changes when you get a chance. Thank you.

axacode:
Hey @Prithvi1994 , i reviewed the code just now. Looks good from my end.

axacode (approved):

thomwebb:

thomwebb:

PR #257, opened 2026-04-07
Interactive Runtime: Queue, Interject, Thinking, etc works.

Nate0-1999:
## Demo Video

[Interactive runtime demo video](https://github.com/Nate0-1999/code_puppy/releases/download/interactive-runtime-demo-2026-04-07/CodePuppyInteractiveRuntimeDemo.mp4)

## Summary

This PR introduces an interactive runtime that lets the user queue prompts and send interjects while their Code Puppy is already running.

The core goal is smooth live interaction across the runtime surfaces that matter in real use:
- prompt input
- queued prompts
- interjects
- ephemeral thinking / status / response preview
- durable final transcript output
- shell commands
- multiple parallel agents
- Wiggums of parallel agents
- combinations in between

I tested these surfaces personally across macOS and Windows, and the interactive runtime / state handling feels solid.

The latest updates also address the CodeRabbit correctness comments without broadening the design:
- active interactive runtime is always cleared on exit paths
- queue-full interject rejection no longer cancels active work
- prompt-only execution correctly unpacks `(response, task)`
- background command lifecycle only emits `completed` on real success
- shell runtime notifications cannot short-circuit cleanup
- Antigravity auth no longer reports success if model registration fails
- prompt-surface reasoning-tool suppression handles partial streamed prefixes

## What this introduces

- a prompt-surface interactive runtime
- queueable prompts while a run is already active
- interjects while a run is already active
- ephemeral live thinking / status / response preview before durable output is printed
- smooth coordination with shell commands
- smooth coordination with multiple parallel agents
- smooth coordination with Wiggum flows and Wiggums of parallel agents
- foreground ownership of ephemeral UI so background/sub-agent activity does not clobber it
- terminal-safe live rendering behavior across macOS and Windows
- graceful handling for malformed `replace_in_file` payloads

## Tested surfaces

I tested:
- prompt only
- prompt + queue
- prompt + interject
- prompt + reasoning
- prompt + tool calls
- prompt + shell commands
- prompt + shell carriage-return progress
- prompt + Wiggum
- prompt + parallel agents
- prompt + Wiggum + parallel agents
- prompt + shell + interject
- prompt + shell + queue
- prompt + parallel agents + foreground ephemeral UI
- macOS terminal behavior
- Windows terminal behavior

## Guardrails and regression coverage

The guardrails and regression coverage for this work enforce a few core runtime rules:

- reasoning should surface through structured reasoning output, not low-level tool-call token spam
- mutable live tool progress and shell carriage-return progress should stay ephemeral and update in place instead of polluting the durable transcript
- live response text can appear ephemerally while the run is active, but the final durable `AGENT RESPONSE` still renders once
- durable structured outputs still render above the prompt normally
- foreground ephemeral UI is foreground-only, so parallel sub-agents cannot overwrite or clear it
- terminals that cannot safely support live redraw should degrade cleanly instead of leaking raw ANSI / control-sequence garbage
- Windows clipboard fallback should continue to work cleanly
- malformed file-edit payloads should fail as normal tool errors instead of bubbling up as raw internal exceptions

## Notes

This is not a broad product redesign. It introduces the interactive runtime and makes the live combinations between prompt, queue, interject, shell, Wiggum, and parallel agents behave correctly.

## Summary by CodeRabbit

* **New Features**
  * Interactive prompt queuing and interjection with a configurable queue limit (/set queue_limit).
  * Background-friendly OAuth flows now run as cancellable background commands.

* **Bug Fixes**
  * Cleaner live-progress updates: carriage-return progress, ephemeral status/preview, and spinner behavior no longer corrupt the transcript or duplicate final responses.
  * Improved shell output handling across terminals and paste fallback behavior.

* **Documentation**
  * Added Implementation Guardrails and Interactive Regression Checklist.


mpfaffenberger:
Hey @Nate0-1999 - I noticed a visual artifact at 1:46 in the video as well as flickering at other points. This is the main reason why I avoid TUI-esque features. I think we can support queue + interjection without a full TUI rewrite - but there has to be an ability to pause output, even if the output is accumulating internally. This requires a pretty big message system rewrite, and the addition of new lifecycle hooks. 

I had taken a stab at this in ~december, and couldn't quite get it working as it's pretty complex. I'll take another stab at this at some point.

thomwebb:

PR #261, opened 2026-04-11
fix(autocomplete): register /clear command to enable autocomplete suggestions (#254)

### LINKED ISSUE #254: `/clear` command not listed in autocomplete popup
### Summary

The `/clear` command is supported and works as expected when executed directly, but it does not appear in the autocomplete popup when typing `/cl`. Only `/clipboard` is shown. This misleads users into thinking `/clear` is unavailable.

### Background

I frequently use `/clear` in code-puppy to reset the context between tasks. While the command itself works, it's missing from the command suggestion popup, which impacts discoverability and user experience.

### Steps to Reproduce

- Start code-puppy.
- Type `/cl` in the input box.
- Observe the autocomplete popup for available commands (see attached screenshot).

### Screenshot

Image

### Expected Behavior

The popup should list both `/clear` and `/clipboard` as available commands when `/cl` is typed.

### Actual Behavior

The popup only lists `/clipboard`. `/clear` is missing from the suggestions, even though it is a valid and working command.

### Notes

Executing `/clear` manually works, so this appears to be an issue with the command registry or autocomplete logic, not the command handler itself.

zifrz:
## Summary
Fixes #254: The `/clear` command now appears in autocomplete suggestions when typing `/cl`.

## Root Cause
The `/clear` command was previously handled via an inline conditional check in the main interactive loop rather than being registered through the `@register_command` decorator. Since `SlashCompleter` only yields suggestions from commands in the registry (`get_unique_commands()`), `/clear` was invisible to tab-completion.

## Changes
- Added `@register_command` decorator to the clear handler in `code_puppy/command_line/core_commands.py`
- Implemented `handle_clear_command()` preserving original behavior:
  - Clears conversation history via `agent.clear_message_history()`
  - Rotates autosave session via `finalize_autosave_session()`
  - Clears pending clipboard images
  - Emits appropriate user feedback messages
- Removed `/clear` from the inline handler in `cli_runner.py`; the bare `clear` 
  (without slash) is retained there for backwards compatibility
- `/clear` is now fully owned by the registry — autocomplete, `/help` listing, 
  and execution all go through `handle_clear_command()`
  
## Out of Scope: `/exit` and `/quit`
Applying the same split to `/exit` and `/quit` is not straightforward. The inline handler for these commands needs to `break` out of the interactive loop — a side effect that cannot be expressed in the current registry return protocol (`True`/`str`/`False`). 
Moving them to the registry would require either introducing a sentinel return value or restructuring the loop exit mechanism, both of which are beyond the scope of this fix. 
Flagged for a follow-up.

## Testing
- [x] Autocomplete: Typing `/cl` + TAB now shows both `/clear` and `/clipboard`
- [x] Execution: `/clear` properly resets conversation history and rotates autosave sessions
- [x] Clipboard: Pending clipboard images are cleared on execution
- [x] Backwards compatibility: Plain `clear` (without `/`) continues to work
- [x] Tests & Linting: `uv run pytest tests/command_line/ -v` & `uv run ruff check . && uv run ruff format .`

## Notes
- No breaking changes; only adds missing registration metadata
- Follows the same registration pattern as other core commands (`/clipboard`, `/cd`, etc.)

## Summary by CodeRabbit

* **New Features**
  * Added a clear command to wipe the in-memory conversation and rotate to a new autosave session.
  * After clearing, the app warns that prior interactions won’t be remembered and shows the new session id.
  * Clears any pending clipboard image attachments and reports how many were removed.

* **Refactor**
  * Clear activation narrowed to the plain "clear" input (previous alternate form no longer triggers).

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -625,8 +625,8 @@ async def interactive_mode(message_renderer, initial_command: str = None) -> Non
             # The renderer is stopped in the finally block of main().
             break
 
-        # Check for clear command (supports both `clear` and `/clear`)
-        if task.strip().lower() in ("clear", "/clear"):
+        # Check for clear command (supports only `clear`)
+        if task.strip().lower() in ("clear"):
             from code_puppy.command_line.clipboard import get_clipboard_manager
             from code_puppy.messaging import (
                 emit_info,
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -626,7 +626,7 @@ async def interactive_mode(message_renderer, initial_command: str = None) -> Non
             break
 
         # Check for clear command (supports only `clear`)
-        if task.strip().lower() in ("clear"):
+        if task.strip().lower() == "clear":
             from code_puppy.command_line.clipboard import get_clipboard_manager
             from code_puppy.messaging import (
                 emit_info,
--- a/code_puppy/command_line/core_commands.py
+++ b/code_puppy/command_line/core_commands.py
@@ -170,7 +170,7 @@ def handle_paste_command(command: str) -> bool:
 @register_command(
     name="clear",
     description="Clear conversation history and rotate autosave session",
-    usage="/clear, clear",
+    usage="/clear",
     aliases=[],
     category="core",
 )
--- a/code_puppy/command_line/core_commands.py
+++ b/code_puppy/command_line/core_commands.py
@@ -175,6 +175,12 @@ def handle_paste_command(command: str) -> bool:
     category="core",
 )
 def handle_clear_command(command: str) -> bool:
+    """Clear conversation history, rotate autosave session, and clear pending clipboard images.
+
+    Resets the agent's context so it starts fresh, persists the current session
+    to autosave, and notifies the user of the cleanup.
+    """
+
     from code_puppy.command_line.clipboard import get_clipboard_manager
     from code_puppy.messaging import emit_warning, emit_system_message, emit_info
     from code_puppy.agents import get_current_agent
--- a/code_puppy/command_line/core_commands.py
+++ b/code_puppy/command_line/core_commands.py
@@ -168,6 +168,35 @@ def handle_paste_command(command: str) -> bool:
 
 
 @register_command(
+    name="clear",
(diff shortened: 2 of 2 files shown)


mpfaffenberger:
This will be an improvement.

Could you please delete the legacy `/clear` handler as well in this P/R?

zifrz:
Thanks for the feedback!

At the moment the inline handler in `cli_runner.py` handles both `/clear` and `clear`, 
while the command registered in `core_commands.py` provides autocomplete for `/clear`.

If we remove the inline handler, `clear` (without the slash) would stop working, which would break existing usage.

We’re following the same pattern for `/exit` and `/quit` to support their non-slash variants as well.

I can think of 3 ways to resolve this:
1. We can update the inline handler to support only `clear`, 
and let `/clear` be handled entirely by the registered command so autocomplete works as expected. 
However, that would make the design a bit inconsistent.

2. We can also remove legacy handlers for `/clear` and `/exit`, `/quit`.
This would make things more consistent, but at the cost of breaking all non-slash variants (`clear`, `exit`, and `quit`).

3. We can have all slash variants handled by `register_command` and keep non-slash variants in the legacy handlers. 
This would be a larger change.

Or we could also keep things as they are and make no further changes as part of this PR. 😅

Also, I noticed a small issue in my PR and fixed it.
The `usage` string in `register_command` for `clear` now correctly uses only `/clear` instead of `/clear, clear`.

Let me know the approach you prefer.

mpfaffenberger:
I'd recommend option 3

zifrz:
I've applied fix for `/clear`, it will now be handled by `handle_clear_command()` in 
the registry and not by inline handler in `cli_runner.py`.

For `/exit` and `/quit`, I ran into an issue.
The inline handler needs to `break` out of the interactive loop, and there's no way to express that through the existing registry return convention (`True`/`str`/`False`) from `core_commands.py`. 
Moving them would require either:
1. a sentinel return value or 
2. restructuring the loop exit mechanism

Both kind of feel like they belong in a separate PR. 
Happy to open a follow-up for that if you agree.

mpfaffenberger:
hey, at some point this functionality got added, sorry I didn't catch your P/R before that

PR #293, opened 2026-04-17
Fix Gemini streaming output issue

tymorton-google:
Enables argument streaming in API request and handles partial arguments in response iterator.

## Summary by CodeRabbit

* **New Features**
  * Added automatic tool-calling configuration for Gemini streaming responses.

* **Bug Fixes**
  * Improved streaming function-call delta handling to properly accumulate tool state and arguments across streamed responses.

### DIFF
--- a/code_puppy/gemini_model.py
+++ b/code_puppy/gemini_model.py
@@ -678,6 +678,53 @@ class GeminiModel(Model):
         )
 
 
+_MISSING = object()
+
+
+def _extract_partial_value(p_arg: dict) -> Any:
+    for key in [
+        "stringValue",
+        "numberValue",
+        "boolValue",
+        "nullValue",
+        "structValue",
+        "listValue",
+    ]:
+        if key in p_arg:
+            val = p_arg[key]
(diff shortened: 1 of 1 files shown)


tymorton-google:
Thanks for the detailed review! Here are my responses to your points:

1. **Support for `streamFunctionCallArguments`**: While it is documented as an experimental feature in some places, the Google Developer Knowledge base confirms that `streamFunctionCallArguments` is supported in both `v1` and `v1beta` of the Gemini API for Gemini 3 Pro and later models. Our benchmark tests verified that it successfully enables continuous thinking streams during tool calls, so we have decided to keep it.

2. **JSONPath Handling and Value Types**: You are absolutely right about the naive JSONPath slicing and missing value types. I have pushed a new commit that implements proper JSONPath parsing (handling nested objects and arrays) and supports all value types (`stringValue`, `numberValue`, `boolValue`, `nullValue`, `structValue`, `listValue`).

Please review the updated changes!

mpfaffenberger:
Awesome! I'll check it out!

PR #297, opened 2026-04-19
feat: add model visibility toggle to picker and settings menus

ma-juang:
## Summary
Adds model visibility toggling for the model selection menu.

## Problem
Currently the model selection menu that shows up after `/model` will show you a bunch of models, some of which you may not be interested in. Typing the name of a model can be tricky because it's hard to remember all the prefix/postfix for the full model name, and the model list may contain sonnet-4.5, sonnet-4.6, ... and so on.

## Solution
Add toggling functionality to hide any model you don't want to see. This persists by writing to a config file. You can still search for and see hidden models, and you can hit a button to show all models including hidden ones. If you want to unhide a model, show all -> reselect the ones you want.

The default look and feel is unchanged, except for the hints about the two new keybinds (space and tab). What's added is the ability to customize.

Warning: this does change the core logic but I think it probably needs to. Let me know if there's a better way!

I did look into changing the /agents menu too but the conventions are a little different there than the model selection menu. This PR doesn't include any change for menus other than model selection.

### Files in scope
- `code_puppy/command_line/model_picker_completion.py` (changed)
- `code_puppy/command_line/model_settings_menu.py` (changed)
- `code_puppy/item_visibility.py` (added) - maybe could be leveraged if you want to extend this to other menus
- `tests/command_line/test_model_visibility_tui.py` (added)
- `tests/test_item_visibility.py` (added)

### Changes
- Add VisibilityStore class for reusable visibility persistence
- Add Space key to toggle model visibility (hides selected model)
- Add Tab key to show/hide all (session-level filter)
- Persist hidden models across sessions via JSON config, and handle corrupted configs
- Add tests for visibility functionality
- Tests preserve user's visibility config during test runs

### In Action
Screenshot_20260418_223656

Model List (before making changes)

Screenshot_20260418_223510

Model List (after hiding some stuff)

Screenshot_20260418_223550

Model List (after hiding some stuff and toggling the Show all)

### DIFF
--- a/code_puppy/command_line/model_picker_completion.py
+++ b/code_puppy/command_line/model_picker_completion.py
@@ -16,6 +16,10 @@ from code_puppy.command_line.pagination import (
     get_total_pages,
 )
 from code_puppy.config import get_global_model_name
+from code_puppy.item_visibility import (
+    load_hidden_models,
+    prune_stale_entries,
+)
 from code_puppy.list_filtering import query_matches_text
 from code_puppy.model_switching import set_model_and_reload_agent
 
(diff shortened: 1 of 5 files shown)


mpfaffenberger:
Hi @ma-juang schedule a call with me to show me

ma-juang:
Sorry for falling off the face of the earth here. I've been busy with other things, and also actually coming around on just using `/model` and the autocomplete hints to quickly get to what I want.

Let me close this for now while I think about if I still want/need this functionality and recommend for others, as it's adding some complexity.

Maybe if there's work on a rewrite of a lot of the TUI components like agent selector and model selector and many of these menus, you could take this idea or something else to help manage the discovery/selection process of these different options in a different way.

PR #312, opened 2026-04-30
Autosave interrupted interactive sessions

nhicks00:
## Summary

Fixes autosave persistence for interrupted Code Puppy sessions.

If a user cancels an agent run with Ctrl-C and then exits, the interactive loop previously skipped the normal autosave path because cancelled runs return `None` and immediately continue to the next prompt. That could leave the autosave file missing the latest in-memory conversation history.

## What Changed

- Adds a small `_save_autosave_snapshot()` helper around `auto_save_session_if_enabled()`.
- Saves immediately before returning to the prompt after a cancelled agent run.
- Saves once during CLI shutdown so Ctrl-D, `/exit`, top-level interrupt, or process cleanup has a final persistence attempt.
- Saves from single-prompt mode in `finally` as a defensive cleanup path.
- Adds test coverage for the cancelled interactive prompt path.

## Validation

```bash
uv run pytest --no-cov tests/test_cli_runner_coverage.py tests/test_cli_runner_full_coverage.py -q
uv run ruff check code_puppy/cli_runner.py tests/test_cli_runner_full_coverage.py
uv run ruff format --check code_puppy/cli_runner.py tests/test_cli_runner_full_coverage.py
```

Results:

- `73 passed`
- Ruff check passed
- Ruff format check passed

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -51,6 +51,13 @@ from code_puppy.version_checker import default_version_mismatch_behavior
 plugins.load_plugin_callbacks()
 
 
+def _save_autosave_snapshot() -> bool:
+    """Persist the current autosave snapshot if autosave is enabled."""
+    from code_puppy.config import auto_save_session_if_enabled
+
+    return auto_save_session_if_enabled()
+
+
 async def main():
     """Main async entry point for Code Puppy CLI."""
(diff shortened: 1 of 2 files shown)


mpfaffenberger (approved):

mpfaffenberger:
This looks straightforward - merge?

Issue #333, opened 2026-05-09
copy-paste bug


thomwebb:
<img width="504" height="106" alt="Image" src="https://github.com/user-attachments/assets/63264b74-b4ee-47a7-9901-92c5a48a1083" />

<img width="744" height="409" alt="Image" src="https://github.com/user-attachments/assets/47113b5d-e209-4865-a686-b8fc4296c425" />

mpfaffenberger:
what is it and how do we fix it?

thomwebb:
When the agent's response contains a "text" box the user's clipboard gets overwritten with the text box contents.

Solution TBD

mpfaffenberger:
https://github.com/mpfaffenberger/termflow - the bug is in here probably

thomwebb:
@mpfaffenberger you're correct. PR #335 has the fix

mpfaffenberger:
Closing completed.

PR #335, opened 2026-05-09
fix(display): disable clipboard features in termflow rendering

thomwebb:
- Prevent termflow from enabling clipboard-related render features in both streamed and non-streamed markdown output
- Reduce the risk of terminal side effects when rendering assistant responses in the console
- Keep display behavior focused on text rendering while avoiding unnecessary clipboard integration


thomwebb:
Resolves #333

mpfaffenberger (approved):

PR #348, opened 2026-05-20
feat: @file expansion in rules files + CLAUDE.md fallback

breedx:
## What

- Adds `@path` reference expansion to `load_puppy_rules()` — any line consisting solely of `@path` in a rules file is replaced with the contents of that file, resolved relative to the rules file's directory
- Expansion is **one level deep only** — refs inside included files are left as-is
- Falls back to `CLAUDE.md` / `claude.md` in the project root when no `AGENTS.md` exists anywhere

## Why

Allows `AGENTS.md` (or `CLAUDE.md`) to compose rules from multiple files via `@` references, rather than requiring all rules to be inlined. Useful for referencing external philosophy/rule sets without duplicating content into a single flat file.

## Security

Path expansion is constrained to the directory containing the rules file:
- Absolute paths (e.g. `@/etc/hostname`) are rejected
- Parent traversal (e.g. `@../../.env`) is rejected  
- Symlinks that resolve outside the base directory are rejected
- Unknown/unsafe refs are left as their original `@token` so the agent can see what failed

## Changes

- `code_puppy/agents/_builder.py`: add `_AT_REF_RE`, `_resolve_at_ref()`, `_expand_at_references()`, `_read_rules_file()`, update `load_puppy_rules()`
- `tests/agents/test_builder_load_puppy_rules.py`: add `TestExpandAtReferences` (7 tests), `TestResolveAtRef` (7 tests), `TestClaudeMdFallback` (6 tests)

## Tests

- 30/30 targeted tests passing
- 572/572 agent tests passing
- Zero ruff warnings

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -39,24 +39,45 @@ from code_puppy.model_factory import ModelFactory, make_model_settings
 _AGENT_RULE_FILES = ("AGENTS.md", "AGENT.md", "agents.md", "agent.md")
 _CLAUDE_FALLBACK_FILES = ("CLAUDE.md", "claude.md")
 _CODE_PUPPY_DIR = ".code_puppy"
-_AT_REF_RE = re.compile(r"^@(\S+)", re.MULTILINE)
+_AT_REF_RE = re.compile(r"^@([^\s]+)\s*$", re.MULTILINE)
+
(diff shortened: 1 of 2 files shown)


breedx:
closing this for now as the underlying code has changed. will re-open after review.

PR #356, opened 2026-05-21
fix: skip sanitization for LiteLLM __thought__ smuggled tool_call_ids

breedx:
## Problem

`sanitize_tool_call_ids` in `code_puppy/agents/_history.py` runs on every request via `history_processors`. It rewrites any `tool_call_id` containing characters outside Anthropic's required `^[a-zA-Z0-9_-]+$` pattern.

That pattern is Anthropic-specific — the OpenAI spec doesn't restrict `tool_call_id` characters. The sanitizer applies it unconditionally to all providers.

Gemini's native API carries `thoughtSignature` as a separate field on `FunctionCall`. The OpenAI-compat schema has no such field, so LiteLLM smuggles it into `tool_call_id` as `<id>__thought__<base64-sig>`. Standard base64 uses `+` and `/`, which trip the regex. Gemini requires the full encoded value to round-trip byte-for-byte through LiteLLM — even the `_<6digit>` collision-guard suffix alone is enough to corrupt the signature and cause a 400 on the next tool turn.

Result: every Gemini tool follow-up after the first 400s with **"Corrupted thought signature."** — deterministic, total breakage.

## Fix

Add `_LITELLM_THOUGHT_RE = re.compile(r"__thought__[A-Za-z0-9+/=]+$")` and skip ids matching it in the `bad_ids` collection pass. The id is left byte-for-byte intact.

The collision guard isn't needed for these ids anyway — the embedded signature makes each id globally unique, so two different ids can never sanitize to the same base string.

The Anthropic-tier sanitization for other providers (Kimi, Firepass, etc.) is unaffected.

## Trade-off

If a user switches from a Gemini model to an Anthropic model mid-conversation while tool calls are in flight, the `__thought__`-bearing ids will survive and Anthropic will reject them with a 400. This is a rare edge case; the prior behaviour on that path was silently corrupt tool linkage, not a clear error.

### DIFF
--- a/code_puppy/agents/_history.py
+++ b/code_puppy/agents/_history.py
@@ -282,6 +282,9 @@ def filter_huge_messages(
 _ANTHROPIC_TOOL_ID_RE = re.compile(r"^[a-zA-Z0-9_-]+$")
 # Character-level replacement: swap any character NOT in the allowed set.
 _BAD_TOOL_ID_CHAR_RE = re.compile(r"[^a-zA-Z0-9_-]")
+# LiteLLM smuggles Vertex/Gemini thoughtSignature blobs as
+# `<id>__thought__<base64-payload>` at the end of tool_call_id.
+_LITELLM_THOUGHT_RE = re.compile(r"__thought__[A-Za-z0-9+/=]+$")
 
 
 def sanitize_tool_call_ids(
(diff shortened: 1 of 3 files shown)


breedx:
Note: the `ruff format` CI failure is pre-existing — `code_puppy/plugins/frontend_emitter/emitter.py` and `tests/plugins/test_frontend_emitter_session_channel.py` were already unformatted on `main` before this branch. Happy to fix them here if that's preferred, but didn't want to mix unrelated style changes into this bug fix.

mpfaffenberger (approved):

PR #359, opened 2026-05-23
fix(mcp): /mcp start auto-binds server to current agent

JeffMelton:
## Problem

When an MCP server is added to `~/.code_puppy/mcp_servers.json` outside the install wizard — by hand-editing, by a third-party tool, or by an agent helpfully writing the config — no corresponding entry is created in `~/.code_puppy/mcp_agent_bindings.json`. The server is registered but bound to nothing.

Running `/mcp start <name>` then gives the user a perfectly cheerful response:

```
🚀 Starting server: nu (subprocess starting in background)
Agent reloaded with updated servers
```

...with zero new tools available, because `get_servers_for_agent(agent_name=...)` silently filters out the unbound server (the skip is a `logger.debug`, invisible at default log level).

### Reproduction

1. Add to `~/.code_puppy/mcp_servers.json`:
   ```json
   {"mcp_servers": {"nu": {"type": "stdio", "command": "/opt/homebrew/bin/nu", "args": ["--mcp"], "enabled": true}}}
   ```
2. Restart Code Puppy.
3. Run `/mcp start nu`.
4. **Expected:** agent has `nu`'s tools available.
5. **Actual:** start message looks fine, but no tools are injected. `mcp_agent_bindings.json` is never written.

The current workaround is `/agents → B` to manually bind the server, which is undiscoverable for most users.

## Fix

Treat `/mcp start <name>` as the user's explicit binding intent. After a successful `start_server_sync`, and **before** the agent reload (so the rebuilt toolset actually picks up the new server), check `is_bound(current_agent, server_name)` and call `set_binding(..., auto_start=True)` if the binding is missing. A single `emit_info` line tells the user what just happened and how to manage it later:

```
Also bound 'nu' to agent 'code-puppy' (use /agents → B to manage).
```

The helper is best-effort: any exception from the binding layer is logged but never propagated, so a corrupted bindings file can't break `/mcp start`.

### Why this approach

This is **additive**, not redefinitive. It doesn't change the meaning of strict opt-in anywhere else in the codebase — it adds one new path *into* the bindings system, justified by the fact that typing `/mcp start <name>` is an unambiguous signal of "yes, I want this." Existing users with curated `mcp_agent_bindings.json` see no behavior change (the new branch only fires when a binding is missing).

An alternative considered and rejected: redefining "no entry for agent X" to mean "all servers visible" until the user creates their first binding. That design is more invasive, has no equivalent of the explicit user-facing notice, and is harder to defend in review. Happy to revisit if maintainers prefer it.

### Optional escape hatch

Not implemented in this PR, but trivial to add if desired: a `--no-bind` flag on `/mcp start` to preserve pure strict-opt-in behavior on demand. Suggesting we leave that for follow-up unless requested here.

## Testing

- 4 new tests in `tests/command_line/mcp/test_start_command.py`:
  - `test_auto_binds_when_unbound` — happy path
  - `test_no_rebind_when_already_bound` — idempotency
  - `test_auto_bind_failure_does_not_break_start` — corrupted bindings file doesn't break start
  - `test_bind_runs_before_reload` — ordering invariant (binding must persist before agent reload)
- Full file: 14/14 passing
- `tests/command_line/mcp/` + `tests/mcp/` regression: **1241 passed, 6 skipped, 0 failed**
- `ruff check --fix` and `ruff format` both clean

## Scope

This PR fixes the most common manifestation. Two related follow-ups exist as ideas worth tracking separately:

- A warning when `get_servers_for_agent` skips a registered-but-unbound server, for the user who hand-edits and never runs `/mcp start` at all. (Belt-and-suspenders; today's fix covers everyone who *does* run `/mcp start`.)
- A look at the MCP install wizard UX, which is what pushes users toward hand-editing in the first place.

Both intentionally out of scope here to keep this PR small and reviewable.

### DIFF
--- a/code_puppy/command_line/mcp/start_command.py
+++ b/code_puppy/command_line/mcp/start_command.py
@@ -95,8 +95,16 @@ class StartCommand(MCPCommandBase):
                 # Reload the agent to pick up the newly enabled server
                 # NOTE: We don't block or wait - the server will be ready
                 # when the next prompt runs (pydantic-ai handles connection)
+                #
+                # Auto-bind to the current agent if not already bound. The
+                # user just typed `/mcp start <name>` -- that's an explicit
+                # opt-in. Without this, get_servers_for_agent() will silently
+                # filter the new server out of the rebuilt toolset and the
+                # user gets a cheerful "started" message with zero tools.
+                # See CPUP-ne1 for the gory details.
                 try:
                     agent = get_current_agent()
+                    self._ensure_binding(agent.name, server_name, group_id)
                     agent.reload_code_generation_agent()
                     # Clear MCP tool cache - it will be repopulated on next run
                     agent.update_mcp_tool_cache_sync()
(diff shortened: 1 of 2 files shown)


JeffMelton:
Cross-link: the belt-and-suspenders companion to this PR is now open as #361 (CPUP-6ym). This PR auto-binds on the active `/mcp start` path; #361 covers the passive hand-edit-`mcp_servers.json`-and-restart path by emitting a one-shot warning when `get_servers_for_agent` would silently skip a registered-but-unbound server. The two are independent and can land in either order.

mpfaffenberger:
Oops. So I merged your other P/R and then realized /mcp start doesn't auto-bind so I just implemented it. I didn't even see this one. 

The full functionality of both PRs now works as you intended.

JeffMelton:
> Oops. So I merged your other P/R and then realized /mcp start doesn't auto-bind so I just implemented it. I didn't even see this one. 
> 
> 
> 
> The full functionality of both PRs now works as you intended.

Thanks! Functionality is what we're after here.

PR #363, opened 2026-05-24
Remove line that gives warning

ryan-duve:
Currently, starting Code Puppy gives a warning from the TOML file because this feature takes an absolute date, not a relative date.  This PR removes the line entirely since I don't think we want a 2026-05-17 type line calcified in the repo.


ryan-duve:
It won't let me assign to @mpfaffenberger.

PR #365, opened 2026-05-25
feat(plugins): add notifier plugin — standalone desktop notifications

sskmlm:
Three public surfaces:
1. Python API  — from code_puppy.plugins.notifier import send_notification
2. Agent tool  — send_notification(...) via register_tools; any sub-agent can call it
3. Slash cmd   — /notify <title> [| <message>] for quick manual testing

Timeout / persistent options (all platforms):
  timeout=N / timeout=0 / persistent=True

Platform backends: alerter + osascript (macOS), WinRT + balloon (Windows), notify-send (Linux).

Focus-aware suppression — two-path strategy per host type:

  macOS:
    osascript frontmost-process query.

  Windows — two paths, both needed:

    Path 1 (classic PowerShell / cmd.exe / pwsh.exe standalone):
      GetConsoleWindow() returns the conhost.exe HWND backing our process.
      Compare directly to GetForegroundWindow() — if they match, we are
      focused.  Can NOT use the ancestor walk here: conhost.exe is a
      *child* of powershell (spawned by it), so it never appears in the
      upward ancestor chain from python.exe.

    Path 2 (Windows Terminal — ConPTY):
      GetConsoleWindow() returns 0 / a hidden stub (no classic console).
      The visible HWND is owned by WindowsTerminal.exe, which IS an
      ancestor (WT → pwsh → python).  CreateToolhelp32Snapshot walk
      finds it correctly at any nesting depth.

  Linux: always notifies (no reliable cross-desktop focus API).

All focus backends fail open (error → False → notification sent).

Pure-function extraction for cross-platform testability:
  _pid_is_ancestor_of(target, start, parent_map) — no ctypes in tests

65 tests, ruff clean.


mpfaffenberger:
This is super cool - please run linters / tests

sskmlm:
> This is super cool - please run linters / tests

checks are passing now!

thomwebb:

PR #369, opened 2026-05-30
fix(agents): pass resolved model name to prepare_prompt_for_model hooks

thomwebb:
When a model fails to load (e.g., missing API key) and falls back to an alternative model, Code Puppy now correctly informs the agents and plugins about the actual model being used.

Previously,  passed  (the originally requested model) to , which meant plugins registered on / hooks received the wrong model name.

This fixes the issue so that hooks receive  (the fallback model that was actually loaded), ensuring plugins can properly customize prompts and behavior based on the model in use.


thomwebb:
Closing to recreate against updated base

PR #370, opened 2026-05-30
fix(agents): pass resolved model name to prepare_prompt_for_model hooks

thomwebb:
When a model fails to load (e.g. missing API key) and falls back to an alternative model, Code Puppy now correctly informs the agents and plugins about the actual model being used.

Previously, _assemble_instructions() passed agent.get_model_name() (the originally requested model) to prepare_prompt_for_model(), which meant plugins registered on prepare_model_prompt/get_model_system_prompt hooks received the wrong model name.

This fixes the issue so that hooks receive resolved_model_name (the fallback model that was actually loaded), ensuring plugins can properly customize prompts and behavior based on the model in use.


thomwebb:
Closing to wait for PR #368 to merge first

PR #377, opened 2026-06-01
feat: Add per-invocation sub-agent model overrides

rshide:
## Summary

Adds support for per-invocation model overrides when calling sub-agents.

This lets a caller run a sub-agent with a specific configured model for a single task without changing the agent’s pinned model or global/default model configuration.

## Why

Some sub-agent calls benefit from temporarily using a different model:

- Use a stronger model for one complex planning/review task.
- Use a cheaper/faster model for a simple delegated task.
- Experiment with a model on a sub-agent without repinning that agent permanently.

## Examples

From the CLI, a user can ask Code Puppy to delegate to a sub-agent with its normal configured model:

```text
> Ask the reviewer agent to review this implementation.
```

For a deeper one-off review, the user can request a stronger model just for that sub-agent call:

```text
> Ask the reviewer agent to do a deeper architecture review of this change using gpt-5.
```

For simpler delegated work, the user can request a cheaper/faster model without repinning the sub-agent:

```text
> Ask the planner agent to outline the next steps using gpt-4o-mini.
```

Under the hood, this is passed through the `invoke_agent` tool as an optional `model_name`:

```text
invoke_agent(
  agent_name="reviewer",
  prompt="Do a deeper architecture review of this change",
  model_name="gpt-5"
)
```

## What changed

- Added optional `model_name` support to the `invoke_agent` tool.
- Propagated model override metadata through sub-agent invocation messages.
- Added temporary model override handling so the override applies only to that invocation.
- Preserved model precedence:
  1. per-invocation override
  2. pinned agent model
  3. global/default model
- Updated rendering/tests for the new model override behavior.

### DIFF
--- a/code_puppy/agents/agent_code_puppy.py
+++ b/code_puppy/agents/agent_code_puppy.py
@@ -86,7 +86,6 @@ Important rules:
 - Read existing files before modifying them
 - Prefer replace_in_file over create_file. Keep diffs small (100-300 lines).
 {r["loop_rule"]}
-- When invoking sub-agents, you may pass model_name to use a specific configured model for that invocation only. Use it only when the user/project instructions specify model preferences or when escalating a failed/complex subtask; do not change the global model just to run a sub-agent.
 - Continue autonomously unless user input is definitively required
 """
         # NOTE: runtime ``load_prompt`` fragments (plugin-injected notes such
--- a/code_puppy/agents/agent_code_puppy.py
+++ b/code_puppy/agents/agent_code_puppy.py
@@ -86,6 +86,7 @@ Important rules:
 - Read existing files before modifying them
 - Prefer replace_in_file over create_file. Keep diffs small (100-300 lines).
 {r["loop_rule"]}
+- When invoking sub-agents, you may pass model_name to use a specific configured model for that invocation only. Use it only when the user/project instructions specify model preferences or when escalating a failed/complex subtask; do not change the global model just to run a sub-agent.
 - Continue autonomously unless user input is definitively required
 """
         # NOTE: runtime ``load_prompt`` fragments (plugin-injected notes such
--- a/code_puppy/agents/agent_creator_agent.py
+++ b/code_puppy/agents/agent_creator_agent.py
@@ -256,19 +256,23 @@ DONT USE THE TERMINAL TOOL TO RUN THE CODE WE WROTE UNLESS THE USER ASKS YOU TO.
 #### `list_agents()`
 Use this to list all available sub-agents that can be invoked
 
-#### `invoke_agent(agent_name: str, user_prompt: str, session_id: str | None = None)`
+#### `invoke_agent(agent_name: str, prompt: str, session_id: str | None = None, model_name: str | None = None)`
 Use this to invoke another agent with a specific prompt. This allows agents to delegate tasks to specialized sub-agents.
 
 Arguments:
 - agent_name (required): Name of the agent to invoke
-- user_prompt (required): The prompt to send to the invoked agent
+- prompt (required): The prompt to send to the invoked agent
 - session_id (optional): Kebab-case session identifier for conversation memory
   - Format: lowercase, numbers, hyphens only (e.g., "implement-oauth", "review-auth")
   - For NEW sessions: provide a base name - a SHA1 hash suffix is automatically appended for uniqueness
   - To CONTINUE a session: use the session_id from the previous invocation's response
   - If None (default): Auto-generates a unique session like "agent-name-session-a3f2b1"
+- model_name (optional): Configured model name to use for this invocation only
+  - Overrides the invoked agent's default/pinned model for this run
+  - Does not change global model settings or future invocations
+  - Use only when the user/project instructions specify model preferences or an escalation is warranted
 
-Returns: `{{response, agent_name, session_id, error}}`
+Returns: `{{response, agent_name, session_id, model_name, error}}`
 - **session_id in the response is the FULL ID** - use this to continue the conversation!
 
 Example usage:
@@ -276,14 +280,14 @@ Example usage:
 # Common case: one-off invocation (no memory needed)
 result = invoke_agent(
     agent_name="python-tutor", 
-    user_prompt="Explain how to use list comprehensions"
+    prompt="Explain how to use list comprehensions"
 )
 # result.session_id contains the auto-generated full ID
 
 # Multi-turn conversation: start with a base session_id
 result1 = invoke_agent(
     agent_name="code-reviewer", 
-    user_prompt="Review this authentication code",
+    prompt="Review this authentication code",
     session_id="auth-code-review"  # Hash suffix auto-appended
 )
 # result1.session_id is now "auth-code-review-a3f2b1" (or similar)
@@ -291,17 +295,32 @@ result1 = invoke_agent(
 # Continue the SAME conversation using session_id from the response
 result2 = invoke_agent(
     agent_name="code-reviewer",
-    user_prompt="Can you also check the authorization logic?",
+    prompt="Can you also check the authorization logic?",
     session_id=result1.session_id  # Use session_id from previous response!
 )
 
 # Independent task (different base name = different session)
 result3 = invoke_agent(
     agent_name="code-reviewer",
-    user_prompt="Review the payment processing code",
+    prompt="Review the payment processing code",
     session_id="payment-review"  # Gets its own unique hash suffix
 )
 # result3.session_id is different from result1.session_id
+
+# Per-call model override: use only real configured model names
+quick_review = invoke_agent(
+    agent_name="code-reviewer",
+    prompt="Cheap first pass: look for obvious issues only.",
+    model_name="<configured-fast-model-name>"
+)
+
+if quick_review.error or "uncertain" in (quick_review.response or "").lower():
+    deep_review = invoke_agent(
+        agent_name="code-reviewer",
+        prompt="Continue from the previous review and reason more deeply.",
+        session_id=quick_review.session_id,
+        model_name="<configured-strong-reasoning-model-name>"
+    )
 ```
 
 Best-practice guidelines for `invoke_agent`:
@@ -309,6 +328,7 @@ Best-practice guidelines for `invoke_agent`:
 • Clearly specify what you want the invoked agent to do
 • Be specific in your prompts to get better results
 • Avoid circular dependencies (don't invoke yourself!)
+• `model_name` is a temporary runtime override; omit it unless the task/user instructions justify a specific configured model
 • **Session management:**
   - Default behavior (session_id=None): Each invocation is independent with no memory
   - For NEW sessions: provide a human-readable base name like "review-oauth" - hash suffix is auto-appended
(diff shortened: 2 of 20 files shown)


mpfaffenberger:
The problem with this is that not all models in your model list may actually be properly configured. Like I might have a synthetic model with an expired key or usage capped.

rshide:
> The problem with this is that not all models in your model list may actually be properly configured. Like I might have a synthetic model with an expired key or usage capped.

@mpfaffenberger That's fair. I don't think we should treat a model being in the model list as proof that the model is healthy. I added a small change to raise if the model is unable to be initialized so that the user can choose a different model.

Do you think that is sufficient, or do you have any other thoughts?  Or is this something you're not interested in currently?

mpfaffenberger:
Can we connect and talk about this feature? I think it's a neat idea.

WSxDemise (approved):
This looks good to me - no glaring issues that I see.

AndrewTilson:
Hey @rshide — really glad to see this work! I'm working on the same feature in a parallel implementation and had some design observations that might be worth considering:

**1. The optional `model_name` on `invoke_agent` will get over-supplied by LLMs.** This is a well-documented failure mode — when a tool has an optional parameter, models tend to fill it in even when it's unnecessary. In practice that means routine sub-agent calls that should just use the agent's pinned model will end up overriding with whatever model the parent happened to think of. The tool's *name* is a much stronger guardrail than its docstring.

Suggestion: make the override a **separate, dedicated tool** (e.g. `invoke_agent_with_model`) rather than an optional kwarg on the existing tool. Within that dedicated tool, `model` could still be optional (defaulting to the agent's current model) — that keeps the flexibility for power-user agents that want a single tool, but anyone *not* explicitly opting into the dedicated tool can't accidentally override.

Bonus: it also enables tool-scoping. An agent's `available_tools` list can include `invoke_agent` without the override capability, or include the override tool only for specific power-user / judge-orchestrator archetypes.

**2. There's no companion model-discovery tool.** Without one, the LLM has to either remember model identifiers from training (unreliable — hallucinations are common) or be told them via the system prompt (brittle and stale). The override tool is barely usable without a way to discover valid identifiers in-band.

Suggestion: add a `list_available_models` tool that returns the set of valid model names plus lightweight metadata (`type`, `provider`, `context_length`, `description`, `supported_settings`). Worth being deliberate about the projection — `ModelFactory.load_config()` returns the full model config including `custom_endpoint` blocks with `$ENV_VAR`-style API key references. Those shouldn't be exposed to the LLM. An explicit allowlist of safe fields keeps future config additions from silently leaking.

Aside: your `BaseAgent.temporary_model_name_override()` context manager is genuinely a nice architectural seam — the override flows transparently to every deeper-stack reader of `get_model_name()`, which is cleaner than threading a local variable through.

rshide:
> Hey @rshide — really glad to see this work! I'm working on the same feature in a parallel implementation and had some design observations that might be worth considering:
> 
> **1. The optional `model_name` on `invoke_agent` will get over-supplied by LLMs.** This is a well-documented failure mode — when a tool has an optional parameter, models tend to fill it in even when it's unnecessary. In practice that means routine sub-agent calls that should just use the agent's pinned model will end up overriding with whatever model the parent happened to think of. The tool's _name_ is a much stronger guardrail than its docstring.
> 
> Suggestion: make the override a **separate, dedicated tool** (e.g. `invoke_agent_with_model`) rather than an optional kwarg on the existing tool. Within that dedicated tool, `model` could still be optional (defaulting to the agent's current model) — that keeps the flexibility for power-user agents that want a single tool, but anyone _not_ explicitly opting into the dedicated tool can't accidentally override.
> 
> Bonus: it also enables tool-scoping. An agent's `available_tools` list can include `invoke_agent` without the override capability, or include the override tool only for specific power-user / judge-orchestrator archetypes.
> 
> **2. There's no companion model-discovery tool.** Without one, the LLM has to either remember model identifiers from training (unreliable — hallucinations are common) or be told them via the system prompt (brittle and stale). The override tool is barely usable without a way to discover valid identifiers in-band.
> 
> Suggestion: add a `list_available_models` tool that returns the set of valid model names plus lightweight metadata (`type`, `provider`, `context_length`, `description`, `supported_settings`). Worth being deliberate about the projection — `ModelFactory.load_config()` returns the full model config including `custom_endpoint` blocks with `$ENV_VAR`-style API key references. Those shouldn't be exposed to the LLM. An explicit allowlist of safe fields keeps future config additions from silently leaking.
> 
> Aside: your `BaseAgent.temporary_model_name_override()` context manager is genuinely a nice architectural seam — the override flows transparently to every deeper-stack reader of `get_model_name()`, which is cleaner than threading a local variable through.

hey @AndrewTilson , thanks for the feedback!  And glad to hear that you're working on something similar.

good point about `model_name` getting provided on the default `invoke_agent` tool. I'll update this and split it into `invoke_agent` as the default path and `invoke_agent_with_model` for overrides.

do you feel strongly that `model_name` should be optional on `invoke_agent_with_model`? since calling `invoke_agent_with_model` without a model is basically just `invoke_agent`.

I’ll also add `list_available_models`, I was originally assuming callers would pass exact model names, but I agree that’s brittle.

long term I’d love to see support for model profiles like `cheap-and-fast`, `strong-reasoning`, `strong-architect`, etc., and let those resolve to configured aliases.

happy to coordinate with your implementation too if there’s a direction you’d prefer me to align with.

AndrewTilson:
> > Hey @rshide — really glad to see this work! I'm working on the same feature in a parallel implementation and had some design observations that might be worth considering:
> > **1. The optional `model_name` on `invoke_agent` will get over-supplied by LLMs.** This is a well-documented failure mode — when a tool has an optional parameter, models tend to fill it in even when it's unnecessary. In practice that means routine sub-agent calls that should just use the agent's pinned model will end up overriding with whatever model the parent happened to think of. The tool's _name_ is a much stronger guardrail than its docstring.
> > Suggestion: make the override a **separate, dedicated tool** (e.g. `invoke_agent_with_model`) rather than an optional kwarg on the existing tool. Within that dedicated tool, `model` could still be optional (defaulting to the agent's current model) — that keeps the flexibility for power-user agents that want a single tool, but anyone _not_ explicitly opting into the dedicated tool can't accidentally override.
> > Bonus: it also enables tool-scoping. An agent's `available_tools` list can include `invoke_agent` without the override capability, or include the override tool only for specific power-user / judge-orchestrator archetypes.
> > **2. There's no companion model-discovery tool.** Without one, the LLM has to either remember model identifiers from training (unreliable — hallucinations are common) or be told them via the system prompt (brittle and stale). The override tool is barely usable without a way to discover valid identifiers in-band.
> > Suggestion: add a `list_available_models` tool that returns the set of valid model names plus lightweight metadata (`type`, `provider`, `context_length`, `description`, `supported_settings`). Worth being deliberate about the projection — `ModelFactory.load_config()` returns the full model config including `custom_endpoint` blocks with `$ENV_VAR`-style API key references. Those shouldn't be exposed to the LLM. An explicit allowlist of safe fields keeps future config additions from silently leaking.
> > Aside: your `BaseAgent.temporary_model_name_override()` context manager is genuinely a nice architectural seam — the override flows transparently to every deeper-stack reader of `get_model_name()`, which is cleaner than threading a local variable through.
> 
> hey @AndrewTilson , thanks for the feedback! And glad to hear that you're working on something similar.
> 
> good point about `model_name` getting provided on the default `invoke_agent` tool. I'll update this and split it into `invoke_agent` as the default path and `invoke_agent_with_model` for overrides.
> 
> do you feel strongly that `model_name` should be optional on `invoke_agent_with_model`? since calling `invoke_agent_with_model` without a model is basically just `invoke_agent`.
> 
> I’ll also add `list_available_models`, I was originally assuming callers would pass exact model names, but I agree that’s brittle.
> 
> long term I’d love to see support for model profiles like `cheap-and-fast`, `strong-reasoning`, `strong-architect`, etc., and let those resolve to configured aliases.
> 
> happy to coordinate with your implementation too if there’s a direction you’d prefer me to align with.

We can try model_name as required in the new version of the tool. Just need to test it out and see that it handles it ok. Regarding model profiles maybe another idea is to add a tagging option to models so it can be used like you mentioned along with standard tags like thinking, vision, etc.

rshide:
> > > Hey @rshide — really glad to see this work! I'm working on the same feature in a parallel implementation and had some design observations that might be worth considering:
> > > **1. The optional `model_name` on `invoke_agent` will get over-supplied by LLMs.** This is a well-documented failure mode — when a tool has an optional parameter, models tend to fill it in even when it's unnecessary. In practice that means routine sub-agent calls that should just use the agent's pinned model will end up overriding with whatever model the parent happened to think of. The tool's _name_ is a much stronger guardrail than its docstring.
> > > Suggestion: make the override a **separate, dedicated tool** (e.g. `invoke_agent_with_model`) rather than an optional kwarg on the existing tool. Within that dedicated tool, `model` could still be optional (defaulting to the agent's current model) — that keeps the flexibility for power-user agents that want a single tool, but anyone _not_ explicitly opting into the dedicated tool can't accidentally override.
> > > Bonus: it also enables tool-scoping. An agent's `available_tools` list can include `invoke_agent` without the override capability, or include the override tool only for specific power-user / judge-orchestrator archetypes.
> > > **2. There's no companion model-discovery tool.** Without one, the LLM has to either remember model identifiers from training (unreliable — hallucinations are common) or be told them via the system prompt (brittle and stale). The override tool is barely usable without a way to discover valid identifiers in-band.
> > > Suggestion: add a `list_available_models` tool that returns the set of valid model names plus lightweight metadata (`type`, `provider`, `context_length`, `description`, `supported_settings`). Worth being deliberate about the projection — `ModelFactory.load_config()` returns the full model config including `custom_endpoint` blocks with `$ENV_VAR`-style API key references. Those shouldn't be exposed to the LLM. An explicit allowlist of safe fields keeps future config additions from silently leaking.
> > > Aside: your `BaseAgent.temporary_model_name_override()` context manager is genuinely a nice architectural seam — the override flows transparently to every deeper-stack reader of `get_model_name()`, which is cleaner than threading a local variable through.
> > 
> > 
> > hey @AndrewTilson , thanks for the feedback! And glad to hear that you're working on something similar.
> > good point about `model_name` getting provided on the default `invoke_agent` tool. I'll update this and split it into `invoke_agent` as the default path and `invoke_agent_with_model` for overrides.
> > do you feel strongly that `model_name` should be optional on `invoke_agent_with_model`? since calling `invoke_agent_with_model` without a model is basically just `invoke_agent`.
> > I’ll also add `list_available_models`, I was originally assuming callers would pass exact model names, but I agree that’s brittle.
> > long term I’d love to see support for model profiles like `cheap-and-fast`, `strong-reasoning`, `strong-architect`, etc., and let those resolve to configured aliases.
> > happy to coordinate with your implementation too if there’s a direction you’d prefer me to align with.
> 
> We can try model_name as required in the new version of the tool. Just need to test it out and see that it handles it ok. Regarding model profiles maybe another idea is to add a tagging option to models so it can be used like you mentioned along with standard tags like thinking, vision, etc.

I've updated the branch and split the override into `invoke_agent_with_model` with model_name being required and added `list_available_models`.  removed model_name entirely from `invoke_agent`.

I also ended up splitting `code_puppy/tools/agent_tools.py` into `code_puppy/tools/subagent_invocation.py` since it got pretty big with the model override

agreed on tags/profiles too.

let me know if you have any other suggestions.

AndrewTilson (approved):

AndrewTilson:
I have done initial smoke test's on this by creating a new evaluation agent and running through a couple model comparisons. The new tools look to function as expected.

PR #381, opened 2026-06-02
fix(agents): stop baking runtime metadata into cloned agents

weegens-aaron:
Cloning a class-based agent persisted get_full_system_prompt(), which froze runtime-only load_prompt fragments (live timestamp/CWD, memory recall, file-permission notes) and the per-instance identity ID into the static JSON definition. Reloaded clones then carried stale metadata and a second, conflicting identity.

Move load_prompt injection into BaseAgent.get_full_system_prompt so it is applied fresh at runtime for every agent (including JSON clones), keep get_system_prompt authored-only, and clone the authored prompt. Also drop the now-duplicate load_prompt append in the subagent path, which was double-injecting fragments for class-based agents.

Adds regression tests for authored-prompt purity and clone output.

### DIFF
--- a/code_puppy/agents/agent_code_puppy.py
+++ b/code_puppy/agents/agent_code_puppy.py
@@ -2,7 +2,6 @@
 
 from code_puppy.config import get_owner_name, get_puppy_name
 
-from .. import callbacks
 from .base_agent import BaseAgent
 
 
@@ -89,8 +88,9 @@ Important rules:
 {r["loop_rule"]}
 - Continue autonomously unless user input is definitively required
 """
-
-        prompt_additions = callbacks.on_load_prompt()
-        if len(prompt_additions):
-            result += "\n".join(prompt_additions)
+        # NOTE: runtime ``load_prompt`` fragments (plugin-injected notes such
+        # as environment context, file-permission rules, memory recall, ...)
+        # are intentionally NOT appended here — they're injected fresh at
+        # runtime by ``BaseAgent.get_full_system_prompt`` so they never get
+        # baked into a cloned/persisted agent definition.
         return result
--- a/code_puppy/agents/agent_manager.py
+++ b/code_puppy/agents/agent_manager.py
@@ -658,7 +658,12 @@ def clone_agent(agent_name: str) -> Optional[str]:
                     agent_instance.display_name, clone_index
                 ),
                 "description": agent_instance.description,
-                "system_prompt": agent_instance.get_full_system_prompt(),
+                # Persist the AUTHORED prompt only. ``get_full_system_prompt``
+                # would bake in runtime ``load_prompt`` fragments (live
+                # timestamp/CWD, kennel memory, ...) and the instance identity
+                # ID — all of which must be (re)injected fresh at runtime, not
+                # frozen into a static clone definition.
+                "system_prompt": agent_instance.get_system_prompt(),
                 "tools": _filter_available_tools(agent_instance.get_available_tools()),
             }
 
--- a/code_puppy/agents/agent_planning.py
+++ b/code_puppy/agents/agent_planning.py
@@ -2,7 +2,6 @@
 
 from code_puppy.config import get_puppy_name
 
-from .. import callbacks
 from .base_agent import BaseAgent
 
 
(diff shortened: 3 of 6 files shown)


asx8678:
#  Review — `fix/agent-clone`: clone no longer freezes runtime prompt fragments

> [!NOTE]
> **Verdict: Approve.** The fix is correct, well-contained, and—critically—does **not** regress any existing prompt-assembly path. It also removes a real double-injection bug and makes prompt assembly uniform across every agent. 

---

##  What this fixes

Cloning a class-based agent persisted `get_full_system_prompt()`, which froze **runtime-only** data into the static JSON definition:

-  live timestamp / CWD
-  kennel memory recall
-  file-permission notes
- 🆔 the per-instance identity ID

Reloaded clones then carried **stale metadata** *and* a **second, conflicting identity** (the frozen ID **plus** a freshly-injected one).

**The approach:** centralize runtime injection in `BaseAgent.get_full_system_prompt`, keep `get_system_prompt` authored-only, and clone only the authored prompt.

---

##  Claim-by-claim verification

| # | Claim | Status | Evidence |
|---|-------|:------:|----------|
| 1 | `load_prompt` injection moved into `BaseAgent.get_full_system_prompt` (fresh at runtime for every agent, incl. JSON clones) | &#10004; | `base_agent.py:130-148` |
| 2 | `get_system_prompt` kept authored-only | &#10004; | `agent_code_puppy.py:56`, `agent_planning.py:38`, `json_agent.py:106` |
| 3 | Clone persists the **authored** prompt only | &#10004; | `agent_manager.py:666` |
| 4 | Duplicate `load_prompt` append dropped from the subagent path | &#10004; | `agent_tools.py:404-417` |
| 5 | Regression tests for authored-prompt purity + clone output | &#10004; | `tests/agents/test_agents_remaining_coverage.py` |

---

<details>
<summary><b> Deep-dive: the "second identity" is structurally gone</b></summary>

Both runtime concerns now have **exactly one** production injection site:

- `on_load_prompt()` → only called at `base_agent.py:145`
- `get_identity_prompt()` → only called at `base_agent.py:148`

Since the authored prompt carries no identity, clones persist no identity, and a reloaded clone receives **exactly one** fresh ID. The double-identity path is no longer reachable.

</details>

<details>
<summary><b> Deep-dive: no production path drops or duplicates fragments</b></summary>

**Only two production callers invoke `get_system_prompt()` directly:**

- `base_agent.py:144` — inside the assembly method (intended)
- `agent_manager.py:666` — clone persistence (authored-only is the whole point)

**Every real prompt-to-model path goes through `get_full_system_prompt()`:**

- `_builder._assemble_instructions` → `_builder.py:333`
- `_runtime._should_prepend_system_prompt` → `_runtime.py:255`
- subagent path → `agent_tools.py:404`
- token accounting → `plugins/context_indicator/usage.py`, `base_agent._estimate_context_overhead`

**No agent bypasses injection.** The only production `def get_full_system_prompt` is `base_agent.py:130`; every agent (CodePuppy, planning, qa_kitten, creator, helios, JSONAgent, obsidian, shell_safety) inherits it. (Other definitions are test doubles under `tests/`.)

**No agent re-adds fragments/identity.** Read the bodies: none of the other `get_system_prompt` implementations call `on_load_prompt` or append identity, so there's no double-injection.

</details>

---

##  Behavior changes — both are improvements

- **Subagents fixed.** Class-based subagents previously got the fragment block **twice** (their `get_system_prompt` + the explicit append that was removed). Now once → shorter prompt, fewer wasted tokens.
- **Uniform across agents.** Every agent now receives file-permission rules + kennel memory through one path, instead of depending on whether a given agent baked them in.

---

##  Existing tests stay valid

- `tests/test_json_agents.py` — `JSONAgent.get_system_prompt()` always returned authored-only; unaffected.
- `tests/plugins/test_context_indicator_plugin.py` — the clamp test is mock-based and explicitly defensive ("guard against custom agents that override `get_system_prompt` and skip `on_load_prompt`").
- `usage.py` carves out and clamps the kennel block, so `/context` token accounting stays correct under the new layering.

### Added regression tests

- `test_code_puppy_authored_prompt_excludes_runtime_additions` — authored prompt contains **no** fragment and **no** identity (purity).
- `test_clone_class_agent_does_not_bake_runtime_metadata` — persisted clone JSON contains **no** `KENNEL-SECRET-BLOCK` and **no** `Your ID is` (clone output).
- `test_code_puppy_prompt_allows_callback_additions` — full prompt still includes fragments (no functional regression).

---

> [!WARNING]
> **Pre-existing caveat (NOT introduced by this PR).** For `prepend_system_to_user=True` models (claude-code, etc.) the full system prompt is folded into the first user message and stored in message history. A **saved/resumed** session therefore freezes that turn's fragments/identity. This was equally true before this change — it's a property of session persistence, not of this PR.

---

##  Optional follow-ups (non-blocking)

- [ ] Add an end-to-end test: clone → reload as `JSONAgent` → assert `get_full_system_prompt()` contains exactly **one** `Your ID is`.
- [ ] Note in the PR/docs that **legacy** clone JSONs created before this fix still carry stale data; re-cloning copies `system_prompt` verbatim (string-path branch in `agent_manager.py`).

---

##  Verdict

The change is contained, the injection is centralized to a single inherited method, no path drops or duplicates fragments/identity, and the test suite stays coherent. **Ship it.**

PR #392, opened 2026-06-09
feat: cap each AGENTS.md at 10,000 chars with truncation notice

AndrewTilson:
## Summary

Cap each `AGENTS.md` the CLI loads at a configurable character budget before
injecting it into the system prompt. **Defaults to 10,000 chars**, settable
via `/set agents_md_max_chars=<int>`. Over-cap files keep the first N chars
verbatim and get a labelled warning notice appended that names the source
file, the original character count, how many characters were dropped, and
hints at the `/set` knob so users can self-discover the override. The
notice is addressed at the agent so it surfaces the issue back to the user
on the next turn.

## Usage

From the TUI, set the per-file character cap with:

```
/set agents_md_max_chars=<int>
```

Where `<int>` is the maximum number of characters allowed per `AGENTS.md`
file. Tab-completion is wired up for the key. Examples:

| Command | Effect |
|---|---|
| *(not set)* | 10,000-char default applies |
| `/set agents_md_max_chars=20000` | Raise the cap to 20,000 chars |
| `/set agents_md_max_chars=500000` | Raise to 500,000 chars (for 1M-token-context models) |
| `/set agents_md_max_chars=2000` | Lower the cap to 2,000 chars |
| `/set agents_md_max_chars=0` (or any garbage) | Falls back to the 10,000 default |

The key name is `agents_md_max_chars`. The setting persists in
`~/.code_puppy/puppy.cfg` and applies to every subsequent `AGENTS.md`
load (global and project).

## Why

Users have started writing very large `AGENTS.md` files (kitchen-sink style)
that bloat the system prompt, increase token cost on every turn, and degrade
model attention to the rest of the prompt. A per-file cap keeps overhead
bounded without anyone having to think about it, and the configurable knob
lets users with large-context models (1M+ tokens) opt into bigger files
when it genuinely makes sense.

### `code_puppy/config.py`

- New constant `AGENTS_MD_MAX_CHARS_DEFAULT = 10_000`.
- New getter `get_agents_md_max_chars()` reads the `agents_md_max_chars`
  config key (settable via `/set agents_md_max_chars=<int>`). Falls back
  to the default on values that can't be a sensible cap — non-numeric,
  zero, or negative. **No upper clamp:** any positive integer is honoured
  verbatim so 1M-context-window users aren't artificially limited.
- New key registered in `get_config_keys()` so `/set` tab-completes it.

### `code_puppy/agents/_builder.py`

- Module re-exports `AGENTS_MD_MAX_CHARS = AGENTS_MD_MAX_CHARS_DEFAULT`
  for back-compat with anything that imported the original constant.
- Pure helper `_truncate_agents_md(content, source, max_chars)` —
  takes the cap as a parameter so it stays library-agnostic and easily
  testable. Under the cap returns `content` unchanged; over the cap
  returns `content[:max_chars]` plus a delimited warning that includes
  the source label, original length, dropped count, and the
  `/set agents_md_max_chars=<int>` hint.
- Tiny helper `_friendly_path(candidate)` renders the global config
  path as `~/...` rather than the absolute home path so `$HOME` doesn't
  leak into the system prompt.
- `load_puppy_rules()` resolves the cap once per call via
  `get_agents_md_max_chars()` and wraps each read (global, project
  `.code_puppy/`, project root fallback) with the truncation helper,
  independently per file. Per-file rather than post-join so a fat global
  `AGENTS.md` doesn't silently drop the smaller project file a developer
  is actively editing.

### Tests

- `tests/agents/test_builder_load_puppy_rules.py` — 31 tests:
  - `TestTruncation` covers the helper directly (under-limit, at-limit,
    over-limit), end-to-end through the loader (under, at, over,
    per-file isolation in both directions, both-files truncated, custom
    cap passed to helper, override raises the cap, override lowers the
    cap), and the `_friendly_path` helper.
  - `TestGetAgentsMdMaxChars` covers the new config getter:
    unset → default, valid integer, garbage → default, zero/negative →
    default, very-large-values-pass-through-uncapped (asserts 500,000
    and 99,999,999 are honoured verbatim — proves the no-ceiling
    contract), and presence in `get_config_keys()`.
- `tests/test_config.py` — two pre-existing `get_config_keys()`
  assertions updated to include the new `agents_md_max_chars` key.

## Behaviour matrix

| Scenario | Outcome |
|---|---|
| `AGENTS.md` ≤ cap | Loaded verbatim. No warning. |
| `AGENTS.md` > cap | First `cap` chars kept verbatim + warning appended. |
| Global oversized, project small | Global truncated, project intact. Single warning. |
| Project oversized, global small | Project truncated, global intact. Single warning. |
| Both oversized | Both truncated, two distinct warnings (each names its own file). |
| `/set agents_md_max_chars=N` (positive int) | Cap becomes `N`. No upper clamp. |
| `/set agents_md_max_chars=banana` (or `0`, `-1`) | Falls back to 10,000 default. |
| Key unset entirely | Cap is 10,000 default. |

## Verification

- 31/31 tests pass in `tests/agents/test_builder_load_puppy_rules.py`.
- 75/75 tests pass across `tests/agents/test_builder_load_puppy_rules.py`
  + `tests/test_config.py`.
- 612 tests collected across `tests/agents/` (full suite green locally).
- Validated end-to-end via headless `code-puppy --prompt ...`:
  - Default cap, 42,399-char `AGENTS.md`: agent quoted the notice
    citing `42,399` original / `32,399` dropped (matches
    `original − cap`), and surfaced the `/set` hint organically.
  - `/set agents_md_max_chars=20000`, 12,650-char `AGENTS.md`: agent
    confirmed "no truncation notice present" — content intact.
  - `/set agents_md_max_chars=2000`, 5,000-char `AGENTS.md`: agent
    extracted from the notice: cap=2,000, original=5,000, dropped=3,000.
  - `/set agents_md_max_chars=250000`, 220,000-char `AGENTS.md`
    (would have been clamped to 200k and truncated under any reasonable
    upper bound): agent confirmed "no truncation notice present" —
    proves the no-ceiling contract end-to-end.

## Design notes / non-goals

- **Configurable via `/set`, no CLI flag, no env var.** The TUI's
  `/set` mechanism is the canonical user-facing knob for runtime
  config; adding a CLI flag or env var would split the surface for no
  added power. Tab-completion is wired up via `get_config_keys()`.
- **No upper clamp on the override.** Earlier iterations had a
  200,000-char ceiling as a typo-guard, but 1M-token-context models
  (Gemini 1.5 Pro, Claude extended context, etc.) make any fixed
  ceiling short-sighted. Garbage / zero / negative still fall back to
  the default, so the footgun-guard for unintentional bad input is
  intact; the only thing removed is the cap on values that are clearly
  intentional and positive.
- **No `emit_warning` to the terminal.** The agent already gets the
  signal in its prompt and will tell the user. A second channel would
  be noise.
- **No exact-length assertions in the tests.** They pin structure
  (prefix-length, marker presence, source label, numeric counts) so the
  warning text can drift without test rewrites.
- **Truncation is at code-point boundary.** `len(str)` counts code
  points, so emoji-heavy files cap at code-point N with no
  partial-character mojibake (Python strings can't have partial code
  points).
- **Truncation notice self-documents the override.** The warning text
  ends with `(or raise the cap via /set agents_md_max_chars=<int>)`
  so the first time a user hits the cap they discover the knob
  organically — no need to read this PR or the docs.

## Downstream impact (free wins, no code changes)

- `plugins/context_indicator/usage.py` consumes `load_puppy_rules()` output,
  so its AGENTS.md token bucket now reflects the truncated size — a small
  accuracy improvement for the `/context` command.
- Sub-agents (`tools/subagent_invocation.py`) inherit the truncated content
  + warning, which is the right behaviour: they can also tell the user to
  trim (or raise the cap).
- Claude-code first-turn path (`_runtime.py`) prepends the same truncated
  content + warning.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -27,7 +27,9 @@ from code_puppy.callbacks import (
     on_wrap_pydantic_agent,
 )
 from code_puppy.config import (
+    AGENTS_MD_MAX_CHARS_DEFAULT,
     CONFIG_DIR,
+    get_agents_md_max_chars,
     get_global_model_name,
     get_value,
 )
@@ -38,13 +40,13 @@ from code_puppy.model_factory import ModelFactory, make_model_settings
 _AGENT_RULE_FILES = ("AGENTS.md", "AGENT.md", "agents.md", "agent.md")
 _CODE_PUPPY_DIR = ".code_puppy"
 
-# Hard cap on the characters from a single AGENTS.md file that get injected
-# into the system prompt. Users with kitchen-sink AGENTS.md files were
-# bloating context and degrading model attention; this keeps overhead
-# bounded without taking a CLI flag. Per-file (not post-join) so a fat
-# global doesn't silently nuke the project file a developer is actively
-# editing. Bump if you really must — but consider trimming AGENTS.md first.
-AGENTS_MD_MAX_CHARS = 10_000
+# Re-export the default so callers that imported AGENTS_MD_MAX_CHARS from
+# here keep working. The *effective* cap on any given load is whatever
+# ``get_agents_md_max_chars()`` returns (user override via
+# ``/set agents_md_max_chars=<int>``); this constant is just the fallback
+# documented in the warning notice and used by tests that don't care about
+# the override path.
+AGENTS_MD_MAX_CHARS = AGENTS_MD_MAX_CHARS_DEFAULT
 
 
 def _friendly_path(candidate: Path) -> str:
@@ -59,29 +61,32 @@ def _friendly_path(candidate: Path) -> str:
         return str(candidate)
 
 
-def _truncate_agents_md(content: str, source: str) -> str:
-    """Cap one AGENTS.md file at ``AGENTS_MD_MAX_CHARS`` with a labelled notice.
+def _truncate_agents_md(content: str, source: str, max_chars: int) -> str:
+    """Cap one AGENTS.md file at ``max_chars`` with a labelled notice.
 
     Returns ``content`` unchanged when it's within the cap. When it overflows,
-    keeps exactly the first ``AGENTS_MD_MAX_CHARS`` characters of the original
-    and appends a delimited warning addressed to the agent (so the agent can
+    keeps exactly the first ``max_chars`` characters of the original and
+    appends a delimited warning addressed to the agent (so the agent can
     surface it to the user on the next turn). ``source`` is a human-readable
     label for the file — used in the warning so the agent can tell the user
-    which specific file to trim when multiple files overflow.
+    which specific file to trim when multiple files overflow. ``max_chars``
+    is resolved once per load by the caller (see ``get_agents_md_max_chars``)
+    so a session-wide ``/set`` override is honoured.
     """
     original_len = len(content)
-    if original_len <= AGENTS_MD_MAX_CHARS:
+    if original_len <= max_chars:
         return content
-    dropped = original_len - AGENTS_MD_MAX_CHARS
+    dropped = original_len - max_chars
     notice = (
         f"\n\n--- AGENTS.md truncated ---\n"
         f"The {source} content was truncated: original was "
         f"{original_len:,} chars, {dropped:,} chars dropped. Please tell "
-        f"the user to trim {source} below {AGENTS_MD_MAX_CHARS:,} "
-        f"characters so the full rules can take effect.\n"
+        f"the user to trim {source} below {max_chars:,} "
+        f"characters so the full rules can take effect (or raise the cap "
+        f"via `/set agents_md_max_chars=<int>`).\n"
         f"--- end truncation notice ---"
     )
-    return content[:AGENTS_MD_MAX_CHARS] + notice
+    return content[:max_chars] + notice
 
 
 def load_puppy_rules() -> Optional[str]:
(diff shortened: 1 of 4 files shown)


AndrewTilson:
Two follow-up commits pushed:

- `395ad4f` — `style:` ruff format fix for the failing quality check (two function signatures that ruff prefers on one line).
- `4203ea2` — `feat:` promote the 10,000-char cap to a `/set`-controllable config knob (`agents_md_max_chars`). Default is still 10,000 — no behaviour change for anyone who doesn't set it. Hard-clamped to `[1, 200_000]` so a typo can't blow up the system prompt. New tests cover the getter (default / valid / garbage / zero-or-negative / ceiling-clamp / present-in-`get_config_keys`), the helper's caller-provided cap, and end-to-end raise+lower override paths through `load_puppy_rules()`. Updated two pre-existing config-keys-list assertions in `tests/test_config.py` for the new key.

Validated end-to-end with the internal CLI: with a 15k AGENTS.md and `agents_md_max_chars=20000` set, the agent confirmed no truncation notice present and all content intact; with the cap reset, the agent quoted the notice citing the correct dropped-char count.

AndrewTilson:
Follow-up: `501de22` removes the 200k upper clamp. With 1M-token context models becoming real, a hard ceiling on AGENTS.md size was short-sighted. Garbage / zero / negative still fall back to the 10k default; any positive integer is now honoured verbatim. Test `test_above_ceiling_is_clamped` was replaced with `test_very_large_values_pass_through_uncapped` asserting that `500_000` and `99_999_999` survive the getter intact.

PR #465, opened 2026-06-10
feat(plugins): optional yantrik_memory — semantic memory + supersession (opt-in)

spranab:
## What this adds

An **optional, off-by-default** plugin (`code_puppy/plugins/yantrik_memory/`) that
adds a YantrikDB-backed semantic/learning memory layer **alongside** `puppy_kennel`
— not a replacement. Heads-up discussion: #464.

It mirrors the kennel's plugin contract (`load_prompt`, `user_prompt_submit`,
`agent_run_end`, `register_tools`, `/yantrik` command), and:

- **distills** durable facts/preferences from natural conversation (ignores chatter),
- **supersedes** outdated facts via `correct()` instead of accumulating them,
- surfaces **query-relevant** memories passively each turn,
- reinforces options that keep getting accepted.

## Why (honest framing)

The kennel's P0 prefs band already handles basic fact recall well — so on "is the
fact recalled," a prefs-band baseline **ties** this; I'm **not** claiming better
recall. What an append-only prefs store *can't* do, and this adds:

- **Supersession** — after "we rebranded blue→green", the kennel's P0 keeps **both**
  (it's append-only), so the agent gets contradictory context; this drops the stale
  value. Verified stable from 1k→10k memories (the append-only band still carries
  both at 10k; this carries only green).
- **Passive semantic recall** — the kennel's passive layer is "all prefs + recent";
  finding the right past artifact otherwise relies on recency or the agent calling
  the BM25 search tool.

## Safety / footprint

- **Opt-in**: default OFF; enable with `/yantrik enable`.
- **Fail-soft**: if `yantrikdb` isn't installed (or it's disabled), the plugin is
  inert — registers no callbacks, never crashes boot or a turn.
- **Zero changes to existing code** — a new plugin directory only.
- One optional dependency (`yantrikdb`); embedder is configurable down to a bundled
  no-ONNX option.

## Testing

- **Integration test** (`test_integration.py`): distills facts, supersession works
  (recall's current band = green, stale blue gone), chatter adds nothing.
- **Live load test**: loads cleanly alongside all 35 builtin plugins (incl. the
  kennel); via the real callback bus, `on_user_prompt_submit` distills and
  `on_load_prompt` injects the correct (superseded) recall block.

## Known limitations (full disclosure)

- The distiller runs a local-LLM call + a short flush per user turn — correct but
  not free; worth making async. Optional and model-configurable.
- Not yet run in a full interactive TUI session (validated via the real loader +
  callback dispatchers).
- YantrikDB's automatic conflict detection doesn't currently fire on free-text
  attribute-value updates, so supersession uses an explicit `correct()` at the app
  layer (the distiller detects the update). The engine side is being addressed
  separately.

Happy to adjust to your plugin/dependency conventions — flagging this as a starting
point for discussion as much as a merge candidate.

### DIFF
--- a/code_puppy/plugins/yantrik_memory/.gitignore
+++ b/code_puppy/plugins/yantrik_memory/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc
--- a/code_puppy/plugins/yantrik_memory/__init__.py
+++ b/code_puppy/plugins/yantrik_memory/__init__.py
@@ -1,0 +1,19 @@
+"""Yantrik Memory — a YantrikDB-backed semantic/learning memory for Code Puppy.
+
+Where the ``puppy_kennel`` plugin stores *verbatim* drawers in SQLite + FTS5,
+this plugin layers a **learning** memory on top of YantrikDB:
+
+* Every user turn is logged verbatim as an ``episodic`` memory.
+* A local distiller (Ollama) reads the turn and extracts the durable facts /
+  preferences worth keeping, writing them as ``semantic`` memories.
+* When a new turn *updates* a prior fact, the old fact is superseded via
+  ``correct()`` so only the authoritative value surfaces in recall.
+
+Recall is **banded**: a "current" band of authoritative facts/prefs (always
+surfaced into the system prompt) and a "history" band of query-relevant
+episodic context.
+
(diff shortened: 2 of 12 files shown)


thomwebb:

thomwebb:

PR #480, opened 2026-06-13
feat(paw_sign): append puppy signature to code-puppy git commits

dudeinthemirror:
## Summary

Adds a builtin plugin that paw-signs commit messages composed by code-puppy. When code-puppy runs a `git commit -m "..."`, the plugin tacks on an extra `-m` paragraph:

```
[-- paw-signed: code-puppy  --]
```

So the audience knows the commit came from a very good boy.

## How it works

Hooks `pre_tool_call` and mutates the `agent_run_shell_command` args in place (same pattern as the `emoji_filter` plugin). Adds the signature as a second `-m`, which git stacks as a trailer-style paragraph at the bottom of the message.

## Safety / design

A corrupted `git commit` is far worse than a missing paw print, so the detector is deliberately conservative:

- **Analyse, don't rebuild.** Uses `shlex` only to inspect the command (respecting quotes), never to reconstruct it — so shell substitutions `$(...)`, heredocs, and exotic quoting in the original are preserved untouched.
- **Only signs when the `git commit -m` is the LAST segment** in the chain. This means `git add . && git commit -m "x"` gets signed, but `git commit -m "x" && git push` is left alone (no paw landing on `push`).
- **Idempotent** — already-signed commands are skipped (no double paw).
- **Skips editor commits** (no `-m`) and anything `shlex` can't parse cleanly.
- **Never blocks or crashes** — exceptions are swallowed.
- The paw glyph is stored as a `\\U0001F43E` unicode escape so the repo's `emoji_filter` doesn't strip it from source on write.

## Testing

- 18 unit tests covering happy path, conservative no-op cases, and helper internals — all pass.
- Verified no regressions: full plugins suite **2357 passed, 5 skipped**, plus plugin-loader/callback tests **122 passed**.

## Config

No toggle — always on, keep it simple (YAGNI).

### DIFF
--- a/code_puppy/plugins/paw_sign/__init__.py
+++ b/code_puppy/plugins/paw_sign/__init__.py
@@ -1,0 +1,7 @@
+"""Paw-sign plugin.
+
+Appends a playful puppy-themed signature to commit messages that
+code-puppy composes, so the audience knows the commit came from a very
+good boy. Intercepts ``git commit -m`` shell commands via the
+``pre_tool_call`` hook and tacks on an extra ``-m`` paragraph.
+"""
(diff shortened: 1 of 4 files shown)


thomwebb:

thomwebb:

PR #491, opened 2026-06-19
fix: stop a fast Ctrl+C double-tap from exiting the whole REPL

JulienEllie:
## Problem

While a task is running, pressing **Ctrl+C twice in quick succession** exits the entire process. This is especially annoying because Ctrl+C is *the* way to stop the running command/agent — so an over-eager double-tap kills your whole session instead of just cancelling the turn.

Intended behavior:
- **Ctrl+C #1:** cancel the running shell command.
- **Ctrl+C #2:** cancel the agent's thinking/inference.
- **Ctrl+D:** quit.

Process exit should never happen from Ctrl+C.

## Root cause

Ctrl+C handling juggles **three** different `SIGINT` handlers depending on state:

| State | SIGINT handler | Effect |
|---|---|---|
| Idle / between turns | Python default | **raises `KeyboardInterrupt`** |
| Agent run (`_runtime.py`) | `keyboard_interrupt_handler` | schedules a task cancel (never raises) |
| Shell command (`command_runner.py`) | `_shell_sigint_handler` | kills the shell |

The per-run and per-shell handlers **save and restore** whatever handler preceded them. The bug lives in the *gap between handlers*:

1. Ctrl+C #1 cancels the agent task → `run_with_mcp`'s `finally` **restores the previous handler**, which between turns is Python's **default**.
2. Ctrl+C #2 lands in that brief unwind window → hits the **default** handler → raises a bare `KeyboardInterrupt`.
3. `KeyboardInterrupt` is a `BaseException`, so the interactive loop's `except Exception` doesn't catch it → it bubbles to `main_entry`'s `except KeyboardInterrupt: return 0` → **the whole process exits.**

Slow taps work because by the time the second one arrives, the run/shell handler is back in place. Fast taps slip through the crack.

## Fix

1. **Session-wide baseline SIGINT guard.** `interactive_mode` installs a benign handler for the lifetime of the REPL that *swallows* Ctrl+C in the gaps. Now the saved/restored "original" handler is never the process-killing default, so a stray Ctrl+C in any unwind window is a no-op instead of an exit. The per-run/per-shell handlers still own real cancellation while they're active.
2. **Defense-in-depth `except KeyboardInterrupt`** in the turn body — if one ever sneaks through, it's treated as a turn cancel and the loop continues. Ctrl+D remains the only way out.

## Verification

`tests/test_ctrl_c_guard.py`:
- The guard returns `None` and never raises.
- A **real delivered SIGINT** (`os.kill(getpid(), SIGINT)`) is swallowed with the guard installed.
- **Control test:** the same SIGINT *does* raise `KeyboardInterrupt` under the default handler — proving the guard is what fixes it.

All existing `cli_runner` tests pass (45), `ruff format --check .` clean, `ruff check` clean.

## Scope

- `code_puppy/cli_runner.py` — add `_interactive_sigint_guard`, install it in `interactive_mode`, add the turn-body `KeyboardInterrupt` catch.
- `tests/test_ctrl_c_guard.py` (new).

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -11,6 +11,7 @@ apply_all_patches()
 import argparse
 import asyncio
 import os
+import signal
 import sys
 import traceback
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


mpfaffenberger:
Why closed? I can't reproduce this on my end but I have seen stuff like this before once in awhile.

JulienEllie:
Cause code puppy was too eager and opened the PR before I told it to do so and it had run all the tests!  Now it has, re-opened.

JulienEllie:
Fixing. Timing issue too, my test bled into another test.

PR #504, opened 2026-06-25
fix: drain stdin on turn cancel so the next prompt isn't truncated

JulienEllie:
## The bug

Cancel a turn with **Ctrl+C**, immediately start typing your next instruction, and the agent receives a **truncated, corrupted prompt** — the leading characters of what you typed silently vanish.

### Why it happens

Every agent turn, `run_with_mcp` spawns a POSIX key-listener thread (`agents/_key_listeners.py`) that:

1. Puts the terminal into **cbreak mode** (`tty.setcbreak`)
2. Sits in a `select` loop reading stdin **one byte at a time** (`stdin.read(1)`), watching for Ctrl+X (shell cancel) and Ctrl+T (pause/steer).

cbreak leaves `ISIG` on, so Ctrl+C still fires SIGINT → cancels the agent. The problem is the **gap between your Ctrl+C and that thread actually dying**:

1. Ctrl+C → SIGINT → schedules cancel → `CancelledError` → `finally` runs `handle.stop()` (just *sets an event*) then `thread.join(timeout=1.0)`.
2. The listener only checks the stop event at the top of its `select(..., 0.05)` loop — it doesn't notice instantly.
3. **In that window you're already typing your follow-up** (that's the whole cancel-and-retype workflow). The still-alive listener keeps doing `stdin.read(1)`, **swallowing your leading characters one by one** and discarding them (they don't match Ctrl+X/Ctrl+T).
4. The thread finally exits and restores termios. Only the characters typed *after* that moment survive in the tty buffer.
5. `prompt_async` reads the **surviving tail**, you hit Enter → a **truncated instruction is submitted to the agent.**

This lines up with the observed symptoms:
- The number of dropped chars is *time*-based (how long the async cancel takes to unwind × how fast you type), so it *feels* like a hidden, prompt-length-shaped limit.
- A throwaway empty Enter "fixes" it — by the second prompt the listener thread is long dead and termios is restored.
- It only ever happens on cancel-mid-turn, the only time the cbreak listener is alive *while you're typing fresh input*.

The genuinely nasty part: it's not just cosmetic — a **partial, corrupted instruction actually reaches the agent**, which then confidently acts on half a sentence. Silent data corruption.

## The fix

Drain the terminal's input queue on the cancel path, **after** the listener thread is fully stopped (joined) and termios is restored, so the next prompt starts from a clean slate instead of a half-eaten one.

- New `agents/_key_listeners.flush_stdin_input()` — POSIX `termios.tcflush(TCIFLUSH)` / Windows `msvcrt` drain. TTY-guarded, best-effort, no-ops when stdin isn't a TTY or is unavailable.
- `run_with_mcp` tracks a `run_cancelled` flag and calls the flush in its `finally` **only on cancel**. The happy path is completely untouched — no risk of nuking legitimate input.

### Why discard rather than preserve?

Bytes the listener already `read(1)`'d are gone — they can't be un-read, so true type-ahead preservation isn't possible here. And type-ahead during a turn was already non-functional anyway. So flushing is a pure win: it converts a **silent partial-submit** into a **predictable clean prompt**, which matches the mental model you already have when you hit Ctrl+C ("stop, let me rethink").

## What we added

| File | Change |
|---|---|
| `code_puppy/agents/_key_listeners.py` | New `flush_stdin_input()` helper (+ `__all__` export) |
| `code_puppy/agents/_runtime.py` | Track `run_cancelled`; call the flush in `run_with_mcp`'s `finally`, cancel-path only, after the listener join |
| `tests/test_cancel_stdin_flush.py` | New tests |

### Tests

Uses a **real pseudo-terminal** so `termios.tcflush` genuinely runs — no mocking the primitive under test:

- `test_flush_drains_pending_input` — queued bytes are gone after the flush
- `test_without_flush_input_survives` — control proving the harness really buffers input
- `test_flush_noops_on_non_tty` — pipe/non-TTY stdin is left untouched, no raise
- `test_flush_survives_missing_stdin` — `stdin = None` degrades to a silent no-op

All green locally, plus the existing ctrl-c/cancel suites (`9 passed, 6 skipped`). `ruff check .` and `ruff format --check .` clean repo-wide.

### DIFF
--- a/code_puppy/agents/_key_listeners.py
+++ b/code_puppy/agents/_key_listeners.py
@@ -415,6 +415,56 @@ def _listen_posix(
 
 
 # =============================================================================
+# Stdin drain (cancel cleanup)
+# =============================================================================
+#
+# When a turn is cancelled (Ctrl+C), the listener thread is still sitting in
+# cbreak mode doing ``stdin.read(1)`` right up until it observes the stop
+# event. Any follow-up instruction the user starts typing in that unwind
+# window gets eaten one byte at a time and silently discarded, leaving a
+# *truncated* tail that prompt_toolkit then submits as a corrupt prompt.
+# After the listener is fully stopped (thread joined, termios restored), we
+# drain whatever is left in the tty input queue so the next prompt starts
+# from a clean slate instead of a half-typed mess. Best-effort and TTY-only.
+
+
+def flush_stdin_input() -> None:
+    """Discard any pending bytes in the terminal's input queue.
+
+    Safe to call unconditionally: no-ops when stdin isn't a TTY or the
+    platform primitives are unavailable. Intended for the cancel path,
+    where leftover keystrokes are guaranteed-corrupt rather than useful.
+    """
+    import sys
+
+    stdin = getattr(sys, "stdin", None)
+    if stdin is None or not hasattr(stdin, "isatty"):
+        return
+    try:
+        if not stdin.isatty():
+            return
+    except Exception:
+        return
+
+    if sys.platform.startswith("win"):
+        try:
+            import msvcrt
+
+            while msvcrt.kbhit():
+                msvcrt.getwch()
+        except Exception:
+            pass
+        return
+
+    try:
+        import termios
+
+        termios.tcflush(stdin.fileno(), termios.TCIFLUSH)
+    except Exception:
+        pass
+
+
+# =============================================================================
 # Reentrant suspend context manager
 # =============================================================================
 #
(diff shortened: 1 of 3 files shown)


JulienEllie:
Let's not merge yet. This isn't a full fix unfortunately.

mpfaffenberger (approved):
omg love you

mpfaffenberger:
Why draft? This looked straightforward.
edit: nvm saw your comment.

PR #507, opened 2026-06-26
feat(session): unified autosave store + headless `-r NAME` save-back

AndrewTilson:
## Summary

Consolidates session storage into a single canonical store, adds headless `-r NAME -p ...` save-back, and migrates any legacy `contexts/` directory in one shot.

This was originally developed against an internal fork; opening here because it's pure-core functionality with zero internal dependencies.

### Unified autosave store
Replaces the previous `contexts/` (manual `/dump_context`) + `autosaves/` (periodic) split with a single canonical store at `AUTOSAVE_DIR` (`~/.code_puppy/autosaves/`). All of `-r NAME`, `/dump_context`, `/load_context`, `/autosave_load`, and periodic autosave now read/write the same directory. Pre-existing `contexts/*.pkl` are swept into `autosaves/` once per install (sentinel-gated, atomic moves via `os.replace`, asymmetric-collision guard).

### New `session_lifecycle` module
Thin orchestration layer between pure-I/O `session_storage` and callers. Exposes:
- `is_valid_session_name(name, *, allow_reserved_prefix=False)` — slug validator (`^[A-Za-z0-9._-]{1,128}$`, rejects all-dot names and reserved `auto_session_` prefix unless the caller is the singleton itself)
- `persist_named_session(...)` — wraps `save_session` with the right `auto_saved` flag + success message; deliberately does NOT fire `post_autosave` (reserved for periodic background autosave)
- `resolve_or_create_resume_target(...)` — five-branch resume target resolution: existing `.pkl` path, existing `<sessions_dir>/<name>.pkl`, any path that exists, bare slug normalisation (strips `.pkl` suffix), lazy-create when allowed
- `create_empty_session(...)` — materialises an empty named session through the same write path
- `fire_post_autosave_callback(...)` — `ThreadPoolExecutor` wrap so the sync hook trigger can reach async plugin callbacks even when called from inside a running event loop

### Headless `-r NAME -p ...` save-back
`execute_single_prompt` now persists the completed turn back to the named session in a `finally` block, so cancellation and exceptions still save partial state. Shell-passthrough (`!cmd`) is explicitly guarded — `-r mywork -p '!ls'` will not clobber the existing session bytes. `agent.set_message_history(list(result.all_messages()))` is called before save so the assistant reply lands in the pickle (not just the user prompt).

### Singleton rename + backwards-compatible deprecation
The autosave singleton is renamed from `_CURRENT_AUTOSAVE_ID` to a full-filename store. New API:
- `get_current_session_name()` / `rotate_session_name()` / `pin_current_session_name(name)`

Old API kept as deprecated shims with `DeprecationWarning` for one release:
- `get_current_autosave_id()` (strips `auto_session_` prefix to preserve return shape)
- `rotate_autosave_id()`
- `get_current_autosave_session_name()`
- `set_current_autosave_from_session_name(name)` — now raises `ValueError` for previously-silent-accepted invalid input (control chars, empty string, path separators). Worth a CHANGELOG note if external plugins use it.

All internal call sites have migrated to the new names.

### Autosave menu UX split
Named sessions appear in their own labelled section above auto-flavored entries; both groups sorted mtime-desc. Section headers are cosmetic and do NOT consume entry indices, so existing pagination arithmetic is untouched.

### Intentional asymmetry, documented
`/load_context <name>` and `-r NAME` behave differently *by design*:
- `/load_context` is a **snapshot verb** (think `pg_restore`): rotates the active autosave so the loaded snapshot stays frozen on disk
- `-r NAME` is a **continuation verb**: pins the session so subsequent saves land back in `<NAME>.pkl`

The asymmetry is explained inline at `handle_load_context_command` with the git archaeology to back it up (commit `cc04629b`).

## Test coverage

Adds ~1,250 lines across:

- `tests/test_session_lifecycle.py` — slug validator (positive + traversal/whitespace/reserved-prefix rejection), `create_empty_session` writes both pickle + metadata, `resolve_or_create_resume_target` branch coverage, `fire_post_autosave_callback` async-from-sync paths
- `tests/test_session_migration.py` — sweep moves both pickle + sidecar, idempotent via sentinel, asymmetric-collision skip, error logging for sidecar orphans
- `tests/test_cli_runner.py::TestHeadlessSessionPersistence` — save on success, save on cancel, save on generic exception, no save when `session_name is None`, shell-passthrough does NOT clobber existing session (byte-identical)
- `tests/test_session_commands.py` — `/dump_context` slug validation (traversal/reserved-prefix rejection)
- `tests/test_autosave_menu.py` — section header ordering + index arithmetic

## Validation

Headless end-to-end validation matrix (offline harness driving the real `execute_single_prompt` + `main()` resume block with only the agent + `run_prompt_with_attachments` stubbed):

| Scenario | Result |
|---|---|
| `-p '...'` (no `-r`) is disk-silent | PASS |
| `-r brand-new -p '...'` lazy-creates + writes turn | PASS |
| `-r mywork -p '...'` on existing appends turn | PASS |
| `-r foo.pkl -p '...'` normalises (no `.pkl.pkl`) | PASS (after fix) |
| `-r /abs/path/foo.pkl -p '...'` doesn't escape AUTOSAVE_DIR | PASS (after fix) |
| `-r mywork -p '!ls'` shell-passthrough guard works | PASS |
| Cancel mid-`-r NAME -p '...'` persists partial history | PASS |
| `-r auto_session_*` lazy-create rejected | PASS |
| `-r auto_session_existing` resume still works | PASS |
| `-r ../../etc/passwd` rejected | PASS |
| Legacy contexts/ sweep on startup | PASS |
| Sweep skips on name collision | PASS |
| Sweep idempotent via sentinel | PASS |
| Save-back failure emits error, doesn't propagate | PASS |
| Successive `-r foo -p '...'` calls compound history | PASS |

## Files

```
 README.md (skipped — public README structure differs; deferred to a docs PR)
 code_puppy/cli_runner.py
 code_puppy/command_line/autosave_menu.py
 code_puppy/command_line/session_commands.py
 code_puppy/config.py
 code_puppy/plugins/switch_agent_resume/register_callbacks.py
 code_puppy/session_lifecycle.py             (new)
 code_puppy/session_migration.py             (new)
 code_puppy/session_storage.py
 tests/command_line/test_autosave_menu.py
 tests/command_line/test_session_commands.py
 tests/plugins/test_switch_agent_resume.py
 tests/test_auto_save_session.py
 tests/test_cli_runner.py
 tests/test_config.py
 tests/test_session_lifecycle.py             (new)
 tests/test_session_migration.py             (new)
 tests/test_session_storage_coverage.py
 tests/test_shell_passthrough.py
```

18 files, +2403 / −263.

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -27,7 +27,7 @@ from code_puppy.config import (
     COMMAND_HISTORY_FILE,
     ensure_config_exists,
     finalize_autosave_session,
-    get_current_autosave_session_name,
+    get_current_session_name,
     initialize_command_history_file,
     record_terminal_session,
     save_command_to_history,
@@ -80,57 +80,6 @@ def _render_turn_exception(exc: Exception) -> None:
     from code_puppy.messaging.queue_console import get_queue_console
 
     get_queue_console().print_exception()
-
-
-def _resume_session_from_path(raw_path: str) -> None:
-    """Restore agent message history from a saved .pkl session file.
-
-    Accepts any path (autosaves, contexts, somewhere weird on disk). We don't
-    care where it lives — we just decompose into (parent_dir, stem) and reuse
-    ``session_storage.load_session`` so we stay DRY.
-    """
-    from code_puppy.agents.agent_manager import get_current_agent
-    from code_puppy.messaging import emit_error, emit_success
-    from code_puppy.session_storage import load_session
-
-    session_path = Path(raw_path).expanduser().resolve()
-
-    if not session_path.exists():
-        emit_error(f"--resume: session file not found: {session_path}")
-        sys.exit(1)
-
-    if session_path.suffix != ".pkl":
-        emit_error(
-            f"--resume: expected a .pkl session file, got '{session_path.suffix}': {session_path}"
-        )
-        sys.exit(1)
-
-    try:
-        history = load_session(session_path.stem, session_path.parent)
-    except Exception as exc:
-        emit_error(f"--resume: failed to load session: {exc}")
-        sys.exit(1)
-
-    try:
-        agent = get_current_agent()
-        agent.set_message_history(history)
-    except Exception as exc:
-        emit_error(f"--resume: failed to attach history to agent: {exc}")
-        sys.exit(1)
-
-    # Rotate autosave id so we don't clobber the original file we just resumed.
-    try:
-        from code_puppy.config import rotate_autosave_id
-
-        rotate_autosave_id()
-    except Exception:
-        pass  # autosave rotation is best-effort
-
-    total_tokens = sum(agent.estimate_tokens_for_message(m) for m in history)
-    emit_success(
-        f"✅ Resumed session: {len(history)} messages ({total_tokens} tokens)\n"
-        f"📁 From: {session_path}"
-    )
 
 
 def apply_quick_resume(args) -> bool:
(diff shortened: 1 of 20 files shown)


kvandre12-commits:
│ Hey — this might overlap with an older fork/workstream I had around the
│ autosave/session-resume seam.
│ I’m not claiming this PR came from my code directly, but the area looks
very
│ familiar: AUTOSAVE_DIR, session IDs, /dump_context, /load_context,
autosave
│ menu/resume behavior, and shell passthrough interactions.
│ For reference, the older fork work I’m thinking of included prior commits
│ around:
│ - autosave context
│ - extracted session_storage
│ - stable autosave IDs
│ - autosave rotation
│ - switch-agent resume
│ - autosave menu improvements
│
│ If useful, I’m happy to point at the old fork/commits so you can compare
│ lineage or just ignore if this was convergent evolution.

AndrewTilson:
Appreciate you flagging it! Quick context on where this PR came from: it's our own work, not templated off another branch or fork.

The motivation was user feedback — someone tried running headless mode (`-r NAME -p '...'`) for a multi-turn workflow and noticed turns weren't actually persisting between invocations. When we dug into it, we found headless and interactive sessions were going through different storage paths (interactive wrote to `autosaves/`, headless wrote to `contexts/`), and the two never met. That's what drove the refactor: unify the storage so both flows share one source of truth, add the headless save-back, and include a one-shot migration sweep so existing users don't lose their old `contexts/` files.

For context on the rest of the surface area: the underlying autosave machinery (`AUTOSAVE_DIR`, `session_storage`, stable autosave IDs, autosave rotation, the menu, `/dump_context`, `/load_context`) is all pre-existing in this repo from @mpfaffenberger's work in Oct–Nov 2025 (the PR body cites `cc04629b` for the design archaeology), and the `switch_agent_resume` plugin is @sskmlm's from May 2026. This PR builds on top of all of that rather than introducing those concepts.

WSxDemise (approved):

kvandre12-commits:
│ Thanks for the clarification — that makes sense. Helpful to have the
│ provenance and the bug-driven motivation spelled out.

PR #514, opened 2026-06-29
feat(security): forbid agents from suggesting Snyk-finding suppression

AndrewTilson:
## Problem
Walmart Security has flagged that Code Puppy is openly recommending users add `.snyk` ignore entries or scan-exclusion config to "resolve" Snyk findings. That is circumvention of security controls, not remediation, and it's already showing up in real user transcripts.

## Change
Adds a terse `SECURITY_POLICY_PROMPT` constant to `code_puppy/agents/base_agent.py` and appends it inside `BaseAgent.get_full_system_prompt()` after the `load_prompt` plugin fragments and before the identity tail. Every agent — built-ins, `JSONAgent` (user-defined), and any sub-agent — is now instructed to never recommend or produce Snyk-finding suppressions and to always propose a real fix (patch the code, upgrade the dep) or escalate to security for a documented exception.

The wording is deliberately terse (~30 tokens). At 60K+ users every token added to the base system prompt multiplies.

## Why the base class, not a plugin?
This is a non-negotiable security guarantee. Plugins can be disabled, fail to register, or be missing in a custom build. The chokepoint at `BaseAgent.get_full_system_prompt()` is the only place that is guaranteed to fire for every agent (built-in, JSON, fork).

## Cloning property worth noting
`agent_manager.py` deliberately persists only the *authored* `get_system_prompt()` when cloning an agent to JSON — so the security text is never baked into a clone. The cloned agent re-enters `get_full_system_prompt()` when it runs, so the policy still applies. One source of truth, no staleness.

## Out of scope (intentional)
Non-`BaseAgent` `pydantic_ai.Agent(...)` constructions:
| File | What | Why out of scope |
|---|---|---|
| `code_puppy/summarization_agent.py` | History summariser | Rewrites history, never advises the user. |
| `code_puppy/plugins/wiggum/judge.py` | Eval judge | Scores responses, never advises. |
| `code_puppy/agents/_builder.py` (tool probe) | Token-count probe | Never sent to a model for real responses. |
| `code_puppy/agents/_steer_processor.py` | Steer processor | Inherits parent run's instructions. |
| `code_puppy/plugins/dbos_durable_exec/wrapper.py` | DBOS wrapper | Wraps existing agent, no new instructions. |

Also intentionally out of scope: extending the prohibition to `# nosec`, `// nolint`, Sonar suppressions, etc. Those have legitimate uses (e.g. known false-positive in a unit test) that a blanket prohibition would produce wrong advice for. Easy one-line edit later if needed.

## Tests
New file `tests/agents/test_base_agent_security_policy.py`:
- `_StubAgent(BaseAgent)` minimal subclass → asserts the policy reaches the chokepoint for any `BaseAgent` subclass.
- `JSONAgent` built from tmp_path JSON → asserts the user-defined-agent path also gets the policy.
- Substring guard that pins `"Snyk"`, `"real fix"`, `"escalate"` in the constant so a future revision can't accidentally strip the escape-valve clause.

Local runs: `pytest tests/agents/test_base_agent_security_policy.py` → 3 passed. Regression sweep across surrounding base_agent / json_agent test files → 45 passed, no regressions.

---
_Walmart-internal tracking: PUP-378._

### DIFF
--- a/code_puppy/agents/base_agent.py
+++ b/code_puppy/agents/base_agent.py
@@ -42,7 +42,23 @@ from code_puppy.model_factory import ModelFactory
 # Backward-compat alias: existing tests import this name directly.
 should_retry_streaming_exception = should_retry_streaming
 
-__all__ = ["BaseAgent", "should_retry_streaming_exception"]
+# Non-negotiable security guarantee — appended to every agent's runtime
+# system prompt by ``BaseAgent.get_full_system_prompt``. Lives on the base
+# class chokepoint (not a plugin) so it cannot be silently disabled by a
+# missing/broken plugin or a custom build. Keep terse: this ships on every
+# system prompt of every call across 60k+ users, so every token multiplies.
+SECURITY_POLICY_PROMPT = (
+    "\n\nSecurity: Never recommend or produce suppressions for Snyk findings "
+    "(.snyk files, `snyk ignore`, scan-exclusion config). "
+    "Always propose a real fix (patch the code, upgrade the dep) "
+    "or escalate to security for a documented exception."
+)
+
+__all__ = [
+    "BaseAgent",
+    "SECURITY_POLICY_PROMPT",
+    "should_retry_streaming_exception",
+]
 
 
 def _extract_pydantic_agent_tools(pyd_agent: Any) -> Optional[Dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


AndrewTilson:
Closing — this turned out to be a Walmart-internal security policy ask, not appropriate for the public OSS base prompt. The Walmart fork will carry the rule. Sorry for the noise!

PR #517, opened 2026-07-01
fix(anthropic-thinking): route wire shape by model family (fixes HTTP 400 on classic Claude with adaptive)

AndrewTilson:
## Reported error

```
anthropic.APIStatusError: {error: {message: [HTTP 400] thinking: Input tag
'adaptive' found using 'type' does not match any of the expected tags:
'disabled', 'enabled', type: invalid_request_error}, type: error}
```

Hit on the `claude-4-5-sonnet` alias with default settings.

## Root cause (verified against `puppy-backend.walmart.com/anthropic`)

Anthropic-family Claude models bifurcate into two mutually-exclusive wire
shapes for the `thinking` field:

| Family | Models | Accepts | Rejects |
|---|---|---|---|
| **Classic** | Sonnet 4.5 & earlier, Haiku, Opus ≤ 4.5 | `type:"enabled"` + `budget_tokens`, `type:"disabled"` | `type:"adaptive"` (reporter's error) |
| **Adaptive** | Opus 4.6/4.7/4.8, Sonnet 4.6, Sonnet 5, Fable 5 | `type:"adaptive"` (+ optional `display:"summarized"`, `output_config.effort`) | `type:"enabled"` (`"thinking.type.enabled" is not supported for this model. Use adaptive.`) |

Two bugs conspired to produce the reporter's error:

1. **`_matches_model_tag` boundary check was incomplete.** It only guarded
   `3-5-sonnet` from wrongly matching the `5-sonnet` adaptive tag; it missed
   `4-5-sonnet`. This finishes the fix started in commit `705828e1` which
   caught the Sonnet 3.5 case only.
2. **No wire-shape resolver existed.** `model_factory.py` passed the
   internal `"adaptive"` mode string straight through as `thinking.type`,
   leaking a non-wire-valid value to the API on classic models.

## Fix

- **New `resolve_anthropic_thinking_payload()`** in `model_utils.py` picks
  the wire shape based on `supports_adaptive_thinking()`. Adaptive-
  supporting models get `{type:"adaptive"}` (+ optional
  `display:"summarized"` on 4.7+/Sonnet 5/Fable 5). Classic models get
  `{type:"enabled", budget_tokens:N}`. `off`/`disabled`/unknown returns
  `None` and the caller omits the field.
- **Tightened `_matches_model_tag`**: the old-minor-version guard now
  rejects any `X-5-sonnet` / `X.5-sonnet` pattern where X is a digit, not
  just X=3.
- **`model_factory.py`** calls the resolver and gates
  `output_config.effort` injection on
  *resolved-payload-is-adaptive* AND `model_supports_setting(m, "effort")`.
  Classic models never receive `output_config` (verified: they 400 on it).
- **Extracted `_model_matches_any_tag()`** helper to DRY the two
  classifier predicates.
- Public repo also gets the boundary-aware `_matches_model_tag` ported
  over (it was using naive `tag in c` matching), so classifier behavior is
  identical across both trees.

## Verification

- Direct HTTP probes to `puppy-backend.walmart.com/anthropic` against
  Sonnet 4.5 and Opus 4.7 endpoints confirmed the accept/reject matrix
  above.
- **Private repo**: 246 tests pass.
- **Public repo**: 146 tests pass (2 pre-existing SSL env failures
  unrelated to this change).
- Headless CLI runs on both `claude-4-5-sonnet` and `claude-4-7-opus`
  succeed cleanly end-to-end.

## Blast radius pre-fix

- Users of `claude-4-5-sonnet` alias with default settings → **HTTP 400 on every request**
- Users who manually set `extended_thinking = "adaptive"` in the settings
  menu on a classic model → **HTTP 400**
- Users on canonical model spellings (`claude-sonnet-4-5`, Opus 4.6+,
  Sonnet 5) → unaffected

## Follow-ups (not this PR)

- Settings menu should filter thinking modes by model support — currently
  offers `"adaptive"` on Sonnet 4.5 and `"enabled"` on Opus 4.7 even
  though the API rejects both. Wire is now safe, UX still misleading.
- `_SUMMARY_TAGS ⊂ _ADAPTIVE_TAGS` invariant is implicit — when Sonnet 6
  ships someone will add it to one tuple and forget the other. Worth
  deriving one from the other or pinning with a comment.
- Mode string matching (`enabled`/`adaptive`) is case-sensitive. Could
  `.lower()` for defense in depth.

## Jira

Tried to file a ticket via `pup-ticket-creator`; Jira's
Application/Service option-lookup endpoint is 403'ing right now.
Description above stands in for the ticket body — happy to add a ticket
link in a follow-up commit once Jira is back up.

### DIFF
--- a/code_puppy/model_factory.py
+++ b/code_puppy/model_factory.py
@@ -316,7 +316,7 @@ def make_model_settings(
 
         from code_puppy.model_utils import (
             get_default_extended_thinking,
-            should_use_anthropic_thinking_summary,
+            resolve_anthropic_thinking_payload,
         )
 
         actual_model_id = model_config.get("name", model_name)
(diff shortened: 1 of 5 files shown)


WSxDemise (approved):
Tested locally and on VDI - looks good.

PR #523, opened 2026-07-04
feat(gemini): ship gemini_oauth plugin + fix Code Assist streaming

chucky5o:
## What & why

`model_factory.py` already references `code_puppy.plugins.gemini_oauth` for the
`gemini_oauth` model type, but **the plugin was never shipped** — so that model
type is a dangling import today. This PR ships the plugin and fixes two
streaming bugs, letting users run **Gemini** models through their **local
Gemini / Antigravity CLI's** existing OAuth login — **no API key**.

> Google has been rebranding the Gemini CLI tooling as *"Antigravity"*. Both
> write credentials to `~/.gemini/`, so the same plugin handles either name.

### 1. New `gemini_oauth` plugin (`code_puppy/plugins/gemini_oauth/`)
Reuses the token the Gemini CLI stores at `~/.gemini/oauth_creds.json` and calls
the Code Assist API (`cloudcode-pa.googleapis.com`). Exposes the three symbols
`model_factory` expects:

- `GEMINI_OAUTH_CONFIG` — Code Assist base URL + API version
- `get_valid_access_token()` — reads the CLI creds, auto-refreshes when expired, writes back `chmod 600`
- `get_project_id()` — discovers the managed GCP project via `loadCodeAssist` and caches it

Client credentials (the public installed-app values from the Gemini CLI) are read
from `GEMINI_OAUTH_CLIENT_ID` / `GEMINI_OAUTH_CLIENT_SECRET` env vars — **no
secret is committed**. The plugin README explains how to extract them from a
local `gemini` install to enable auto-refresh.

### 2. Fix streaming in `gemini_code_assist.py`
Two bugs broke the streaming path (which is what the agent runtime uses by default):

- `request_stream()` was missing the `run_context` parameter the runtime passes
  → `TypeError: takes 4 positional arguments but 5 were given`
- the streaming response class didn't inherit from pydantic_ai's `StreamedResponse`
  ABC → `'StreamedResponse' object has no attribute 'get'`. Replaced with
  `CodeAssistStreamedResponse`, a proper `@dataclass` subclass implementing
  `_get_event_iterator()` (mirrors the existing `GeminiStreamingResponse`).

### 3. Docs
- `docs/LOCAL_PROVIDERS.md` — how to wire **Ollama**, **Claude Code**, and
  **Gemini/Antigravity** from their local CLIs, all switchable via `/model`.
- `code_puppy/plugins/gemini_oauth/README.md` — plugin setup + troubleshooting.

## Testing

Verified end-to-end through the `code-puppy` CLI against a real account:

- `/model antigravity-flash` → live `generateContent` call returns a response
- token auto-refresh from an expired `~/.gemini/oauth_creds.json` works
- managed project ID discovered via `loadCodeAssist` and cached
- non-streaming and streaming paths both exercised

## Notes for the maintainer

- Personal model lists live in `~/.code_puppy/extra_models.json` (not committed).
- If you'd prefer to hardcode the public client credentials rather than use env
  vars, that's a one-line change — I used env vars to keep secret scanners happy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/code_puppy/gemini_code_assist.py
+++ b/code_puppy/gemini_code_assist.py
@@ -28,7 +28,9 @@ from pydantic_ai.messages import (
     ToolReturnPart,
     UserPromptPart,
 )
-from pydantic_ai.models import Model, ModelRequestParameters
+from dataclasses import dataclass, field
+from pydantic_ai.messages import ModelResponseStreamEvent
+from pydantic_ai.models import Model, ModelRequestParameters, StreamedResponse
 from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import ToolDefinition
 from pydantic_ai.usage import RequestUsage
@@ -98,7 +100,8 @@ class GeminiCodeAssistModel(Model):
         messages: list[ModelMessage],
         model_settings: ModelSettings | None,
         model_request_parameters: ModelRequestParameters,
-    ) -> AsyncIterator[StreamedResponse]:
+        run_context: Any = None,
+    ) -> AsyncIterator[CodeAssistStreamedResponse]:
         """Make a streaming request to the Code Assist API."""
         request_body = self._build_request(
             messages, model_settings, model_request_parameters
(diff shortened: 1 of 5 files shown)


mpfaffenberger:
Please make linters + tests pass and remove claude code authorship

thomwebb:

PR #531, opened 2026-07-07
Feat/oss secret store

GregKinne:
Adds a generic OS keyring secret store to OSS core with a consolidated
macOS Keychain backend.

**Mechanism only** -- this migrates no existing secrets and ships no secret
values. Downstream distributions can fork the constant `_SERVICE_NAME` to
namespace their secrets.

### Core secret store (`code_puppy/secret_store.py`)
- Public API: `get_secret` / `set_secret` / `delete_secret` /
  `keyring_available`.
- Keyring-first; falls back to a permission-hardened JSON file only
  when no keyring backend is available (headless / CI).

### Consolidated macOS backend (`code_puppy/secret_store_backends.py`)
- `ConsolidatedKeychainBackend`: stores all of a service's secrets in
  a single Keychain item (JSON blob) instead of one item per secret.
- One ACL prompt covers all secrets -- users approve Keychain access
  once, not once per secret.
- In-process lock + `flock` across read-modify-write so parallel
  sessions and MCP subprocesses cannot clobber each other.
- Corruption detection: a corrupt blob raises instead of silently
  wiping all secrets on the next write.
- Lazy install via `_ensure_backend()` on first secret operation;
  Darwin-gated so non-macOS platforms use the native keyring backend
  unchanged.

### Cross-platform fixes (merged from `fix/secret-store-cross-platform`)
- **`ImportError` guard** in `_ensure_backend()`: `secret_store_backends`
  imports `fcntl` (POSIX-only); on Windows, the import is caught and
  the native backend is used instead of crashing.
- **Stranded-secret warning**: if the keyring is healthy but a write
  fails (transient error, dismissed Keychain prompt), `set_secret()`
  now warns instead of silently writing to the fallback file where
  `get_secret()` would never read it.
- **`os.fchmod` to `os.chmod`**: the fallback path exists for
  environments without a keyring (headless Linux, CI, Windows).
  `os.fchmod` is POSIX-only; `os.chmod` works everywhere.

### Dependencies
- `pyproject.toml` adds `keyring>=24.0.0` (+ `uv.lock` update).
  Does **not** add `msal-extensions`.

## Verification
- `pytest tests/test_secret_store.py tests/test_secret_store_backends.py`
  -- **38 tests passed**.
- `ruff check` -- clean.

---

## Review remediation summary (F1–F11)

All eleven review findings are resolved on this branch as focused, individually-reviewable commits. Inline replies are attached to each original thread; this table is the consolidated record.

| # | Severity | Finding | Resolution (mechanism) | Commit |
|---|----------|---------|------------------------|--------|
| F1 | Blocker | `chmod(0o600)` does not establish an owner-only NTFS DACL; Windows security claim invalid | Owner-only DACL applied via `icacls /inheritance:r` + `/grant:r <user>:F`; claim withdrawn with a visible notice when `icacls` is absent; docstrings/tests corrected | `c4c7bfd5` |
| F2 | Blocker | Fallback lookup keyed on `name` only, bypassing the service namespace (cross-distribution read) | Fallback document nested as `{service: {name: value}}`; all access scoped through `_service_name`; legacy flat file migrated under the default namespace | `13a18b03` |
| F3 | Medium | Inconsistent fallback failure contract (`mkstemp` `OSError` escapes; other failures return `False`) | `mkstemp()` moved inside the guarded block; all failure modes return a uniform `False` | `9e91e9ad` |
| F4 | Blocker | `set_secret()` ignored `_write_fallback()` failure and returned normally (silent credential loss) | Raises `SecretStoreError` when the keyring is unavailable/failed and `_fallback_set()` returns `False` | `9e91e9ad` |
| F5 | High | Fallback read-modify-write is unlocked (lost updates across processes) | Serialized under a cross-process advisory lock: `fcntl.flock(LOCK_EX)` (POSIX) / `msvcrt.locking(LK_LOCK)` (Windows) | `202b31f7` |
| F6 | High | Successful keyring write never scrubbed the fallback (plaintext persistence + stale resurrection) | Successful `_keyring_set()` calls `_fallback_scrub(name)` under the F5 lock; test asserts absence post-set | `202b31f7` |
| F7 | High | Chunked overwrite not crash-safe; no cross-process lock on the platform where it activates | Generation-numbered scheme: chunks at `name:cp:<gen>:<i>`, atomic flip of pointer `name:cp:n` → `<gen>:<count>`, best-effort GC of the prior generation | `117a47f1` |
| F8 | Medium | Empty value → false-alarm warning + silent no-op; `.strip()` silently mutated secrets | Empty/whitespace values raise `ValueError` at the boundary; value stripping removed on write/read/fallback (byte-for-byte storage) | `9de46228` |
| F9 | High | `:cp:` reserved namespace unenforced; names could shadow/destroy chunk metadata | Single validation helper rejects any name containing `:cp:` | `e1e2dfcd` |
| F10 | Medium | `delete_secret()` ignored `_write_fallback()` return (delete could report success on read-only FS) | Raises `SecretStoreError` when `_fallback_delete()` returns `False` | `9e91e9ad` |
| F11 | Low | `warnings.warn` invisible in the TUI and deduped by the default filter | User-facing notices route through the messaging bus (`emit_warning`); `warnings.warn` retained only as a bus-unavailable fallback | `e6df57d5` |

### Commit series
```
117a47f1  generation-numbered crash-safe chunking (F7)
e6df57d5  route user-facing notices through the messaging bus (F11)
c4c7bfd5  owner-only DACL on Windows, not false chmod (F1)
13a18b03  namespace the fallback file by service name (F2)
202b31f7  lock fallback RMW + scrub stale plaintext on healthy set (F5/F6)
9e91e9ad  uniform fallback failure contract, surfaced to callers (F3/F4/F10)
9de46228  stop stripping secrets; reject empty values loudly (F8)
e1e2dfcd  enforce reserved ':cp:' namespace on secret names (F9)
```

### Verification
`pytest tests/test_secret_store.py` — 71 passed.


GregKinne:
responding to the failed CI ruff check with commit [a78a30e](https://github.com/mpfaffenberger/code_puppy/pull/531/commits/a78a30e626047927983a71458fa804b34f53c3c9)

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+
+def _warn_fallback_active() -> None:
+    global _warned_fallback
+    if _warned_fallback:
+        return
+    _warned_fallback = True
+    warnings.warn(
+        "No OS keyring backend is available; secrets are stored in the "
+        f"permission-hardened fallback file at {_FALLBACK_FILE} "
+        "(mode 0o600). This is intended for headless/CI use only.",
```
mpfaffenberger: **Blocker — this security claim is false on Windows.** When Credential Manager is unavailable, the fallback contains plaintext secrets, but `chmod(0o600)` does not establish an owner-only NTFS DACL. The tests explicitly skip permission verification on Windows for that reason, while this warning still tells Windows users the file is permission-hardened. Please use a meaningful Windows fallback (for example current-user DPAPI encryption or an owner-only DACL), make insecure plaintext fallback explicitly opt-in, or fail closed with a clear error. At minimum, do not advertise `0o600` as protection on Windows. This is especially important because old/stripped-down Windows is precisely where this fallback will activate.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+    value = _keyring_get(name)
+    if value:
+        return value
+
+    # Always check the fallback file: a prior set_secret may have ended up
+    # there after both keyring paths failed.  Only emit the headless warning
+    # when the keyring backend itself is unavailable.
+    if not keyring_available():
+        _warn_fallback_active()
+    stored = _read_fallback().get(name)
```
mpfaffenberger, line 734: **Blocker — fallback lookup bypasses the configured service namespace.** Keyring entries use `_service_name`, but the fallback is a flat dictionary indexed only by `name`. If distribution A stores `token` in fallback mode, distribution B can read it after calling `configure_service_name("distribution-b")`. The same issue affects overwrite and deletion, contradicting the promise that secrets never bleed across builds. Please namespace fallback data by service (for example `{service: {name: value}}`) and treat any legacy flat format as belonging only to the default service. Add cross-service read/write/delete tests in fallback mode.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+            json.dump(data, f, indent=2)
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp, _FALLBACK_FILE)
+    except OSError:
+        try:
+            os.unlink(tmp)
+        except OSError:
+            pass
+        return False
```
mpfaffenberger, line 653: Please make the fallback failure contract consistent. `mkstemp()` is outside this `try`, so that failure raises raw `OSError`, while failures from chmod/write/fsync/replace are converted to `False`. More importantly, the caller currently ignores this return value. A dedicated `SecretStoreError` for an unsuccessful persistence operation would make these paths explicit and testable.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+            stacklevel=2,
+        )
+    else:
+        _warn_fallback_active()
+
+    normalized = str(value).strip()
+    if not normalized:
+        return
+    data = _read_fallback()
+    data[name] = normalized
```
mpfaffenberger: **Blocker — `_write_fallback(data)` can return `False`, but the result is ignored and `set_secret()` returns normally.** On a headless server with no keyring plus a read-only/full/unwritable config filesystem, this reports apparent success while the credential is lost. Please raise a clear storage exception (preferred) or return a result that callers must handle when both storage tiers fail. Tests should cover directory creation, temp-file creation, fsync/write, and replace failures.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+            f"Keyring write failed for {name!r} despite a healthy backend "
+            "(transient error or backend crash). Storing in the secure file "
+            f"fallback at {_FALLBACK_FILE}.",
+            stacklevel=2,
+        )
+    else:
+        _warn_fallback_active()
+
+    normalized = str(value).strip()
+    if not normalized:
```
mpfaffenberger: **High priority — this fallback update is an unlocked read-modify-write transaction.** Atomic `os.replace()` prevents truncated reads, but it does not prevent lost updates: two processes can both read the same document, add different keys, and the second replacement drops the first key. This is realistic with two Code Puppy sessions and becomes the normal path on headless Linux. Please serialize the entire read/update/write operation with a cross-platform process lock plus an in-process lock, similar to the consolidated backend.

mpfaffenberger (changes requested):
I love the direction here: the generic secret-store API, Windows Credential Manager chunking, and consolidated macOS Keychain backend are all valuable. I do have several concerns in the exact environments where the fallback becomes the primary store—older Windows machines without a usable keystore and barebones/headless Linux servers.

The inline findings below are merge blockers in my view:

1. The Windows fallback is plaintext and `chmod(0o600)` does not provide the owner-only NTFS protection claimed by the warning.
2. The fallback is not scoped by `configure_service_name`, so downstream distributions can read, overwrite, or delete each other's fallback secrets.
3. A failed fallback write is ignored, causing `set_secret()` to return normally after losing the credential.
4. Fallback read-modify-write operations are not locked, so concurrent processes can overwrite each other's secrets.

The focused tests are green (`52 passed`) and Ruff is clean, but the tests skip the Windows permission guarantee and do not cover service isolation, write-failure propagation, or concurrent writers. I recommend fixing these before merge while keeping the overall design.

Review generated by the **GPT 5.6 Sol model**.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+
+    Attempts three strategies in order:
+      1. Direct keyring write (small values).
+      2. Chunked keyring write (oversized values, e.g. Windows CM cap).
+      3. Permission-hardened JSON file fallback (only when both keyring
+         strategies fail -- unexpected backend error or no keyring at all).
+    """
+    _ensure_backend()
+    if _keyring_set(name, value):
+        return
```
mpfaffenberger: **High priority — a successful keyring write never scrubs the fallback file, so rotated secrets linger in plaintext and can be resurrected.**

Scenario on exactly the machines the fallback exists for (older Windows without a working Credential Manager, headless Linux that later gains a keyring):

1. Keyring is unavailable → `set_secret("token", OLD)` lands in `secrets.json`.
2. Keyring becomes healthy → user rotates: `set_secret("token", NEW)` succeeds in the keyring and returns here — the fallback file still holds `OLD` in plaintext, forever. `delete_secret` scrubs the fallback, but a successful *set* does not.
3. If the keyring entry later vanishes (new Windows user profile, reset keychain, reinstalled backend), `get_secret("token")` silently serves the rotated-out `OLD` value from disk.

So this is both a hygiene problem (revoked secret material persisting in plaintext after the store is healthy again) and a correctness problem (silent resurrection of a stale credential, which can be a nasty debugging session when `OLD` is an expired token).

Suggested fix: after a successful `_keyring_set`, remove `name` from the fallback document (under the same lock you add for the RMW issue already flagged at line 420):

```python
if _keyring_set(name, value):
    data = _read_fallback()
    if name in data:
        del data[name]
        _write_fallback(data)
    return
```

A test asserting "fallback entry is gone after a healthy-keyring set" would lock this in.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+        stale += 1
+
+    # Commit: write the count key last as the atomic marker.
+    if not _kr_set_raw(_chunk_count_key(name), str(len(chunks))):
+        for idx in range(len(chunks)):
+            _kr_del_raw(_chunk_key(name, idx))
+        return False
+
+    _kr_del_raw(name)  # remove any stale single-entry write
+    return True
```
mpfaffenberger, line 405: **High priority — the chunked path's crash-safety claim only holds for *first* writes, and the path has no cross-process lock on the one platform where it actually activates.**

The docstring says a mid-write crash "never leaves a partially-readable value behind," but overwrites mutate the old value's chunks *in place* while the old count key still points at them:

1. **Crash mid-overwrite destroys the old value.** Old value = 3 chunks + `count=3`. A new 3-chunk write crashes after rewriting chunk 0: `count=3` now points at one new chunk and two old ones → `_keyring_get` reassembles a Frankenstein value, or the rollback at lines 269-270 *deletes* chunks that were the old value's data, leaving `count=3` pointing at nothing. Either way the previous secret is gone and `_keyring_set` returned `False`, so `set_secret` diverts the *new* value to the plaintext fallback — worst of both worlds.
2. **Prune-before-commit (lines 273-277).** Shrinking from 5 chunks to 3: trailing chunks 3-4 are deleted *before* the count key is updated. A crash in that window leaves `count=5` with chunks 3-4 missing → old value unreadable, nothing committed.
3. **No cross-process lock.** The `flock` serialization lives only in the macOS consolidated backend. Chunking exists specifically for Windows Credential Manager — so on the platform where this path runs, two concurrent sessions (main app + MCP subprocess) can interleave chunk writes for the same name and commit a corrupted assembly with no lock anywhere.

A generation-numbered scheme fixes all three without locks doing the heavy lifting: write chunks under `name:cp:<gen>:<i>`, then atomically flip a single pointer key to `"<gen>:<count>"`, then best-effort GC the old generation. Old value stays intact until the commit flips, rollback never touches prior data, and concurrent writers at worst leak a stale generation instead of corrupting each other.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+    **Large values** (len > _CHUNK_SIZE):
+      - Split into ceil(len / _CHUNK_SIZE) pieces.
+      - Chunk data entries are written first, then the count key last so a
+        mid-write crash never leaves a partially-readable value behind.
+      - Excess chunks from a prior write with more pieces are pruned.
+      - The direct *name* entry is removed to avoid ambiguity on read.
+
+    Returns ``True`` on success, ``False`` if any keyring write fails.
+    """
+    normalized = str(value).strip()
```
mpfaffenberger: **Two related contract problems with value normalization here.**

1. **Empty value → false-alarm warning + silent no-op.** `set_secret("token", "  ")` hits this early return, so `_keyring_set` reports `False` — indistinguishable from a genuine backend failure. `set_secret` then warns *"Keyring write failed for 'token' despite a healthy backend (transient error or backend crash)"* — none of which is true — and finally no-ops at its own emptiness check. The caller gets a scary, misleading diagnostic and no persisted change, with no way to distinguish "you passed garbage" from "your keyring is dying." Validate emptiness up front in `set_secret` and raise `ValueError` (or explicitly treat it as a delete) *before* entering the strategy cascade, and let `_keyring_set`'s `False` mean exactly one thing: the keyring write failed.

2. **`.strip()` silently mutates secrets.** Normalization is applied on write here, on read in `_kr_get_raw`, and in the fallback paths — so a password or API key with legitimate leading/trailing whitespace (they exist, and they're miserable to debug) is corrupted with no error: `get_secret` returns something different from what was passed to `set_secret`. A secret store should be byte-faithful. If stripping is deliberately defensive against copy-paste newlines, that belongs at the input boundary (the UI collecting the secret), not buried in the storage layer — or at minimum documented loudly in the public API docstrings.

code_puppy/secret_store.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+def _chunk_count_key(name: str) -> str:
+    """Keyring entry name for the chunk-count (commit) marker."""
+    return f"{name}{_COUNT_SUFFIX}"
+
+
+def _chunk_key(name: str, i: int) -> str:
+    """Keyring entry name for chunk *i* of *name*."""
+    return f"{name}{_CHUNK_NS}{i}"
```
mpfaffenberger, line 198: **The `:cp:` reserved namespace is not enforced, so chunk metadata can be shadowed or destroyed by ordinary secret names.**

The comment above claims collisions are "essentially impossible," but nothing validates `name`, so they're trivially possible:

- `set_secret("foo:cp:n", "3")` writes a single entry that `_keyring_get("foo")` then reads as a chunk-count marker → it tries to assemble 3 chunks of `foo`, finds none, and returns `None` even when a perfectly good direct `foo` entry exists (the direct read at the end of `_keyring_get` is never reached once a count key is found).
- `delete_secret("foo")` → `_delete_chunks("foo")` deletes the entry named `foo:cp:n`, i.e. destroys a *different* secret.
- Symmetrically, a chunked write of `foo` clobbers any real secret named `foo:cp:0`.

Since this module is explicitly pitched as a generic store for downstream distributions to build on, secret names will eventually be constructed from user- or config-derived strings, and "nobody would ever name a secret that" is not a security posture. Cheap fix: reject names containing `_CHUNK_NS` in the public API (`ValueError`), which also protects `get`/`delete`. A stricter alternative is escaping/hashing the logical name into the entry key, but validation alone closes the hole.

mpfaffenberger (commented):
Adding a second pass focused on the environments where the fallback and chunked paths become the *primary* paths — older Windows machines with no usable Credential Manager and barebones headless Linux servers. These findings are additive to the five inline comments already posted (plaintext Windows fallback, service-name scoping, mkstemp failure contract, ignored `_write_fallback` result, unlocked fallback RMW) — none of them are duplicates.

New inline findings, in rough priority order:

1. **Stale fallback resurrection** — a successful keyring write never scrubs the fallback entry, so a rotated-out secret lingers in plaintext on disk indefinitely and can be silently resurrected later.
2. **Chunked overwrite is not crash-safe and not locked** — the in-place overwrite breaks the documented "never partially readable" guarantee, and chunking only ever activates on Windows, exactly where no cross-process lock exists.
3. **Empty value produces a false-alarm warning + silent no-op**, and universal `.strip()` silently mutates secrets with legitimate leading/trailing whitespace.
4. **`:cp:` chunk-name collision** — the reserved suffix is not validated, so a hostile-or-unlucky secret name can shadow or destroy chunk metadata.

Two smaller notes that don't warrant their own threads:

- `delete_secret()` also ignores the `_write_fallback()` return value (same failure-contract class as the existing `set_secret` comment). For a *delete*, that's arguably worse: on a read-only or full filesystem, "delete" returns normally while the plaintext secret remains on disk. Whatever contract you settle on for `set_secret`, please apply it here too.
- The module signals through `warnings.warn`, which is likely invisible inside the TUI and deduplicated by the default warnings filter. Consider routing user-facing notices through the messaging bus so the headless-fallback warning actually reaches a human.

Overall direction still gets an enthusiastic tail wag — the three-tier strategy and the consolidated Keychain backend are the right shape. The issues above are all about the degraded tiers being trustworthy on the machines that need them most.

Review generated by the model **Fable**.

code_puppy/secret_store.py
GregKinne: Confirmed. `chmod(0o600)` does not constrain an NTFS ACL, so the owner-only guarantee was invalid on Windows.

Change (`c4c7bfd5`): on Windows the fallback file receives an explicit owner-only DACL via `icacls <path> /inheritance:r` then `/grant:r <user>:F`. When `icacls` is unavailable the guarantee is withdrawn and a visible notice is emitted rather than asserted. Docstrings restated to per-platform guarantees; tests no longer skip the assertion silently.

code_puppy/secret_store.py
GregKinne: Confirmed. The fallback lookup keyed on `name` only, ignoring `_service_name`, which permitted cross-distribution reads.

Change (`13a18b03`): the fallback document is nested as `{service: {name: value}}`; all reads and writes are scoped through `_service_name`. A legacy flat document is migrated under the default service namespace on first read, preserving existing secrets without introducing cross-namespace exposure.

code_puppy/secret_store.py
GregKinne: Confirmed. `mkstemp()` sat outside the guarded block, so its failure escaped as raw `OSError` while other failures returned `False`.

Change (`9e91e9ad`): `mkstemp()` moved inside the guarded block; all failure modes (mkstemp/chmod/write/fsync/replace) yield a uniform `False`. The return value is now consumed by callers (see F4/F10).

code_puppy/secret_store.py
GregKinne: Confirmed. The `_write_fallback` result was discarded, so an unwritable config filesystem reported apparent success while the credential was lost.

Change (`9e91e9ad`): when the keyring is unavailable (or both direct and chunked keyring writes fail) and `_fallback_set()` returns `False`, `set_secret()` raises `SecretStoreError`. Failure is surfaced at write time; silent loss at read time is eliminated.

code_puppy/secret_store.py
GregKinne: Confirmed. `os.replace()` prevents torn reads but not lost updates across concurrent writers.

Change (`202b31f7`): the fallback read-modify-write is serialized under a cross-process advisory lock — `fcntl.flock(LOCK_EX)` on POSIX, `msvcrt.locking(LK_LOCK)` on Windows — closing the lost-update window.

code_puppy/secret_store.py
GregKinne: Confirmed. A successful keyring set did not scrub the fallback, leaving rotated plaintext on disk with a stale-resurrection path.

Change (`202b31f7`, as suggested): a successful `_keyring_set()` invokes `_fallback_scrub(name)` under the same lock as F5, removing both the plaintext-persistence and resurrection paths. A test asserts the fallback entry is absent after a healthy-keyring set. Empirically observed: after a keyring rotation, the fallback document was empty while the keyring held the current value.

code_puppy/secret_store.py
GregKinne: Confirmed for all three cases (in-place overwrite, prune-before-commit, absence of a Windows cross-process lock).

Change (`117a47f1`, generation-numbered scheme as proposed): chunks are written under `name:cp:<gen>:<i>`; a single pointer key `name:cp:n` is then flipped atomically to `<gen>:<count>`; the prior generation is garbage-collected best-effort afterward. The committed value is never mutated in place, so a crash during overwrite or prune cannot produce a partially-readable or unreadable value, and chunk crash-safety no longer depends on a lock on Windows.

code_puppy/secret_store.py
GregKinne: Confirmed for both sub-issues.

Change (`9de46228`): (1) empty or whitespace-only values raise `ValueError` at the public boundary, so a no-op is no longer indistinguishable from a backend failure; (2) `.strip()` normalization of secret values was removed from the write, read (`_kr_get_raw`), and fallback paths — values are stored and returned byte-for-byte, preserving credentials with significant leading/trailing whitespace.

code_puppy/secret_store.py
GregKinne: Confirmed. Nothing validated `name`, so an ordinary name could shadow chunk metadata or destroy a different secret.

Change (`e1e2dfcd`): all secret names pass through a single validation helper that rejects any name containing the reserved substring `:cp:`. This closes the shadowing, cross-secret deletion, and chunk-clobber cases. The "unlikely name" assumption is removed.

GregKinne:
## Review remediation summary (F1–F11)

All eleven review findings are resolved on this branch as focused, individually-reviewable commits. Inline replies are attached to each original thread; this table is the consolidated record.

| # | Severity | Finding | Resolution (mechanism) | Commit |
|---|----------|---------|------------------------|--------|
| F1 | Blocker | `chmod(0o600)` does not establish an owner-only NTFS DACL; Windows security claim invalid | Owner-only DACL applied via `icacls /inheritance:r` + `/grant:r <user>:F`; claim withdrawn with a visible notice when `icacls` is absent; docstrings/tests corrected | `c4c7bfd5` |
| F2 | Blocker | Fallback lookup keyed on `name` only, bypassing the service namespace (cross-distribution read) | Fallback document nested as `{service: {name: value}}`; all access scoped through `_service_name`; legacy flat file migrated under the default namespace | `13a18b03` |
| F3 | Medium | Inconsistent fallback failure contract (`mkstemp` `OSError` escapes; other failures return `False`) | `mkstemp()` moved inside the guarded block; all failure modes return a uniform `False` | `9e91e9ad` |
| F4 | Blocker | `set_secret()` ignored `_write_fallback()` failure and returned normally (silent credential loss) | Raises `SecretStoreError` when the keyring is unavailable/failed and `_fallback_set()` returns `False` | `9e91e9ad` |
| F5 | High | Fallback read-modify-write is unlocked (lost updates across processes) | Serialized under a cross-process advisory lock: `fcntl.flock(LOCK_EX)` (POSIX) / `msvcrt.locking(LK_LOCK)` (Windows) | `202b31f7` |
| F6 | High | Successful keyring write never scrubbed the fallback (plaintext persistence + stale resurrection) | Successful `_keyring_set()` calls `_fallback_scrub(name)` under the F5 lock; test asserts absence post-set | `202b31f7` |
| F7 | High | Chunked overwrite not crash-safe; no cross-process lock on the platform where it activates | Generation-numbered scheme: chunks at `name:cp:<gen>:<i>`, atomic flip of pointer `name:cp:n` → `<gen>:<count>`, best-effort GC of the prior generation | `117a47f1` |
| F8 | Medium | Empty value → false-alarm warning + silent no-op; `.strip()` silently mutated secrets | Empty/whitespace values raise `ValueError` at the boundary; value stripping removed on write/read/fallback (byte-for-byte storage) | `9de46228` |
| F9 | High | `:cp:` reserved namespace unenforced; names could shadow/destroy chunk metadata | Single validation helper rejects any name containing `:cp:` | `e1e2dfcd` |
| F10 | Medium | `delete_secret()` ignored `_write_fallback()` return (delete could report success on read-only FS) | Raises `SecretStoreError` when `_fallback_delete()` returns `False` | `9e91e9ad` |
| F11 | Low | `warnings.warn` invisible in the TUI and deduped by the default filter | User-facing notices route through the messaging bus (`emit_warning`); `warnings.warn` retained only as a bus-unavailable fallback | `e6df57d5` |

### Commit series
```
117a47f1  generation-numbered crash-safe chunking (F7)
e6df57d5  route user-facing notices through the messaging bus (F11)
c4c7bfd5  owner-only DACL on Windows, not false chmod (F1)
13a18b03  namespace the fallback file by service name (F2)
202b31f7  lock fallback RMW + scrub stale plaintext on healthy set (F5/F6)
9e91e9ad  uniform fallback failure contract, surfaced to callers (F3/F4/F10)
9de46228  stop stripping secrets; reject empty values loudly (F8)
e1e2dfcd  enforce reserved ':cp:' namespace on secret names (F9)
```

### Verification
`pytest tests/test_secret_store.py` — 71 passed.

mpfaffenberger:
@thomwebb - can you do me a huge favor and test this? we need to get it merged, but I'm scared of breaking something.

If you can test MacOSX, I can test Linux and Windoze 11.

PR #536, opened 2026-07-08
feat: native Agent Client Protocol (ACP) agent + host-agnostic FileSystemBackend seam

JulienEllie:
> **This is a proposal — opening as a draft to get your take on the approach before polishing.** Happy to split it, reshape the API, or adjust naming/scope. Nothing here is precious.

## What & why

Makes Code Puppy a **native agent inside any Agent Client Protocol (ACP) host** (Zed and other editors) via `code-puppy --acp`, and generalizes the tool layer's I/O into host-agnostic seams so a host can back Code Puppy's file and shell operations with no ACP-specific assumptions leaking into core.

**Two commits:**

1. **`feat(plugins): native ACP agent`** — a self-contained `code_puppy/plugins/acp/` plugin implementing the ACP SDK's `Agent` interface: capability negotiation, prompt content-block parsing (incl. images), the full session lifecycle (initialize / new / load / resume / fork / list / close / prompt / cancel / set_model / set_config_option), event bridging (streamed message + thought chunks, tool calls with kind/locations/inline diffs), client-authoritative permissions, and capability-gated I/O delegation to the host's `fs/*` + `terminal/*`. Rides existing callback hooks — no edits to `command_line/` or the CLI runner.

2. **`refactor(tools): make FileSystemBackend own the whole FS surface`** — the pluggable I/O seam the ACP plugin needs, designed to stand on its own (below).

## Session lifecycle correctness

The session lifecycle was hardened end-to-end against a real Zed client. Four behaviors that a persisted, multi-turn, restartable ACP agent must get right — all covered by tests:

- **Survives restart.** `session/list` merges live in-memory sessions with those persisted on disk (deduped, live wins), and `fork` falls back to persisted history — so sessions stay listable and revivable after the `--acp` process restarts. `close` tombstones the persisted copy so a deliberate close doesn't resurrect.
- **Remembers turns.** After each turn, `prompt()` folds `result.all_messages()` back into the agent (exactly as `cli_runner` does). Without this the shared runtime never writes the turn back, so the agent forgets everything and persistence saves only the user's prompt — never the reply.
- **Rebuilds the thread on open.** `load`/`resume` replay the rehydrated history to the client as ordered `session/update` chunks (user / agent / thinking / past tool calls), so the client reconstructs the conversation instead of showing — and discarding — an empty thread.
- **No truncation.** The stream bridge forwards a part's *initial* content (`PartStartEvent`), not just deltas, so messages aren't chopped at the front when the model front-loads the opening tokens.

A per-session `cwd` is honored via a `ContextVar` + `resolve_path()` (no process-global `os.chdir`, so concurrent sessions don't corrupt each other), and off-main-thread SIGINT wiring is skipped (signal handlers are main-thread-only in CPython) so the ACP server thread doesn't crash on startup.

## The FileSystemBackend seam

When a host backend is installed it owns **every** filesystem operation the agent's tools perform, so the agent sees one coherent filesystem — nothing about ACP is assumed in core.

Rationale for owning the whole surface: if only text read/write were routed through a backend and `list_files` / existence / delete / mkdir / grep stayed on local disk, you'd get a **split-brain filesystem** — `read_file` could serve a host buffer while `list_files` reports local disk. Coherent only when the host *is* the local disk; incoherent for any remote/sandbox/virtual backend. So the backend owns every operation; recursive listing and grep are *composed* by core from `list_dir` + `read_text_file`, keeping the backend contract small.

- **`io_backends.py`** — `FileSystemBackend` (`read`/`write` + `exists`/`is_file`/`is_dir`/`list_dir`/`delete_file`/`make_dirs`, with a `DirEntry` type) and `CommandExecutor`, both `None` by default (local I/O), with documented path (always absolute), error (`FileNotFoundError`/`OSError`), and UTF-8 contracts.
- **`fs_access.py`** (new) — a single dispatch facade ("backend or local disk?") for metadata, mutation, read/write, and the recursive `walk` that `list_files`/`grep` compose from. One place to make the decision (DRY); coherence guaranteed by construction.
- **tools** — `list_files`, `grep`, existence, read, delete, mkdir, and undo route through the facade. **The fast local ripgrep path is retained verbatim for the default (no-backend) case — zero behavior change for a normal install.**
- **ACP adapter** — serves *topology* (list/exists/delete/mkdir) from local disk (its host shares the disk) and *content* from the client. That "topology is local" choice lives in the adapter that legitimately knows its host shares the disk — never baked into the general seam.

### Robustness

- `walk` is iterative (deep trees can't overflow the stack); cycles broken by a realpath visited-set plus a `MAX_WALK_DEPTH` backstop for a backend returning an infinite chain of distinct paths.
- backend `list_files` degrades to a tool error instead of crashing if a backend raises; grep skips unreadable files instead of aborting.

## Testing

- **Default (no-backend) path unchanged** — existing file-op/list/grep/read/mod suites pass untouched.
- `tests/tools/test_fs_backend.py`: a reference in-memory backend (proves the seam is host-agnostic) + contract, coherence (write → immediately listable *and* greppable), tool-level, grep flag-mode, and adversarial (cyclic / deep / raising backend) cases.
- `tests/test_acp_plugin.py`: capability negotiation, prompt parsing, full session lifecycle, **list-across-restart, multi-turn memory, load/resume replay, streaming start-chunk**, tool-call bridging, permission delegation, and delegated fs/terminal backends incl. terminal edge paths.
- Full suite green.

## Notes for reviewers

- Rebased on latest `main`; the only merge point is `_runtime.py`, where the ACP off-main-thread SIGINT guard meets `main`'s pure-keybinding Ctrl+C handling (both preserved).
- `backend grep` uses Python `re` where the local path uses Rust regex — common patterns are identical; exotic features (lookaround/backrefs) differ. `--type` covers ~24 common types and errors clearly on unknown ones.
- Happy to land the seam refactor separately from the ACP plugin if you'd rather review them independently.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -815,10 +815,19 @@ async def _run_with_mcp_impl(
             # console clamp on Windows), ^C arrives as \x03 and the key
             # listener handles cancellation. SIGINT only fires out-of-band
             # (kill -INT, piped stdin with no listener, cooked-mode gaps
-            # between raw readers) — the handler installed here is that
+            # between raw readers) -- the handler installed here is that
             # fallback: it cancels when ^C IS the cancel gesture, and only
             # hints at the real key when cancel is remapped.
-            if sigint_fallback_cancels():
+            #
+            # Off the main thread (ACP server, embeds, worker threads) we skip
+            # SIGINT wiring entirely: signal handlers are main-thread-only in
+            # CPython, so installing one there raises "signal only works in
+            # main thread". Cancellation still flows through
+            # schedule_agent_cancel, and the stdin key listener no-ops when
+            # stdin isn't a TTY, so no cancel affordance is lost.
+            if threading.current_thread() is not threading.main_thread():
+                pass  # original_handler stays None -> restore is a no-op
+            elif sigint_fallback_cancels():
                 original_handler = signal.signal(
                     signal.SIGINT, keyboard_interrupt_handler
                 )
--- a/code_puppy/plugins/acp/README.md
+++ b/code_puppy/plugins/acp/README.md
@@ -1,0 +1,266 @@
+# `acp` — Code Puppy as a native ACP agent
+
+Run Code Puppy inside any ACP-capable editor's agent panel (e.g.
+[Zed](https://github.com/zed-industries/zed))
+by speaking the **Agent Client Protocol (ACP)** — the same JSON-RPC-over-stdio
+protocol these editors use to host Gemini CLI and Claude Code.
+
+This plugin is built on the **official [`agent-client-protocol`](https://pypi.org/project/agent-client-protocol/)
+Python SDK** (`acp`). The SDK owns the wire — stdio binding, JSON-RPC framing,
+typed Pydantic models, and the bidirectional connection — so this plugin is
+pure *behaviour*: mapping ACP sessions to Code Puppy agents, translating
+runtime events into `session/update`s, and routing Code Puppy's I/O + approval
+edges to the client.
+
+> **Status:** live over real stdio. Implemented: `initialize`, `session/new`,
+> `session/load` (replays persisted history), `session/prompt`,
+> `session/cancel` (also kills local shells), `session/list`, `session/close`,
+> `session/fork` (duplicates history), `session/resume` (rehydrates across
+restarts), `session/set_mode`, `session/set_config_option`. Assistant text +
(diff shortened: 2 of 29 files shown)


JulienEllie:
This one is chunky with core changes. Should get extra attention :)

JulienEllie:
Screenshot 2026-07-08 at 5 24 03 PM
Runs inside Zed!

JulienEllie:
Well this was painful. The protocol is evolving and Zed/other are not all using it the same way. For now, I think I got it but it's something that I will have to maintain to keep on top of.

PR #545, opened 2026-07-10
feat(callbacks): let run_shell_command hooks rewrite the command

ashishsinghi:
## Summary

Small, backward-compatible extension of the existing `run_shell_command` callback contract: a callback can now return `{"rewrite": "<new command>"}` to transparently transform a shell command before execution, in addition to the existing `{"blocked": True}` allow / block semantics.

This PR ships **infrastructure only** — no policy. Following review feedback on the earlier version, the opinionated `git_coauthor` plugin has been pulled out entirely and will live as a user / marketplace plugin instead of a builtin.

## What changed

**`code_puppy/tools/command_runner.py`** — ~15 lines. After the existing block-check loop, iterate callback results and apply `result["rewrite"]` (in registration order, only when non-empty and different from the current command). Each applied rewrite fires a visible `emit_info` line so users always see what actually ran — rewriting is never sneaky.

**`AGENTS.md`** — one line. Document the new `rewrite` key alongside `blocked` in the `run_shell_command` hook row.

**`tests/tools/test_command_runner_full_coverage.py`** — three new cases in `TestRunShellCommand`:
- `test_rewrite_from_callback_replaces_command` — happy path: rewrite is applied and the transformed command is what `_execute_shell_command` sees.
- `test_rewrite_ignored_when_same_as_original` — no `command rewritten` notice fires when the rewrite string equals the original (no false-positive UI noise).
- `test_rewrite_ignored_for_empty_or_non_string` — defensive: empty strings, `None`, and non-string payloads are ignored so a broken plugin can't wipe out the user's command.

## Backward compatibility

- Callbacks returning `None` → unchanged behavior (allow).
- Callbacks returning `{"blocked": True, ...}` → unchanged behavior (block).
- Callbacks returning anything else → unchanged behavior (ignored), unless they explicitly include a `"rewrite"` key with a non-empty string.

Verified by running the full pre-existing shell-command test suite:

```
259 passed, 2 skipped
```

across `test_command_runner_full_coverage`, `test_command_runner_extended`, `test_shell_safety`, `test_shell_safety_callbacks`, `test_destructive_command_detector`, `test_shell_passthrough`, and `test_shell_line_crlf`.

`ruff check` + `ruff format` clean.

## What this unlocks

Plugins that need to *transform* a command instead of merely blocking it — a class that isn't possible on the current contract:

- **Secret redaction** — strip API keys / tokens accidentally included in a command before it runs.
- **Corporate proxy / mirror injection** — prepend `HTTPS_PROXY=…` or rewrite `pip install` to point at an internal index.
- **Path fixup** — canonicalize relative paths that would otherwise depend on the shell's `pwd`.
- **Attribution trailers** on `git commit` — the opinionated case that motivated this work, now correctly living outside core.

## Context

This is the slimmed-down version of the original PR (which shipped both the hook extension **and** a `git_coauthor` builtin plugin). Per the reviewer's feedback — opinionated attribution defaults belong outside core, and community/marketplace plugins are the right home for them — the plugin has been removed from this PR entirely. Only the enabling infrastructure remains.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ and skills (`<CWD>/.code_puppy/skills/`).
 | `agent_run_start` | Before agent task | `(agent_name, model_name, session_id=None) -> None` |
 | `agent_run_end` | After agent run | `(agent_name, model_name, session_id=None, success=True, error=None, response_text=None, metadata=None) -> None` |
 | `load_prompt` | System prompt assembly | `() -> str \| None` |
-| `run_shell_command` | Before shell exec | `(context, command, cwd=None, timeout=60) -> dict \| None` (return `{"blocked": True}` to block) |
+| `run_shell_command` | Before shell exec | `(context, command, cwd=None, timeout=60) -> dict \| None` (return `{"blocked": True}` to block, `{"rewrite": "<new cmd>"}` to transparently transform) |
 | `file_permission` | Before file op | `(context, file_path, operation, ...) -> bool` |
 | `pre_tool_call` | Before tool executes | `(tool_name, tool_args, context=None) -> Any` |
 | `post_tool_call` | After tool finishes | `(tool_name, tool_args, result, duration_ms, context=None) -> Any` |
--- a/code_puppy/tools/command_runner.py
+++ b/code_puppy/tools/command_runner.py
@@ -1096,6 +1096,23 @@ async def run_shell_command(
                 execution_time=None,
             )
 
+    # Apply any command rewrites requested by callbacks.
+    # A callback can return {"rewrite": "<new command>"} to transparently
+    # transform the command before execution (e.g. inject a git trailer,
+    # redact a secret, prepend a corporate proxy). Rewrites are applied in
+    # callback registration order; each one sees whatever the previous ones
+    # produced. A visible info line surfaces the change so users always know
+    # what actually ran -- rewriting must never be sneaky.
+    for result in callback_results:
+        if result and isinstance(result, dict) and "rewrite" in result:
+            new_command = result["rewrite"]
+            if not isinstance(new_command, str) or not new_command.strip():
+                continue  # ignore empty / non-string rewrites defensively
+            if new_command != command:
+                reason = result.get("rewrite_reason", "plugin")
+                emit_info(f"[dim]\U0001f527 command rewritten by {reason}[/dim]")
+                command = new_command
+
     # Handle background execution - runs command detached and returns immediately
     # This happens BEFORE user confirmation since we don't wait for the command
     if background:
--- a/tests/tools/test_command_runner_full_coverage.py
+++ b/tests/tools/test_command_runner_full_coverage.py
@@ -545,6 +545,115 @@ class TestRunShellCommand:
         assert "nope" in result.error
 
     @pytest.mark.asyncio
+    async def test_rewrite_from_callback_replaces_command(self):
+        """A callback that returns {"rewrite": ...} transforms the command.
+
+        Proves the new hook contract: infrastructure for plugins that want
+        to transparently transform commands (secret redaction, proxy
+        injection, attribution trailers) instead of merely blocking them.
+        """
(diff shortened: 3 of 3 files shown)


mpfaffenberger:
I actually personally dislike that Claude Code does this. Better for the plugin marketplace? We need one externally I guess.

ashishsinghi:
Thanks @mpfaffenberger for taking a look.

Totally fair, the "opinionated auto-trailer by default" pattern that Claude Code / Cursor use is exactly the kind of thing that feels great for the vendor and creepy for the user. On reflection I agree the built-in-plugin approach is the wrong shape here.
 
 Let me propose a split of this PR:
 
 Keep upstream (this PR, slimmed way down):
 - The run_shell_command hook extension that lets callbacks return {"rewrite": "<new cmd>"} to transparently transform a shell command (with a visible emit_info notice so it's never sneaky). ~15 lines in command_runner.py + one line in AGENTS.md.
 - This is pure infrastructure — no policy. It just unlocks a class of plugins that can't exist today (secret redaction, corporate proxy injection, path fixup, and yes, attribution trailers), and it's genuinely useful for the external marketplace you mentioned.
 
 Drop from this PR:
 - The git_coauthor plugin itself + its tests + its README. I'll pull that out and ship it as a separate plugin rather than baking it into everyone's install.
 
 Want me to update the PR to just the core hook change? Happy to close and re-open if that's cleaner. And +1 on needing a real external plugin marketplace, that's the natural home for opinionated stuff like this.
 
 Thanks for the honest feedback, better to catch the philosophy mismatch before merge.

mpfaffenberger:
Fully onboard! I thought we did have that hook ability already, but I trust you if you tried to implement it and it wasn't working.

ashishsinghi:
Split done. This PR is now infrastructure-only.

**What's in this PR now:**
- `code_puppy/tools/command_runner.py` — the ~15-line rewrite handling (allow, block, and now rewrite).
- `AGENTS.md` — one-line doc update on the `run_shell_command` hook row.
- `tests/tools/test_command_runner_full_coverage.py` — three new cases: happy path, no-op rewrite (no UI notice), defensive handling of empty/non-string payloads.

**What's dropped:**
- All of `code_puppy/plugins/git_coauthor/` and its dedicated test file. That opinionated policy plugin is out of scope here, it'll ship as a user/marketplace plugin instead of a builtin.

mpfaffenberger:
fix tests plz

ashishsinghi:
Traced those two `test_model_factory.py` failures — they're **pre-existing on `main`** and unrelated to the hook change. Confirmed by checking out `main` clean and re-running:

```
FAILED tests/test_model_factory.py::test_extra_models_json_decode_error
FAILED tests/test_model_factory.py::test_extra_models_exception_handling
```

Both fail on `main` with the identical `assert 0 > 0` (where `0 = len({})`).

**Root cause:** commit `[290e36f](https://github.com/mpfaffenberger/code_puppy/commit/290e36f9ba43f29830779daa26fd64f8cb395366)` — *"feat(models): ship models.json empty to require explicit model setup"* — deliberately reduced the bundled `code_puppy/models.json` to `{}`. But these two tests still assert `len(config) > 0` after loading against a deliberately-broken `extra_models.json` overlay. The tests were written assuming a non-empty base and never updated when the base was intentionally emptied.

**Fix (pushed as a separate second commit `05bf226`):** drop the now-invalid `len > 0` assertion. The tests' primary contract — that a broken overlay file does not crash `load_config` and that a warning is logged — is fully preserved. Kept it as a distinct commit from the hook change so you can cherry-pick, revert, or split it into its own PR if you prefer.

ashishsinghi:
Rebased onto `main` — conflict resolved by dropping my `05bf226` entirely, since your `651a2f7` (*"feat: improve queued turns and shell backgrounding"*) already fixed those same two `test_extra_models_*` tests.

ashishsinghi:
Thank you @mpfaffenberger

mpfaffenberger:
Thank you for the contribution!

PR #616, opened 2026-07-12
fix(retry): retry in-band SSE 5xx (gateway 502 over a 200 stream) + wrap sub-agent runs

JulienEllie:
## The bug

The recurring `[HTTP 502] <!DOCTYPE html> ... Error 502 (Server Error)` crash that faceplants the REPL is **not an HTTP 502**. A gateway (puppy-backend `custom_anthropic` proxy) delivers the upstream failure as an **in-band SSE `error` event over a connection that itself returned HTTP 200**.

The Anthropic SDK builds the `APIStatusError` from the *stream* response (`anthropic/_streaming.py::__stream__` → `_make_status_error(..., response=self.response)`), so the exception is:

- `status_code` = **200** (not 502)
- a plain `APIStatusError` (not `InternalServerError`)
- `body.error.type` = generic `"internal_error"`
- the *only* trace of the real failure is a `[HTTP 502]` marker in the message text

`_is_retryable_one` keyed on `status_code` / snippet / a fixed error-`type` set — all of which miss this shape — so `should_retry_streaming` returned `False` and the slow 1-2-3 retry never fired. The ExceptionGroup-traversal work (#536/#546) couldn't help: every path led to a classifier that said "not retryable."

Proven: `APIStatusError` built from a 200 stream response has `status_code == 200` and `should_retry_streaming` returned `False` on the exact production shape (both bare and ExceptionGroup-wrapped).

## The fix

- **`_embedded_http_status()`** recovers the real upstream status from a `[HTTP NNN]` marker; **`_is_transient_status()`** centralizes the 429/5xx rule.
- **`_is_retryable_one()`** gains a provider-agnostic early check on the embedded marker (**4xx markers still do NOT retry**) + `"internal_error"` added to the Anthropic body-type set.
- **`subagent_invocation`**: wrap `temp_agent.run()` in `streaming_retry()`. This path was the **only unprotected model-stream call** — `run_agent_task` has the decorator, but sub-agents (e.g. `docs-oracle`) ran raw and surfaced the 5xx straight to the REPL.

## Tests

New `TestInBandSSE5xx` in `tests/agents/test_streaming_retry.py`: the 200-status/`[HTTP 5xx]`-marker shape, the ExceptionGroup wrapping, the `internal_error` type, and the 4xx negative case. Full suite: 73 passed.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -18,6 +18,7 @@ preserved verbatim:
 from __future__ import annotations
 
 import asyncio
+import re
 import signal
 import threading
 import uuid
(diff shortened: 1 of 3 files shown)


JulienEllie:
Didn't fully get it the first time. Annoying.

PR #621, opened 2026-07-14
fix(editor): Enter submits a fully-typed slash command in one press

JulienEllie:
## The bug

In the persistent line editor, pressing **Enter** while the completion menu is open **never submitted** — it always accepted the highlighted item and returned. For a **fully-typed** command like `/help` or `/exit`, the auto-selected item equals what you already typed, so `accept()` is a **no-op**: the menu closed but nothing submitted, forcing a pointless **second Enter**.

## The fix

Compare the buffer around `accept()`:

- **Real pick** (buffer changed, e.g. `/mod` → `/model`) → close the menu, don't submit. Classic editor behavior — first Enter picks, second submits.
- **No-op accept** (buffer unchanged, i.e. fully typed) → fall through to submit/newline.

So `/help` + **one** Enter submits.

```python
if self._completion_open():
    before = self._buffer
    self._completion.accept()
    if self._buffer != before:
        return None          # real pick -> don't submit
    # no-op -> fall through to submit
```

Also fixes the same class of bug in the legacy `get_input_with_combined_completion` path (submits when the highlighted completion wouldn't change the buffer).

## Tests

- `tests/messaging/test_line_editor_enter_completion.py` — no-op completion submits on one Enter; a real pick doesn't (needs a second).
- `tests/test_prompt_toolkit_completion.py` — no-op submits, real pick applies, no-menu submits.

Full messaging + completion suite: **968 passed**.

## Notes
- The persistent line editor is a newer input path (the legacy `prompt_toolkit` PromptSession path had different, correct behavior), so this bug only appears with the persistent prompt enabled.
- Independent of #619 (retry profiles) — separate concern, separate PR.

### DIFF
--- a/code_puppy/command_line/prompt_toolkit_completion.py
+++ b/code_puppy/command_line/prompt_toolkit_completion.py
@@ -791,19 +791,32 @@ async def get_input_with_combined_completion(
 
     # Enter behavior depends on multiline mode AND completion-menu state.
     # Priority order:
-    #   1. If the completion menu is open with a highlighted item, just
-    #      accept that completion and close the menu (don't submit). This
-    #      matches how editors like VSCode/Helix behave — Enter on a popup
-    #      = pick, not commit.
-    #   2. Multiline mode: insert a newline.
-    #   3. Default: submit the prompt.
+    #   1. If the completion menu is open with a highlighted item that would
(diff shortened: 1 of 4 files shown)


wkramme:
Thank you! This is an annoying behavior for sure.

PR #626, opened 2026-07-14
Refactor websocket protocol runtime

sskmlm:
## Summary
- split websocket protocol/runtime concerns into smaller focused modules
- add supporting refactors for session persistence, permission plumbing, and db/query layering
- fix websocket session collisions and fail fast on DB init
- fix shell backgrounding timeout test behavior and make test model fallback deterministic

## Testing
- pytest -x -q
- targeted shell backgrounding regression test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ code_puppy/bundled_skills/
 .json
 
 .state/
+.maestro/*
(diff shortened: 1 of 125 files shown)


mpfaffenberger:
20k lines?

Issue #505, opened 2026-06-25
Agent chat: fenced code blocks rendered with trailing-space padding instead of clean newlines


carloscasellas:
# Agent chat: fenced code blocks rendered with trailing-space padding instead of clean newlines

## Summary

When the agent streams a response containing a fenced code block into the TUI, every line inside the block is right-padded with spaces out to the code width so Rich's syntax-highlight background fills the column. It looks fine in the terminal, but the moment you copy/paste a snippet anywhere else (chat, ticket, scratchpad, file) you get a wall of trailing spaces on every line that has to be cleaned up by hand.

Tested on:
- macOS Terminal.app + iTerm2
- Reproduces with any agent response that returns a fenced code block (` ``` `).

## Repro

1. Ask the agent for any code snippet, e.g. *"give me a python hello-world in a code block"*.
2. Wait for the response to render in the TUI.
3. Select + copy the code block.
4. Paste into any plain-text editor.
5. Observe: every line ends with `len(code_width) - len(line)` trailing spaces.

Expected: each line ends with `\n`, no padding.

## Root cause

`code_puppy/messaging/rich_renderer.py` renders agent output with `rich.markdown.Markdown(...)` (lines ~971, 977, 991, 1031, 1041). Rich's default `CodeBlock` element wraps fenced code in a `Syntax` instance whose default `pad` behavior extends the highlight background to the configured code width by padding each line with trailing spaces. That padding survives copy/paste because terminal selection captures the raw character cells, spaces and all.

## Proposed fix

Follow the existing pattern in `code_puppy/messaging/markdown_patches.py` (which already monkey-patches `Markdown.elements["heading_open"]` for left-justified headings). Add a `NoPadCodeBlock` subclass that:

- Builds the same `Syntax` instance Rich would build, **but** with `pad=False` (or by overriding `__rich_console__` to yield trimmed lines).
- Gets registered in `Markdown.elements["fence"]` (and `"code_block"` for indented blocks, if applicable) via the existing idempotent `patch_markdown_headings()` -> rename/extend to e.g. `patch_markdown()` so all our markdown patches live in one entry point.

Pseudo-sketch:

```python
class NoPadCodeBlock(CodeBlock):
    def __rich_console__(self, console, options):
        code = str(self.text).rstrip()
        syntax = Syntax(
            code,
            self.lexer_name,
            theme=self.theme,
            background_color="default",
            word_wrap=False,
            padding=0,
        )
        yield syntax
```

Then:

[...]

PR #629, opened 2026-07-16
Fix: terminal outputs raw text on code blocks

carloscasellas:
Closes issue #505 .

This fix addresses code blocks coming with trailing right spaces, which in my experience messes up the copy/paste workflow. This is mainly found for me through copy/pasting curls to postman or my terminal, where the curl gets broken apart and not treated as one request.

Before fix:
Screenshot 2026-07-15 at 5 03 45 PM
Screenshot 2026-07-15 at 4 57 54 PM

After fix:
Screenshot 2026-07-15 at 4 55 31 PM
Screenshot 2026-07-15 at 4 59 07 PM

The main tradeoff to this fix is that the text loses the background color it gets when copy/pasted into a notes app or something similar to that. Raw text is the output now with no trailing spaces on lines.

### DIFF
--- a/code_puppy/messaging/__init__.py
+++ b/code_puppy/messaging/__init__.py
@@ -30,11 +30,11 @@ Example (new):
 """
 
 # =============================================================================
-# Apply Rich Markdown patches (left-justified headers)
+# Apply Rich Markdown patches (left-justified headers, unpadded code blocks)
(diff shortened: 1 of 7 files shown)


mpfaffenberger:
Thank you!!!!!!

PR #630, opened 2026-07-16
Extend the token usage data reported through agent_run_end hook

dmontroy:
## Problem statement
Code Puppy reports only output tokens through the agent_run_end interface. While useful, this does not provide enough information for completely understanding how tokens are being consumed.

## Proposed solution
Extend the number of token usage fields reported through the agent_run_end hook.

## Benefits
By doing this, then Code Puppy can be registered in token tracking tools like [tokentelemetry](https://tokentelemetry.com) and [tokentracker](https://tokentracker.cc), which report token usage across a variety of tools by scanning local logs provided by each coding agent. Also, an intangible benefit is that Code Puppy support in those tools will help to expand the visibility of the tool.

## Testing
- `uv run ruff check code_puppy/agents/_runtime.py tests/plugins/test_claude_code_hooks_plugin.py`
- `uv run ruff format --check code_puppy/agents/_runtime.py tests/plugins/test_claude_code_hooks_plugin.py`
- `uv run pytest tests/plugins/test_claude_code_hooks_plugin.py -q` (23 passed)

## Scope
This involves adding additional token count collection fields to the agent runtime and exposing them in the metadata object over the agent_run_end hook. Existing plugins that are looking at the output_tokens are not affected.

## Next Steps
The next step after this is to create a plugin that will log data in a format consumable by these applications, but this is the minimum change required to the Code Puppy core that can enable the use of those tools.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -987,7 +987,14 @@ async def _run_with_mcp_impl(
     run_success = False
     run_error: Optional[BaseException] = None
     run_response_text = ""
-    run_usage_output_tokens: Optional[int] = None
+    run_usage_metadata: dict[str, int | None] = {
+        "usage_input_tokens": None,
+        "usage_output_tokens": None,
+        "usage_total_tokens": None,
+        "usage_cached_read_tokens": None,
+        "usage_cached_write_tokens": None,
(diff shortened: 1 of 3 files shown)


mpfaffenberger (approved):
I hope it doesn't break any plugins (it shouldn't)

PR #635, opened 2026-07-16
feat(cli): add --port-base to shift port-probe range

finklang:
Adds three layers of override for the port probe's start port so users can dodge collisions with other local dev servers without editing source:

  1. --port-base PORT CLI arg (per-invocation)
  2. CODE_PUPPY_PORT_BASE env var (per-shell)
  3. port_base key in puppy.cfg (persistent, seeded on first run)

Precedence: CLI > env > puppy.cfg > 8090 default. All three sources funnel through resolve_port_base() + _coerce_port_base() for DRY validation with a consistent warning UX.

Robustness:
  * CLI arg is type=str so bad input is caught by our validator and falls through to the next source rather than argparse-exiting.
  * MIN_PORT_BASE=1024 / MAX_PORT_BASE=65535-PORT_PROBE_WIDTH reject values that would probe outside the 16-bit port range or hit privileged ports the process can't bind anyway.
  * ensure_config_exists() seeds port_base in fresh puppy.cfg so users discover the knob (matches the existing auto_save_session seeding pattern).

Tests: 13 new tests across TestGetPortBase, TestResolvePortBase, and TestPortAvailability covering precedence, bounds, graceful fallback, empty-string handling, exact boundaries, config seeding, and a regression guard on MAX_PORT_BASE + PORT_PROBE_WIDTH <= 65535.

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -199,6 +199,19 @@ async def main():
         ),
     )
     parser.add_argument(
+        "--port-base",
+        type=str,
+        default=None,
+        metavar="PORT",
+        help=(
+            "Starting port for the local HTTP server (searches PORT..PORT+920). "
+            "Bump this if 8090 collides with another local dev server. "
+            "Falls back to $CODE_PUPPY_PORT_BASE or 'port_base' in puppy.cfg (default 8090). "
+            "Invalid values are warned about and ignored -- next source in the "
+            "precedence chain is used instead of crashing."
+        ),
+    )
+    parser.add_argument(
         "command", nargs="*", help="Run a single command (deprecated, use -p instead)"
     )
 
@@ -272,9 +285,13 @@ async def main():
         # Truecolor warning moved to interactive_mode() so it prints LAST
         # after all the help stuff - max visibility for the ugly red box!
 
-    available_port = find_available_port()
+    from code_puppy.config import PORT_PROBE_WIDTH, resolve_port_base
+
+    port_base = resolve_port_base(cli_value=args.port_base)
+    port_end = port_base + PORT_PROBE_WIDTH
+    available_port = find_available_port(start_port=port_base, end_port=port_end)
     if available_port is None:
-        emit_error("No available ports in range 8090-9010!")
+        emit_error(f"No available ports in range {port_base}-{port_end}!")
         return
 
     # Early model setting if specified via command line
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -228,6 +228,19 @@ async def main():
         ),
     )
     parser.add_argument(
+        "--port-base",
+        type=str,
+        default=None,
+        metavar="PORT",
+        help=(
+            "Starting port for the local HTTP server (searches PORT..PORT+920). "
+            "Bump this if 8090 collides with another local dev server. "
+            "Falls back to $CODE_PUPPY_PORT_BASE or 'port_base' in puppy.cfg (default 8090). "
+            "Invalid values are warned about and ignored -- next source in the "
+            "precedence chain is used instead of crashing."
+        ),
+    )
+    parser.add_argument(
         "command", nargs="*", help="Run a single command (deprecated, use -p instead)"
     )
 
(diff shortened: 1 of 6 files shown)


thomwebb (changes requested):
## Request changes: reproducible `UnboundLocalError` crash on the default interactive launch path

Nice design on `resolve_port_base()` / `_coerce_port_base()` — precedence chain, bounds validation, and graceful fallback on garbage input all check out, and the `MAX_PORT_BASE + PORT_PROBE_WIDTH <= 65535` regression guard is a nice touch. But there's a blocking bug in `cli_runner.py` that breaks the CLI's most common invocation path.

### The bug

`code_puppy/cli_runner.py` already imports `t` at module scope (line 24):

```python
from code_puppy.i18n import t
```

...and uses it earlier in `main()` (interactive banner: `t("cli.loading")`, `t("cli.banner.powered_by")`, `t("cli.banner.observability_pitch")`). This PR then adds, later in the *same function*:

```python
from code_puppy.config import PORT_PROBE_WIDTH, resolve_port_base
from code_puppy.i18n import t
```

Because Python determines a name's scope for the *entire function body* at compile time, this second `import t` makes `t` local to all of `main()` — including the earlier uses, which now execute *before* the local `t` is ever bound. Confirmed both statically and at runtime:

```
$ ruff check code_puppy/cli_runner.py
F823 Local variable `t` referenced before assignment
   --> code_puppy/cli_runner.py:310:33

$ python -c "
import asyncio, sys
sys.argv = ['code-puppy']
from code_puppy.cli_runner import main
asyncio.run(main())
"
Traceback (most recent call last):
  ...
UnboundLocalError: cannot access local variable 't' where it is not associated with a value
```

That's a plain `code-puppy` invocation with no flags — i.e. this crashes on the default interactive entry point for everyone.

### Suggested fix

Drop the redundant `from code_puppy.i18n import t` inside `main()` (it's dead weight — `t` is already available at module scope). While there, consider hoisting `resolve_port_base` / `PORT_PROBE_WIDTH` to the existing top-level `from code_puppy.config import (...)` block too, since `config` is already imported there and there's no circular-import concern for this function.

### Why the new tests didn't catch it

The 13 new tests (`TestGetPortBase`, `TestResolvePortBase`, `TestPortAvailability`) all unit-test `resolve_port_base`/`get_port_base` in isolation — none of them exercise `main()` end-to-end through the banner path where the shadowing bites. Worth adding (or noting as a gap) an integration-style smoke test that runs `main()` with no `-p` flag, since that's exactly the kind of scope-shadowing bug unit tests in isolation can't see.

I'd also flag: `ruff check` (via `lefthook.yml`'s pre-commit hook) does catch this with its default rule set — but pre-commit hooks are skippable/opt-in locally. If CI isn't already running `ruff check` as a required gate independent of the hook, this is a good argument for adding one.

### Minor, non-blocking nit
`cli.error.no_ports` got new `{port_base}`/`{port_end}` placeholders in `en-US.json` only. Not a regression introduced by this PR (the key was English-only before, and `t()` falls back safely), but since the file's already being touched, a parity pass on `es.json`/`fr-CA.json` would be a nice freebie.

Everything else here (bounds checks, precedence order, `puppy.cfg` seeding matching the existing `auto_save_session` pattern, `type=str` on the argparse arg to avoid a hard argparse exit on bad input) looks solid. Once the scope-shadowing bug is fixed this should be good to go.

finklang:
thanks I'll look at this.   I'm running this version locally I'm wondering why this didn't trigger.

finklang:
puppy says this is already fixed, here is what it says:

Fixed in 7a19ad67.

The reported crash was a real bug — a redundant local from code_puppy.i18n import t inside main() shadowed the module-level import (line 24), making t local for the whole function and breaking the earlier banner calls (UnboundLocalError). Leftover from an early revision predating upstream's own module-level t import.

Removed the duplicate import. Verified on a fresh clone of this branch:
• ruff check code_puppy/cli_runner.py → clean
• python -c "...; asyncio.run(main())" (the repro above) no longer raises UnboundLocalError
• 227/227 targeted tests pass

Please re-pull/re-diff against 7a19ad67 — the version you reviewed was stale.

thomwebb (approved):

PR #639, opened 2026-07-17
Polish base Spanish catalog and deprecate es-419

ellysarai13-blip:
-This polishes the base Spanish catalog and deprecates es-419 per guidance (overrides only regional files, es-419 folded into base. 
changes: Inclusive welcome , recovered the fetch dog pun, 'valor predeterminado' over the 'valor de reserva', adopted es-419's 'configurado en'. 

- es-419.json deleted: it differed from base es in only 2 keys, both stylistic rather than regional preferred phrasings folded into es. The fallback chain (es-AR → es-419 → es) skips missing catalogs, so removal is clean.
Regional files (es-MX/AR/CL/CO) left empty nothing in the current 17 keys warrants a regional override.
Side note: en-US has "code puppy" lowercase in version.update_available but capitalized in startup.welcome kept source parity in Spanish, flagging just in case you want to normalize the English.


mpfaffenberger (approved):

ellysarai13-blip:
updated the i18n foundation tests to reflect the es-419 deprecation, the umbrella tests now verify fallback through base es.

mpfaffenberger:
At first I was like "oh snap, do you speak Polish too?"

PR #644, opened 2026-07-18
fix(i18n): stop routing Latin American Spanish through removed es-419

thomwebb:
## What & why

PR #639 folded the `es-419` Latin American Spanish umbrella into base `es.json` and **deleted `locales/es-419.json`** — but `code_puppy/i18n/locale.py` still mapped every Central/South American dialect (`es-MX`, `es-AR`, …) to `es-419` as their CLDR parent via `PARENT_LOCALES`.

Net effect: the resolver kept **probing the now-missing `es-419.json`** on every Latin American lookup (harmless-but-wasteful `FileNotFoundError` swallowed by the loader), and the documented fallback chain no longer matched reality.

## The fix

- Empty out `PARENT_LOCALES` so LatAm dialects truncate straight to `es`:
  - `es-MX -> es -> en-US` (previously `es-MX -> es-419 -> es -> en-US`)
- **Regional override files are untouched** — `es-MX/AR/CO/CL.json` still sit on top of base `es`.
- `PARENT_LOCALES` is kept as an **empty, documented CLDR override seam** (with a one-liner showing how to reintroduce a non-truncation parent), so the generic `fallback_chain` mechanism and its tests stay intact.
- An explicit `es-419` request still degrades gracefully (`es-419 -> es -> en-US`).

## Also updated

- `docs/I18N.md` — locale table, dialect fallback diagram, directory tree, and the `PARENT_LOCALES` guidance.
- `code_puppy_agent/SKILL.md` — i18n reference section.
- `tests/i18n/test_i18n_foundation.py` — fallback-chain assertions + a few now-misleading test names/comments ("umbrella"/"es419").

## Verification

- `uv run pytest tests/i18n/ -q` → **86 passed**
- Runtime spot check:
  ```
  es-MX  -> ['es-MX', 'es', 'en-US']
  es-419 -> ['es-419', 'es', 'en-US']
  PARENT_LOCALES = {}
  es-MX confirm.yes = Sí   # regional override still resolves via base es
  ```

### DIFF
--- a/code_puppy/i18n/locale.py
+++ b/code_puppy/i18n/locale.py
@@ -28,40 +28,17 @@ _POSIX_LOCALE_VARS = ("LC_ALL", "LC_MESSAGES", "LANG", "LANGUAGE")
 _LOCALE_RE = re.compile(r"^[A-Za-z]{2,3}(?:[_-][A-Za-z0-9]+)*")
 
 # CLDR parent-locale overrides: locales whose fallback parent is NOT simply
-# their truncated form. The big one for us is Latin American Spanish, where
-# every Central/South American country falls back through the ``es-419``
-# umbrella before reaching generic (Iberian-ish) ``es``.
+# their truncated form (see :func:`fallback_chain`). Source: CLDR
+# supplementalData <parentLocales>. Kept as an explicit, testable map so a
+# plugin / the private fork can register a non-truncation parent in one line.
 #
-# Source: CLDR supplementalData <parentLocales>. Kept as an explicit map so
-# adding a dialect is a one-line change and the behavior is testable.
-_LATIN_AMERICAN_ES = (
-    # Central America (+ Mexico)
-    "es-MX",
-    "es-GT",
-    "es-HN",
-    "es-SV",
-    "es-NI",
-    "es-CR",
-    "es-PA",
-    "es-BZ",
-    # South America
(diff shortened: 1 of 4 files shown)


thomwebb:
## Adversarial review

**LGTM overall — two items worth verifying before merge:**

### 1. User with `locale: es-419` in config silently degrades

If anyone has `locale: es-419` persisted in `puppy.cfg`, after this PR their fallback chain becomes `["es-419", "es", "en-US"]`. The `_load_file` call silently swallows the `FileNotFoundError` for the now-missing `es-419.json`, so they fall through to `es` transparently. Functionally acceptable — but it's an undocumented behavioral change worth a note in `PARENT_LOCALES`' docblock or the changelog for anyone with that locale pinned.

### 2. Content migration verification (out-of-scope here, but flag before shipping)

This PR drops the routing; PR #639 (already merged) deleted `es-419.json`. The implicit assumption is that `es.json` already contains everything that was in `es-419.json`. If not, es-MX/es-AR users silently regress from Spanish to English for any key that existed only in `es-419`. Worth a quick diff of `es-419.json`'s former contents against `es.json` before this lands to confirm no net content loss.

---

No blocking issues. `PARENT_LOCALES = {}` is clean, the docstring is accurate, and the test renames correctly reflect the new `es-MX → es → en-US` chain.

PR #645, opened 2026-07-18
feat(i18n): static extraction audit tool for CLI strings

thomwebb:
## Why

The i18n extraction workstreams (PUP-480 public CLI, PUP-481 private CLI) have to migrate **~2,000 `emit_*` / `console.print` call sites** from raw literals to `t()` catalog keys. Grinding that by hand is how strings get missed and PRs become un-reviewable. This gives us a **number**.

## What

`code_puppy/i18n/audit.py` — an AST-based scanner that classifies every user-facing emit call site by its first positional argument:

- **extracted** — already a `t()` / `i18n.t()` / `ngettext()` / `lazy()` call
- **raw** — a hard-coded literal (str constant, f-string, or concatenation) → still needs extraction
- **dynamic** — the arg is a bare variable/expression → no literal to extract (reported, not blamed)

Coverage = `extracted / (extracted + raw)`.

```bash
python -m code_puppy.i18n.audit                 # summary + coverage %
python -m code_puppy.i18n.audit --top 15        # worst files to attack first
python -m code_puppy.i18n.audit --list          # every raw site (file:line)
python -m code_puppy.i18n.audit --json          # machine-readable
python -m code_puppy.i18n.audit --fail-under 25 # CI gate (exit 1 if under)
```

Stdlib-only, so it runs as a lightweight CI gate without importing the app.

## Baseline it reports on `main` today

```
extracted (t/ngettext/lazy) : 6
raw literals  (TODO)        : 1383
dynamic (no literal)        : 255
coverage                    : 0.4%

Top files by raw strings:
   59  code_puppy/cli_runner.py
   49  code_puppy/command_line/config_commands.py
   47  code_puppy/command_line/core_commands.py
   ...
```

## Tests & docs

- `tests/i18n/test_i18n_audit.py` — 19 tests: classification (raw / f-string / concat / extracted / dynamic), callee detection (`emit_*` variants, `console.print`, `self.console.print`, ignoring `logging`/`print`/defs), coverage math, the `--fail-under` gate, JSON output, and a smoke test over the live tree.
- `docs/I18N.md` — new *Measuring extraction progress* section; pairs the static audit with the existing runtime pseudolocalization coverage test.
- `uv run pytest tests/i18n/ -q` → **105 passed**. `ruff check`/`format` clean.

## Notes

Purely additive — no runtime/behavior change to Code Puppy itself; this is developer tooling + CI plumbing for the extraction effort. Complements (doesn't overlap) the runtime pseudoloc gate, which proves a *migrated* module leaks nothing.

### DIFF
--- a/code_puppy/i18n/audit.py
+++ b/code_puppy/i18n/audit.py
@@ -1,0 +1,298 @@
+"""Static i18n extraction audit for CLI/TUI user-facing strings.
+
+The extraction workstreams (PUP-480 public CLI, PUP-481 private CLI) have to
+migrate ~2,000 ``emit_*`` / ``console.print`` call sites from raw literals to
+``t()`` catalog keys. Grinding that blind is how strings get missed and PRs
+become un-reviewable, so this module turns "how much is left?" into a number.
+
+What it does
+------------
+Walk the source tree, parse each module with :mod:`ast`, and classify every
+user-facing emit call site by its *first positional argument*:
+
+* **extracted** - the text is already a translation call
+  (``t()`` / ``i18n.t()`` / ``ngettext()`` / ``lazy()`` / ``gettext()``).
+* **raw** - the text is a hard-coded literal (``str`` constant, f-string, or
+  string concatenation). These are the sites still needing extraction.
+* **dynamic** - the argument is a bare variable / other expression, so the
+  string was produced elsewhere. Neither credited nor blamed; reported
+  separately so a reviewer can eyeball them.
+
+Coverage = ``extracted / (extracted + raw)``. Dynamic sites are excluded from
+the denominator because there is no literal here to extract.
+
+Usage
+-----
+::
+
+    python -m code_puppy.i18n.audit                 # human summary
+    python -m code_puppy.i18n.audit --list          # + every raw site
+    python -m code_puppy.i18n.audit --top 15        # worst 15 files
+    python -m code_puppy.i18n.audit --json          # machine-readable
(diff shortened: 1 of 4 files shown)


thomwebb:
## Adversarial review

**LGTM — one latent test fragility worth fixing now:**

### 1. Smoke test will fail the moment extraction is complete

```python
assert len(report.raw) > 0  # extraction is still very much in progress
```

This is a "test that passes until the feature is done" anti-pattern. When the extraction work succeeds — the entire goal of PUP-480 — `report.raw` could reach zero and this assertion fails at exactly the wrong time. The comment even says so itself. Suggest removing the raw-count assertion and keeping only what's already on the next line:

```python
assert report.sites, "expected to find user-facing emit sites"
assert 0.0 <= report.coverage <= 100.0
```

The coverage bound is sufficient to prove the tool ran and produced a sensible number.

### 2. Minor: `preview` is populated but absent from `--json` output

`Site.preview` is computed and shown in `--list` mode, but the JSON payload only includes `{path, line, call}`. Any CI script or jq consumer has no way to see the string content without re-reading the source file. Not blocking — but worth a note in the CLI help or a future `--json` enhancement.

---

The AST classification logic, coverage math, `--fail-under` gate, directory walker, and integration smoke are all solid. Stdlib-only is the right call for a CI gate.

thomwebb:
Fixed in latest push — dropped the `assert len(report.raw) > 0` line. Coverage bounds (`0.0 <= report.coverage <= 100.0`) are sufficient proof the tool ran.

PR #646, opened 2026-07-18
feat(i18n): extract cli_runner.py user-facing strings (PUP-480)

thomwebb:
## What

Migrates **every user-facing `emit_*`/`console.print` string in `cli_runner.py`** to the `t()` catalog — the biggest single extraction target (it topped the audit at 59 raw sites). Covers: startup/loading, model & agent selection + errors, session resume, the interactive help block, initial-command handling, prompt_toolkit bootstrap, autosave load, cancel/goodbye paths, attachment summaries, and the headless `-p` flow.

## Result

- **Static audit for `cli_runner.py`: 59 raw → 1.** The remaining one is `emit_info(f"{user_prompt}\n")` — pure variable interpolation (the agent's own prompt text + a newline), with **no translatable literal**. Not a real miss.
- **61 `cli.*` keys** added to `locales/en-US.json` (76 total now).

## Deliberate scope calls

- **Source-catalog only.** I added keys to `en-US.json` and let `es`/`fr-CA` **fall back to English** rather than machine-translating ~61 strings. Quality translation is the dedicated translator workflow (PUP-483); the coverage tests only forbid *extra* keys in translations, so fallback is fully valid.
- **Precomputed inline params.** The safe `{identifier}` interpolation grammar can't evaluate expressions, so `type(exc).__name__`, `', '.join(available_models)`, etc. are computed into named params (`error_type`, `models`, …).
- **Left developer-only strings alone** — `log_error(...)` diagnostic context is internal telemetry, not user-facing output, so it stays in English (matches the SKILL i18n rules).

## Tests

- `tests/i18n/test_cli_runner_i18n.py` (6 tests) locks in the whole `cli.*` namespace: every key resolves to real source text, **pseudolocalizes** (proves it flows through the runtime), placeholders substitute with no leftovers, and the migrated module imports cleanly. A full pseudoloc run of the giant async `main()` is impractical, so the namespace guard is the pragmatic proof.
- **All 59 existing `cli_runner` tests still pass**; full `tests/i18n/` green.
- `ruff check`/`format` clean.

## Independence

Touches only `cli_runner.py`, `en-US.json`, and the new test — **no overlap** with the es-419 PR (#644) or the audit-tool PR (#645).


thomwebb:
## Adversarial review

**LGTM — two issues worth addressing, one minor:**

### 1. Embedded `\n` in catalog values bakes formatting into translatable text

Several keys have leading newlines hard-coded in the catalog string:

```json
"cli.input.cancelled":  "\nInput cancelled",
"cli.goodbye_ctrld":    "\nGoodbye! (Ctrl+D)",
"cli.turn.cancelled":   "\nCancelled"
```

And `cli.autosave.loaded` spans two lines:
```json
"cli.autosave.loaded": " Autosave loaded: ... \n From: {path}"
```

Translators working on these strings won't know the `\n` is intentional formatting they must preserve — and many CAT tools strip or warn on whitespace-only differences. The i18n best practice is to own the newline at the call site (the caller emits the blank line) and keep catalog values as pure message text. Functional behaviour is unchanged in English since the strings are preserved verbatim, but it'll cause translator confusion (and potentially broken layout) in other locales.

### 2. `importlib.reload(cli_runner)` re-triggers `apply_all_patches()`

`test_cli_runner_imports_cleanly` calls `importlib.reload()` on a module whose top-level code runs `apply_all_patches()`. Re-applying pydantic patches a second time in the same process is probably idempotent in practice, but it's undefined behaviour for patch functions designed to run once at startup. A simpler guard:

```python
import code_puppy.cli_runner as cli_runner
assert hasattr(cli_runner, "interactive_mode")
```

Drop the `importlib.reload` — the initial import already proves the module loads cleanly, and the attribute check is still meaningful.

---

Extraction coverage (59 → 1) and catalog key correctness all look good. The one remaining raw site (`emit_info(f"{user_prompt}\n")`) is correctly identified as non-translatable.

thomwebb:
All three issues addressed in latest push:

1. **Embedded `\n`** — stripped from `cli.input.cancelled`, `cli.goodbye_ctrld`, `cli.turn.cancelled`; newline now prepended at each call site.
2. **`cli.autosave.loaded`** — split into two independent keys (`cli.autosave.loaded` + `cli.autosave.loaded_path`); call site emits two separate messages so no newline lives in either catalog string.
3. **`importlib.reload()`** — dropped from `test_cli_runner_imports_cleanly`; initial import is sufficient proof, and reload was silently re-triggering `apply_all_patches()`.

PR #647, opened 2026-07-18
feat(i18n): extract config_commands.py user-facing strings

thomwebb:
## Stack order

> **Merge #646 first.** Both PRs add keys to `en-US.json` from the same base; merging #646 first and then rebasing this branch removes the conflict automatically (different key namespaces, zero logic overlap).

---

## What

Migrates the user-facing emit_* strings in `command_line/config_commands.py` to the `t()` catalog — the #2 extraction target at **49 raw audit sites** (36 genuinely translatable; 13 are false-positive f-strings with only dynamic content).

Covers four commands: `/set`, `/pin_model`, `/unpin`, `/diff`, and `/colors`.

## Result

- **49 raw → 13** — all remaining hits are structural loop formatters with no English text (`f"  {model}"`, `f"  {agent} - {desc}"`, etc.)
- **28 `cfg.*` keys** added to `en-US.json` (43 total on this branch)
- Shared listing headers (`cfg.listing.available_models/builtin_agents/json_agents`) are reused across both `/pin_model` and `/unpin` — no duplication

## Key groupings

| Namespace | What |
|---|---|
| `cfg.set.*` | `/set`: key-required, yolo-unchanged, success, agent-reloaded, reload-failed |
| `cfg.listing.*` | Shared listing headers (models, built-in agents, JSON agents) |
| `cfg.model.*` | Model not-found + available-list |
| `cfg.agent.*` | Agent not-found |
| `cfg.pin_model.*` | Usage, success, agent-reloaded, reload-failed, general failure |
| `cfg.unpin.*` | Usage, success, agent-reloaded, reload-failed, general failure |
| `cfg.diff.*` | Apply-failed error |
| `cfg.colors.*` | Saved, apply-failed, all-available header, usage hint, notes |

## Deliberate scope calls

- **Source-catalog only.** `es`/`fr-CA` fall back to English. Machine-translating 28 new keys risks quality; that's PUP-483's job.
- **13 remaining raw hits are false positives.** They are loop-emitted f-strings with only dynamic agent/model/path variables (no English literals) — extracting them would produce untranslatable catalog entries.
- **`{color_type}`** in `cfg.colors.usage` is a literal placeholder in the catalog (not a param). The safe interpolation engine leaves unknown fields intact, so `t("cfg.colors.usage")` renders as `\nUsage: /diff {color_type} <color_name>` — correct behaviour.

## Tests

`tests/i18n/test_config_commands_i18n.py` (6 tests): cfg.* namespace resolves to source text, pseudolocalizes, placeholders substitute cleanly, module imports without errors. All 86 existing i18n tests still pass. `ruff` clean.

## Independence

Touches only `config_commands.py`, `en-US.json`, and the new test — **no logic overlap** with #644 (locale.py), #645 (audit tool), or #646 (cli_runner). The only merge consideration is `en-US.json` key ordering — resolved by merging #646 first.


thomwebb:
## Adversarial review

**One real bug, one repeated issue from #646:**

### 1.  `cfg.colors.usage` should use `{{color_type}}` not `{color_type}`

The catalog value is:
```json
"cfg.colors.usage": "\nUsage: /diff {color_type} <color_name>"
```

The translation engine's `_FIELD_RE` regex matches any `{identifier}` and attempts substitution. Since no `color_type` param is ever passed to `t("cfg.colors.usage")`, the engine's forgiving fallback leaves it intact — so it *works* today. But:

1. **Future breakage:** if any call site ever passes `color_type=...` for another reason, the placeholder silently substitutes with the wrong value.
2. **Translator confusion:** translators see `{color_type}` and reasonably assume it's a runtime param they should preserve, not display text they should translate. A French translator might write `Utilisation: /diff {type_couleur} <nom_couleur>` — and now it never renders.
3. **The `test_no_leftover_placeholder_for_supplied_params` test actually catches this:** it substitutes `color_type="X"` and the placeholder disappears, which means the test is silently validating the *wrong* behaviour.

Fix: escape as `{{color_type}}` in the JSON — the engine converts `{{` → `{` during rendering, so the output is still the literal text `{color_type}` the user sees.

### 2. Embedded `\n` in listing headers (same issue as #646)

```json
"cfg.listing.builtin_agents": "\nAvailable built-in agents:",
"cfg.listing.json_agents":    "\nAvailable JSON agents:",
"cfg.colors.all_available":   "\n All Available Rich Colors:",
"cfg.colors.usage":           "\nUsage: /diff {color_type} <color_name>"
```

Same translator-confusion/CAT-tool issue flagged on #646. The leading `\n` is presentation, not message content — move it to the call site.

---

Everything else (pin/unpin key groupings, shared `cfg.listing.*` reuse, DRY catalog design, all tests passing) looks solid. The `{{color_type}}` fix is the only blocking item.

thomwebb:
All issues addressed in latest push:

1. **`{{color_type}}` escape** — `cfg.colors.usage` now uses `{{color_type}}` so the engine renders it as the literal string `{color_type}` rather than treating it as a substitution slot. Added `test_colors_usage_renders_as_literal_placeholder` to lock this in explicitly.
2. **Embedded `\n` in headers** — stripped from `cfg.listing.builtin_agents`, `cfg.listing.json_agents`, `cfg.colors.all_available`, and `cfg.colors.usage`; newline prepended at all 8 call sites instead.
3. **`test_no_leftover_placeholder_for_supplied_params`** — updated to skip entries containing `{{` escapes, since their rendered output intentionally contains literal braces that would otherwise trip the assertion.

thomwebb:
**Root cause of the CI failures + what was fixed**

The GitHub "Update branch" merge (01cc9cc7) brought in #644, #645, and #646 but landed *just before* Mike's follow-up commit `0ed6911b` ("test: fix 5 stale tests broken by empty-by-default models.json") hit `main`. So the branch was running against tests that assumed a bundled default model -- which no longer exists.

**Resolution:** rebased the two branch commits cleanly onto current `main` (37c78a61) instead of patching the botched merge:

- `en-US.json` conflict resolved: kept all 61 `cli.*` keys (from #646) and all 28 `cfg.*` keys (from this PR)
- `0ed6911b` is now part of the base, so all 5 previously-failing tests pass
- Verified locally: 118 i18n tests + 987 broader tests (inc. the 4 test files from `0ed6911b`) -- all green

No content changes -- the force-push is a pure rebase of the same two commits.

PR #658, opened 2026-07-20
chore: ignore local .worktrees/ directory

thomwebb:
One-liner: adds `.worktrees/` to `.gitignore` so contributors using `git worktree add .worktrees/<name>` for parallel branch development don't have to worry about worktree scaffolding leaking into commits or cluttering `git status`.

No behavior change; ignore-only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ code_puppy/bundled_skills/
 .json
 
 .state/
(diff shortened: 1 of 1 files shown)


thomwebb:
looks like it was already added to .gitignore. closing

PR #661, opened 2026-07-21
feat: Code_Puppy destructive command authZ engine

collin-kierce:
This PR aims to improve the way destructive commands are handled by code-puppy. This is meant to help keep users from accidentally running malicious commands on their devices.
**What is currently in place:**
- A cheap substring search looking for key commands such as "rm", "format", "rmdir"
- A "is_real_invocation" function which checks to see if the found keyword is the first word in the command
- An expensive regex matching to confirm wether this is a malicious command
- If any of these three stages don't find anything suspicious the command will run fine. If all three stages agree it's a harmful command, the user will be alerted about the harmfulness of the command and prompted for permission to proceed or stop.
- Two different callbacks/guardrails for destructive commands and git commands
- All data regarding commands and substrings is stored inside file
- iterates through all expensive regex patterns looking for a match

**What this PR does:**
- Start by normalizing command, removes common obfuscation methods such as empty "" and ''
- Splits command on compound operators such as && or || to check all parts of compound command
- Conducts a cheap search for keywords in the command/commands
- Does not do a "is_real_invocation" check as this was leading to false negatives with commands such as sudo rm -rf /
- Matches regex patterns to see if there is a hit or not
- If there is a regex match the command will either be blocked immediately or the user will be prompted for permission depending on what the command is. A command like rm -rf / is blocked immediately
-  Merges Git guardrail and destructive command guardrail into one
- Loads all data from JSON files inside new directory called patterns 
- Separates expensive regex patterns by categories and only searches through categories found in cheap substring search


thomwebb:
@collin-kierce plugins have been moved to https://github.com/mpfaffenberger/code_puppy_core_plugins. please re-open there

PR #668, opened 2026-07-23
perf(theme): defer heavy imports until user actions need them

thomwebb:
## Summary

Move eight sibling imports in `code_puppy/plugins/theme/register_callbacks.py`
from module top-level into the functions that actually use them, and
replace one eagerly-registered callback with a lazy shim.

The theme plugin's startup callback only needs to apply a default theme
on first run. Everything else — `.picker`, `.themes`,
`.prompt_toolkit_theme`, `.bundled_palettes`, plus a handful of heavier
`code_puppy` internals — is only needed when the user runs `/theme` or
triggers a render.

## Measured impact

Honest numbers — this is a modest win:

| Metric | Before | After |
|---|---:|---:|
| `register_callbacks` isolated import (warm) | ~21 ms | ~11 ms |
| `register_callbacks` isolated import (cold) | ~82 ms | ~53 ms |
| End-to-end `--version` (warm) | 1.78 s | 1.75 s |

The end-to-end saving is within measurement noise (~30 ms). Verified via a
`sys.modules` snapshot after `plugins.load_plugin_callbacks()` completes:

```
code_puppy.plugins.theme.themes:               not loaded (good)
code_puppy.plugins.theme.picker:               not loaded (good)
code_puppy.plugins.theme.prompt_toolkit_theme: not loaded (good)
code_puppy.plugins.theme.bundled_palettes:    not loaded (good)
```

## Scope caveat (honest correction)

An earlier draft of this PR claimed the theme plugin "no longer
transitively pulls in `prompt_toolkit` at plugin-discovery time." That
claim was **false** and has been retracted from the in-file comment and
this description.

Repro:
```python
from code_puppy.callbacks import register_callback  # baseline
before = set(sys.modules)
import code_puppy.plugins.theme  # just the package
# → 115 prompt_toolkit.* modules already loaded
```

`code_puppy/plugins/theme/__init__.py` eagerly imports `.content_styles`,
`.osc_palette`, and `.rich_themes`, and calls `reapply_from_config()` on
each at package-import time. That path transitively pulls in
`code_puppy.messaging.rich_renderer`, which loads ~115 `prompt_toolkit`
modules. Nothing this PR does to `register_callbacks.py` can shortcut
that — those modules are already resident by the time `register_callbacks`
is imported.

The measured 10-30 ms savings this PR delivers therefore come from
deferring `.picker` (TUI widgets), `.themes` (large color-menu tables),
and `.prompt_toolkit_theme` — not from avoiding `prompt_toolkit`
altogether. A future PR could defer the `__init__.py` reapply sweep to
a startup callback, which would recover the `prompt_toolkit` cost for
`--version` / `--help` invocations that skip startup. Out of scope here.

## Why we did it anyway

- Establishes a clean lazy-import pattern for other plugins to follow.
- The measured 10-30 ms is real, if modest.
- Test suite got stronger: 9 white-box patches previously targeted
  `register_callbacks.X` (fragile — patching a re-import
  point instead of the source). They now target the actual source
  modules, which is the healthier shape.
- Adds a `_prompt_toolkit_style` lazy shim with a strict signature
  (`(style: BaseStyle | None) -> BaseStyle`) so
  `merge_with_active_style` gets called on-demand rather than requiring
  its module at registration time.

## Tests

- All 110 theme-related tests pass (`test_theme_plugin.py` +
  `test_prompt_toolkit_theme.py`).
- Retargeted 9 white-box patches from `register_callbacks.X` (which no
  longer exists at module scope after this change) to the actual source
  modules (`code_puppy.config.get_value`,
  `code_puppy.plugins.theme.osc_palette.get_saved_palette`, etc.).
- Added two new tests covering the `_prompt_toolkit_style` shim: one
  verifies it delegates its single argument to
  `merge_with_active_style` (regression guard against silent signature
  widening), and one confirms the shim preserves `__name__` for error
  logging.

## Broader startup-perf context

While profiling this I measured the import cost of each plugin's
`register_callbacks` in isolation (snapshot `sys.modules` before/after,
bucket the delta by heavy library). A few results worth passing along:

| Plugin | Modules added | Notable heavies |
|---|---:|---|
| `claude_code_oauth` | +2173 | openai(557) anthropic(445) pydantic_ai(56) playwright(53) |
| `theme` (before this PR) | +940 | pydantic_ai(52) prompt_toolkit(115) rich(60) |
| `aws_bedrock` | +932 | pydantic_ai(52) playwright(53) |
| `obsidian_agent` | +378 | pydantic_ai(19) mcp(82) |

Takeaways for future PRs (not part of this change):

- **`pydantic_ai` (~52 modules) loads per-plugin.** A single root import
  point is the leverage lever — likely via the type imports referenced
  at module scope in the agent-registration API.
- **Two plugins unexpectedly pull in `playwright`** (`aws_bedrock`,
  `claude_code_oauth`). Neither does browser automation. Likely a
  `code_puppy.tools`-registry-level bug where registering browser tools
  requires importing playwright.
- **`claude_code_oauth` pulls in the entire OpenAI SDK** (557 modules).
  A plugin for Claude authentication should not need OpenAI's Python
  client.
- **The dominant wall-clock cost at startup is the synchronous pypi.org
  version check** — ~0.8 s median on a healthy network, up to
  10-plus seconds when pypi.org is slow or blocked by a corporate
  proxy. Kicking it to a background task is the single biggest UX lever
  and doesn't require touching plugin structure.

Happy to file follow-up issues on this repo for any of the above if
they'd be useful — I've been tracking them internally.

## Files touched

- `code_puppy/plugins/theme/register_callbacks.py` — 8 sibling imports
  moved into functions; `_prompt_toolkit_style` lazy shim added; NOTE
  comment rewritten for accuracy.
- `tests/plugins/test_theme_plugin.py` — patch targets retargeted from
  `register_callbacks.X` to source modules; two new tests for the
  shim added.


kvandre12-commits:
This is the best one! This should open the door! Its the heavy browser
automation and some early calls that kept the puppy off my phone --
probably off servers too -- maybe even smaller devices and robots? At that
point I
A)used brave as a default browser to restore some browser capabilities but
it still felt like we were crawling pixel by pixel with a flashlight(would
love cells input)
B)abd cdp bridge connection. A made a small governance layer on my phone so
I dont blow it up with your tech; and this is the route I use to keep
things on the up and up(playing but text editor works great too here, code
puppy import pages and use that to get a bit of the layout)
C)own the viewer. Code puppy is the engine. Literally just make the browser
next for these smaller devices.

Kudos friends, you guys rock!

PR #674, opened 2026-07-24
PUP-544: Return per-run token usage (with cache-token breakdown) and run timestamps from sub-agent invocation tools

AndrewTilson:
## Summary

`invoke_agent` and `invoke_agent_with_model` now report per-run **token usage
(with a cache-token breakdown)** and **run timestamps** on the success path, so
benchmarking and model-comparison workflows can measure cost directly from the
tool result instead of reconstructing it from BigQuery/OpenObserve telemetry
(which is hard to attribute to individual sub-agent sessions).

Jira: **PUP-544** — https://jira.walmart.com/browse/PUP-544

## Final field set on `AgentInvokeOutput`

All fields are **nullable and additive** (default `None`), populated only on the
success path:

| Field | Meaning |
|---|---|
| `input_tokens` | Regular **non-cached** input tokens |
| `cache_read_input_tokens` | Cached input tokens read (cache hits) |
| `cache_creation_input_tokens` | Input tokens written to cache (Anthropic only) |
| `output_tokens` | Generated/response tokens |
| `total_tokens` | Provider total when present, else sum of the buckets above |
| `num_requests` | LLM request count for the run |
| `start_time` / `end_time` | Timezone-aware UTC ISO-8601 timestamps around the run |
| `duration_ms` | Monotonic wall-clock latency |

## Per-provider cache-token mapping

Verified against the installed `pydantic-ai==1.56.0` + `genai-prices`. The
library folds cached tokens **into** its combined `input_tokens` for every
provider we use, so we subtract them back out to keep `input_tokens` strictly
non-cached (`input + cache_read + cache_creation + output == total`, no
double-counting).

| Provider | cache_read source | cache_creation |
|---|---|---|
| Anthropic | `details["cache_read_input_tokens"]` | `details["cache_creation_input_tokens"]` |
| OpenAI / codex | first-class `cache_read_tokens` attr (nested `prompt_tokens_details.cached_tokens` is not copied into `details`) | None (no such concept) |
| Gemini | `details["cached_content_tokens"]` | None (no such concept) |

For OpenAI, a genuine `cache_read == 0` surfaces as `None` (the attribute default
is indistinguishable from "not reported") — we never fabricate a `0`.

## What changed

- **`code_puppy/tools/agent_tools.py`** — `AgentInvokeOutput` gains the cache
  buckets + `start_time`/`end_time`; docstring updated.
- **`code_puppy/tools/subagent_invocation.py`** — instrumented once at the shared
  `_invoke_agent_impl`: `_pick_reported_tokens` (details-first, attribute
  fallback, `None`-vs-`0` aware), `_extract_usage_metrics` (per-provider cache
  sourcing + non-cached normalization), UTC timestamps around the run, best-effort
  usage extraction that never breaks the response path.
- **`tests/test_subagent_invocation_usage.py`** — per-provider unit tests using
  the **real** library payload shapes, plus integration tests (success populates
  every field with `start_time <= end_time`, Anthropic creation bucket,
  deterministic duration, `usage()` raising, error-path all-`None`).

## Behavior contract

- **Success:** all fields populated; buckets sum to `total_tokens`.
- **Any error path** (recursion/validation guards, empty model name, run
  exception): all new fields remain `None`; existing `error` unchanged.
- Purely additive — callers ignoring the new fields are unaffected.

## Testing

- `pytest tests/test_subagent_invocation_usage.py` → 15 passed
- Regression across adjacent suites (agent_tools, run_stats) → all green (178 total)
- `ruff check` → clean

## Notes / follow-ups

- No Jira key in source (only PR metadata + commit body).
- `code_puppy/agents/agent_creator_agent.py` (~lines 277-280) still documents the
  old return shape in a prompt string — intentionally out of scope; follow-up.
- PUP-544 is labeled `public` + `private`; the internal fork should pick this up
  via the standard OS-merge process.

### DIFF
--- a/code_puppy/tools/agent_tools.py
+++ b/code_puppy/tools/agent_tools.py
@@ -201,13 +201,25 @@ class ListAgentsOutput(BaseModel):
 
 
 class AgentInvokeOutput(BaseModel):
-    """Output for the invoke_agent tool."""
+    """Output for the invoke_agent tool.
+
+    The token-usage and latency fields are populated on the success path so
+    benchmarking/model-comparison callers can measure per-run cost without
+    reconstructing it from downstream telemetry. They are purely additive and
+    remain ``None`` on every error path, so existing callers that ignore them
+    are unaffected.
+    """
 
     response: str | None
     agent_name: str
     session_id: str | None = None
     model_name: str | None = None
     error: str | None = None
+    input_tokens: int | None = None
+    output_tokens: int | None = None
+    total_tokens: int | None = None
+    num_requests: int | None = None
+    duration_ms: float | None = None
 
 
 def register_list_agents(agent):
--- a/code_puppy/tools/agent_tools.py
+++ b/code_puppy/tools/agent_tools.py
@@ -203,11 +203,17 @@ class ListAgentsOutput(BaseModel):
 class AgentInvokeOutput(BaseModel):
     """Output for the invoke_agent tool.
 
-    The token-usage and latency fields are populated on the success path so
+    The token-usage and timing fields are populated on the success path so
     benchmarking/model-comparison callers can measure per-run cost without
     reconstructing it from downstream telemetry. They are purely additive and
     remain ``None`` on every error path, so existing callers that ignore them
     are unaffected.
+
+    Token accounting is normalized so the input buckets never overlap:
+    ``input_tokens`` counts only regular (non-cached) input, while cached input
+    is reported separately as ``cache_read_input_tokens`` (cache hits) and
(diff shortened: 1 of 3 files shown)


mpfaffenberger (approved):

AndrewTilson:
@mpfaffenberger i didnt intend on this changing the default invoke_agent tool just the one with model. Wasnt sure we wanted to modify the default one. Whats your thoughts on this? I can submit another pr to fix unless you think its fine for the default invoke_agent as well?

mpfaffenberger:
oh yeah we need to fix that

PR #675, opened 2026-07-26
fix(subagents): preserve interrupted sessions so Ctrl-C is resumable

JulienEllie:
## Why this change

Cancelling a runaway subagent should stop the work; it should not erase the
only record that the work ever happened.

Before this change, interrupting an `invoke_agent` run discarded the partial
history and the session ID together:

- Subagent history was saved only after `temp_agent.run()` returned, or on the
  crash path guarded by `except Exception`.
- `asyncio.CancelledError` (what Ctrl-C raises) inherits from `BaseException`,
  so it slipped past `except Exception` entirely and no partial save ran.
- Async teardown can also wrap that cancellation in a `BaseExceptionGroup`,
  which was likewise treated as an ordinary crash.

The intended invariant: once a subagent has made progress, cancellation must not
make that progress untraceable.

## What changed

The invocation error handler is widened from `except Exception` to
`except BaseException`, and cancellation is classified explicitly:

1. Plain `CancelledError`/`KeyboardInterrupt`, and cancellation nested inside a
   `BaseExceptionGroup`, are treated as an interruption (not a crash).
2. On interruption the partial history is persisted through the existing session
   store, a resume hint naming the exact session ID is emitted, and the original
   exception is re-raised. Persistence never converts Ctrl-C into a successful
   return.
3. The partial-save logic is extracted into one shared helper used by both the
   interruption path and the ordinary-failure path (DRY).
4. Non-cancellation `BaseException`s (e.g. `SystemExit`) propagate untouched
   instead of being swallowed into a failure result.

Resume itself is unchanged and already supported: re-invoking with the same
`session_id` loads the saved history via the existing
`_load_session_history` / `_save_session_history` pair. The session ID's hash
suffix is baked in before saving, so the resumed session keeps the exact ID.

## Scope (intentionally minimal)

This is the minimum fix for the interruption-loses-work bug. It deliberately
does not add: a JSONL recovery journal, continuous atomic checkpointing, the
`/subagents list|show|resume` command surface, fork pre-start ID reservation,
panel/lifecycle-message plumbing, or the invocation-module split. Those are
separable enhancements and are not required to stop cancellation from erasing
work.

Known scope boundary: a cancellation that fires before any turn completes has no
history to persist, so it leaves only the on-screen resume hint (no on-disk
record yet). Reserving zero-message metadata up front is out of scope here.

## User-visible outcome

Interrupting a subagent now leaves a permanent message like:

```text
code-reviewer interrupted - 18 message(s) saved. Resume: invoke_agent(session_id='code-reviewer-session-a3f2b1')
```

Ctrl-C still cancels the work. It just stops deleting the record too.

## Testing

Focused regressions added to `tests/test_subagent_invocation_usage.py`
(`TestCancellationPersistence`):

- plain cancellation persists partial history and re-raises;
- grouped cancellation (`BaseExceptionGroup`) is treated as interruption;
- zero-message cancellation reports nothing saved;
- ordinary failure still returns the existing error result (no false warning).

```text
tests/test_subagent_invocation_usage.py .......................... 22 passed
tests/test_agent_tools_coverage.py + tests/agents/test_run_signals_cancel.py  40 passed
ruff check + ruff format: clean
```

Diff: +209 / -26 across 2 files.

### DIFF
--- a/code_puppy/agents/_run_signals.py
+++ b/code_puppy/agents/_run_signals.py
@@ -174,6 +174,54 @@ def prepare_queued_steer_injection(agent: Any, result: Any) -> Optional[Any]:
     return content
 
 
+def inject_interrupted_subagent_notes(agent: Any) -> None:
+    """Tell the agent about any sub-agents interrupted since the last run.
+
+    Ctrl+C cancels both the delegated sub-agent task and the parent run, and
+    the parent's return-less ``invoke_agent`` tool call is pruned from history
+    as a dangling call -- so without this the model would have no memory that
+    it ever delegated. We drain the records left by the sub-agent cancel path
+    and append one plain user-message note per interrupted session (the same
+    injection shape the steer processor uses), which survives the interrupted
+    tool-call prune because it is a valid standalone message.
+
+    Called at run start (never nested) so a foreground cancel surfaces on the
+    user's next turn, and a ``/fork`` cancelled while the agent was idle
+    surfaces on the next run too.
+    """
+    from pydantic_ai.messages import ModelRequest, UserPromptPart
+
+    from code_puppy.tools.subagent_invocation import drain_interrupted_subagents
+
+    if not hasattr(agent, "_message_history"):
+        return
+    records = drain_interrupted_subagents()
+    if not records:
+        return
+
+    injected = []
+    for rec in records:
+        session_id = rec["session_id"]
+        saved = rec["saved_count"]
+        saved_phrase = (
+            f"{saved} message(s) of its work were saved"
+            if saved is not None
+            else "no completed messages had been produced yet"
+        )
+        note = (
+            f"[system note] The sub-agent '{rec['agent_name']}' you invoked was "
+            f"interrupted by the user before it finished; {saved_phrase}. Its "
+            f"partial session is saved as '{session_id}'."
+        )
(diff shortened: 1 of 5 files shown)


mpfaffenberger:
Are we sure this should take 38 files changed? I think what you want shoudl be achievable very easily.

JulienEllie:
You're right, the change came with many bells and whistles that aren't needed. Here is the minimal version of the fix.

JulienEllie:
Follow-up commit `bdcc3001`: close the parent-awareness gap.

Interrupting a sub-agent stopped the work but left the *parent* agent with no memory the delegation happened — Ctrl-C tears down the parent run too, and its return-less `invoke_agent` tool call gets pruned from history as a dangling call.

Now the shared cancel path (foreground `invoke_agent` and detached `/fork` both go through `_invoke_agent_impl`) records each interrupted session, and at the next run start the parent's history gets a plain note naming the agent, saved message count, and the exact `invoke_agent(session_id='...')` resume call. It survives the interrupted-tool-call prune because it's a valid standalone message. A `/fork` cancelled while the agent was idle surfaces on the next run too.

+112 lines across 3 files + tests (single/multi/zero-saved/empty/no-history injection, plus record-on-cancel and no-record-on-crash). Full `tests/agents/` + fork plugin: 855 passed.

PR #684, opened 2026-07-28
Clarify optional model pinning and refresh Agent Creator examples

thomwebb:
## Summary
- clarify that Agent Creator does not require a model pin
- document that omitting `model` uses the global model
- remove stale model-pinning instructions and fix JSON examples
- refresh the model examples to `claude-sonnet-5` and `gpt-5.6-terra`

## Validation
- `uv run pytest -q tests/test_agent_creator_full_coverage.py` (21 passed)
- adversarial code review: PASS
- `git diff --check`

## Notes
- Existing unrelated untracked files were not included.

### DIFF
--- a/code_puppy/agents/agent_creator_agent.py
+++ b/code_puppy/agents/agent_creator_agent.py
@@ -84,8 +84,8 @@ You specialize in:
 3. List ALL available tools so they can see other options
 4. Ask them to confirm their tool selection
 5. Explain why each selected tool is useful for their agent
-6. Ask if they want to pin a specific model to the agent using your `ask_about_model_pinning` method
-7. Include the model in the final JSON if the user chooses to pin one
+6. Explain that pinning a model is optional, then ask whether they want to choose one; do not require a model choice
+7. Include the `model` field in the final JSON only if the user explicitly chooses to pin one; otherwise omit it so the agent uses the global model
 
 ## JSON Agent Schema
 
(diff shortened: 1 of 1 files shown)


code_puppy/agents/agent_creator_agent.py
```diff
@@ -84,29 +84,29 @@ def get_system_prompt(self) -> str:
...
+6. Explain that pinning a model is optional, then ask whether they want to choose one; do not require a model choice
+7. Include the `model` field in the final JSON only if the user explicitly chooses to pin one; otherwise omit it so the agent uses the global model
 
 ## JSON Agent Schema
 
 Here's the complete schema for JSON agent files:
 
 ```json
 {{
-  "id": "uuid"                       // REQUIRED: you can gen one on the command line or something"
```
AndrewTilson, line 96: I belive this field needs to be kept.

thomwebb: Good catch — I restored the `id` field in the schema example and kept the JSON formatting valid. Focused Agent Creator tests still pass (21 passed).

thomwebb: I checked the closed-source wm repo. `JSONAgent._validate_config()` requires only `name`, `description`, `system_prompt`, and `tools`; `id` is not validated or used by the loader. The wm repo tests also construct valid agents without `id`. I removed it from the example again so the documented schema matches runtime requirements. Focused tests pass (21 passed).

AndrewTilson (approved):

PR #685, opened 2026-07-28
PUP-549: Raise GPT-5.6 sub-agent depth cap from 1 to 2

AndrewTilson:
## Summary
Raises the GPT-5.6 sub-agent invocation depth cap from **1** to **2** and makes the cap tunable via a new config key `subagent_recursion_limit_gpt_5_6` (default `2`). Still GPT-5.6-specific. A GPT-5.6 immediate caller may now delegate one level deeper (`main -> level 1 -> level 2`); a third hop from a GPT-5.6 caller is still blocked with a clearer, depth-aware error message.

Jira: [PUP-549](https://jira.walmart.com/browse/PUP-549)

## Motivation
The prior fix landed a hard `depth == 1` cap to stop GPT-5.6 runaway-delegation. In practice that turned out to be too tight for legitimate two-hop workflows. Bumping the default cap to 2 restores useful delegation without re-opening the runaway-chain door, and adding a config knob (mirroring the existing generic `subagent_recursion_limit`) gives operators the escape hatch when they need one.

## Behavior boundary
Both limits below refer *only* to the new GPT-5.6-specific guard. The generic `subagent_recursion_limit` (default 4) fires first and can still block any of these rows on its own — it is unchanged.

| Immediate caller | Current depth | Attempted child depth | Blocked by GPT-5.6 guard? |
|---|---|---|---|
| main agent (any model) | 0 | 1 | no (no caller-model tracked at depth 0) |
| GPT-5.6 sub-agent | 1 | 2 | no *(newly permitted; used to be blocked)* |
| GPT-5.6 sub-agent | 2 | 3 | yes |
| non-GPT-5.6 sub-agent under a GPT-5.6 ancestor | any | any | no (guard reads immediate caller only) |

Immediate-caller-wins-over-ancestor semantics are preserved by design — broadening to a full-chain scan would be a policy change beyond the ticket's scope. See `code_puppy/tools/subagent_invocation.py::_gpt_5_6_recursion_blocked` docstring for the rationale.

## Changes
- `code_puppy/config.py`
  - New `DEFAULT_SUBAGENT_RECURSION_LIMIT_GPT_5_6 = 2` and `get_subagent_recursion_limit_gpt_5_6()`, mirroring the existing generic pattern (missing/invalid/negative values fall back to the default).
  - Registered `subagent_recursion_limit_gpt_5_6` in `get_config_keys()`.
- `code_puppy/command_line/set_menu_catalog.py`
  - New `/set` entry for `subagent_recursion_limit_gpt_5_6` under Behavior, with a description that flags the safety implication of raising it.
- `code_puppy/tools/subagent_invocation.py`
  - Removed the module-level `GPT_5_6_SUBAGENT_DEPTH_LIMIT` constant; guard now reads the live config value via `get_subagent_recursion_limit_gpt_5_6()`.
  - Rewrote `_gpt_5_6_recursion_blocked()` in attempted-child-depth terms: short-circuits `False` for non-GPT-5.6 immediate callers, otherwise blocks when `get_subagent_depth() + 1 > get_subagent_recursion_limit_gpt_5_6()`.
  - Enforcement site passes the live `depth` and `limit` into the i18n message.
- `code_puppy/i18n/locales/en-US.json`
  - Reworded `subagent.gpt_5_6_recursion_blocked` to the action-first phrasing sol suggested: *"Cannot invoke '{agent}' at sub-agent depth {depth}: GPT-5.6 callers are limited to depth {limit}."* Other locales don't override the key; they fall back to en-US.
- `code_puppy/agents/_builder.py`
  - Replaced the constant `_GPT_5_6_INVOKE_AGENT_GUARD_TEXT` with a function `_build_gpt_5_6_invoke_agent_guard_text()` that reads the live cap at prompt-assembly time. Kills any drift between model-facing guidance and runtime enforcement.
- `tests/agents/test_builder_gpt_5_6_subagent_depth.py`
  - Depth-1 allowed / depth-2 blocked / immediate-caller-wins-over-ancestor boundary cases. Fixed a misleading comment that overstated the guard's authority.
- `tests/agents/test_builder_gpt_5_6_subagent_guard.py`
  - New parametrised test proving the guard text interpolates the live config value (`[2, 3, 5]`).
  - New "prompt and runtime agree" test that patches the config to `3`, asserts the prompt contains `3`, and asserts the runtime blocks at attempted depth 4 but not at attempted depth 3 — binding both sides to the same source of truth.
- `tests/tools/test_subagent_recursion_limit.py`
  - New default/invalid/dynamic-value coverage for `get_subagent_recursion_limit_gpt_5_6`.
  - `subagent_recursion_limit_gpt_5_6` added to the discoverable-config-keys assertion.
  - Updated the two-level-cap tool test to expect the reworded error string; new depth-1-allowed test that stops at `load_agent()` to prove the guard didn't fire.

## Deliberate non-goals
- No broadening of the rule to inspect the full ancestor chain — immediate-caller semantics preserved on purpose.
- No changes to the generic `get_subagent_recursion_limit` machinery or to non-GPT models.

## Tests
- Targeted (three files): **42 passed** (up from 31 with the added config + semantic-prompt coverage).
- `tests/tools` + `tests/agents` + `tests/command_line`: **4252 passed, 12 skipped, 0 failed**.
- `ruff check` on all touched files: passed.
```
uv run ruff check code_puppy/config.py code_puppy/tools/subagent_invocation.py \
  code_puppy/agents/_builder.py code_puppy/command_line/set_menu_catalog.py \
  tests/agents/test_builder_gpt_5_6_subagent_depth.py \
  tests/agents/test_builder_gpt_5_6_subagent_guard.py \
  tests/tools/test_subagent_recursion_limit.py
uv run pytest --no-cov \
  tests/agents/test_builder_gpt_5_6_subagent_depth.py \
  tests/agents/test_builder_gpt_5_6_subagent_guard.py \
  tests/tools/test_subagent_recursion_limit.py
uv run pytest --no-cov tests/tools tests/agents tests/command_line
```

## Reviewer checklist
- [ ] `_gpt_5_6_recursion_blocked` boundary math (`get_subagent_depth() + 1 > get_subagent_recursion_limit_gpt_5_6()`) matches your reading of the policy.
- [ ] The immediate-caller-wins-over-ancestor behavior is intentional and adequately tested.
- [ ] The reworded error string is clearer than the original.
- [ ] Prompt guidance and runtime enforcement stay in lockstep via the new "prompt and runtime agree" test.
- [ ] Description of the new `/set` key correctly warns about the safety implication.


thomwebb:
## Review: changes requested

The GPT-5.6 recursion-cap change itself looks coherent, but this PR also includes a large macOS Computer Use plugin with several safety/correctness issues that need fixing before merge.

### 1. Blocker: `computer_screenshot` can overwrite arbitrary user files
`code_puppy/plugins/computer_use/backend.py:574` and `tools.py:246`

The model-controlled `path` argument is expanded and passed directly to the native capture helper, which writes the PNG at that location. A call such as `computer_screenshot(app_name="Safari", path="~/.ssh/config")` can replace an existing user file once Computer Use is enabled. Atomic replacement still destroys the previous contents.

**Suggested fix:** remove the public `path` argument, or restrict resolved paths (including symlink traversal) to a plugin-owned `0700` cache/temp directory. Do not create arbitrary caller-provided parent directories. Add a rejection test for paths outside that directory.

### 2. Blocker: overlapping snapshots can act on a different app than the policy-approved state
`code_puppy/plugins/computer_use/backend.py:153-154` and `182-197`

`snapshot()` stores elements in the shared mutable `self._elements`, while `get_app_state()` copies that map later. Two overlapping `get_app_state()` calls for apps A and B can therefore produce a state labeled and policy-checked for A containing B's AX element objects. A later action validates/activates A but manipulates B. This bypasses the intended app/policy boundary.

**Suggested fix:** return the element map from `snapshot()` and build the `AppState` from that returned per-request value; do not use shared backend state for request-scoped elements. Please add a deterministic concurrent-snapshot regression test.

### 3. High: failed batches leave a stale revision reusable after a successful mutation
`code_puppy/plugins/computer_use/tools.py:293-304` and `328-329`

Batch actions run with `consume=False` and the revision is consumed only after every step succeeds. If an early click/navigation succeeds and a later step fails, the function returns without consuming the original revision. The caller can then reuse a screenshot/accessibility tree from before the successful mutation.

**Suggested fix:** consume/invalidate the revision immediately after the first successful mutating step, including failure paths; or conservatively consume it before starting the batch after validation. Add a test for a successful first mutation followed by a failed second step.

### Validation run
- Focused pytest suite: **84 passed**
- Focused Ruff check + format check: **passed**
- `git diff --check`: **passed**

Also, since the PR is described as a GPT-5.6 depth-cap change but includes ~3.7k lines across 49 files (including this new Computer Use feature and usage metrics), please consider splitting the unrelated features into separate PRs for safer review and rollback.

AndrewTilson:
Hey @thomwebb — thanks for the review pass! I want to make sure we're looking at the same thing though, because the substance here doesn't seem to line up with what's actually in this PR and I don't want to accidentally ignore real feedback if I'm the one missing something.

The three blocking items all cite files under `code_puppy/plugins/computer_use/` (`backend.py`, `tools.py`), and the meta-note describes "~3.7k lines across 49 files (including this new Computer Use feature and usage metrics)". As best I can tell, PR #685 doesn't touch any of that. The full diff on this branch is 9 files / +259 / -19, all scoped to the GPT-5.6 sub-agent depth cap:

```
+24 -7   code_puppy/agents/_builder.py
+13 -0   code_puppy/command_line/set_menu_catalog.py
+29 -0   code_puppy/config.py
+ 1 -1   code_puppy/i18n/locales/en-US.json
+24 -4   code_puppy/tools/subagent_invocation.py
+22 -3   tests/agents/test_builder_gpt_5_6_subagent_depth.py
+65 -1   tests/agents/test_builder_gpt_5_6_subagent_guard.py
+ 2 -0   tests/test_config.py
+79 -3   tests/tools/test_subagent_recursion_limit.py
```

No `computer_use/` files, no usage-metrics changes. Is it possible the review was aimed at a different PR, or generated against a local worktree that had additional in-flight work layered on top? Happy to address any real concerns about the depth-cap change itself once we're on the same page.

(Closed this one for now to do more validation on the internal fork before re-proposing — but wanted to reply here so the thread's not left hanging.)

thomwebb:
You are correct — my review was based on an incorrect comparison range. I compared the PR head against my stale local `upstream/main` (`a139fca`) instead of GitHub's actual PR base (`169f81e`), which pulled in already-merged Computer Use and usage-metrics commits. The three Computer Use findings and the scope note therefore do **not** apply to PR #685.

Sorry for the noise, and thanks for catching the mismatch. I've rechecked the PR metadata: its actual head-only changes are the four GPT-5.6 recursion-cap/config/test commits you listed. Please disregard my prior review comment.

thomwebb (approved):

Issue #607, opened 2026-07-12
Investigate preventing shell-enabled agents from terminating Code Puppy


mpfaffenberger:
## Problem

An agent with access to `agent_run_shell_command` can potentially terminate the Code Puppy host process (or a parent/process tree that takes Code Puppy down with it). That lets the agent accidentally or intentionally kill its own runtime, losing the active run and potentially leaving child processes or terminal state behind.

We should determine whether Code Puppy can place a **non-optional guard** around shell execution that protects its own process and required ancestors without breaking legitimate command management.

## Current architecture / likely interception point

- `code_puppy/tools/command_runner.py::run_shell_command()` invokes the `run_shell_command` callback before launching foreground or background commands.
- `code_puppy/plugins/destructive_command_guard/` already uses that hook for dangerous filesystem/git/container commands.
- The current destructive detector does not appear to cover process termination (`kill`, `pkill`, `taskkill`, `Stop-Process`, API calls from scripts, etc.).
- A command-string regex alone is not a complete boundary: an agent can write or invoke Python/PowerShell/C/native code, use aliases or indirection, target a process tree, or launch a delayed/background killer.

## Investigation goals

1. Map the process topology on Windows, Linux, and macOS:
   - Code Puppy host PID;
   - terminal/shell ancestors whose termination would also take down or corrupt the session;
   - shell-tool child process groups/jobs;
   - detached/background commands.
2. Define the protected set precisely. At minimum, protect the active Code Puppy process; determine whether selected ancestors and helper processes must also be protected.
3. Evaluate enforcement options and document their guarantees and bypasses:
   - mandatory `run_shell_command` policy guard for obvious direct attempts;
   - stripping/protecting PID information passed to children where practical;
   - OS-level isolation or reduced credentials/capabilities for shell children;
   - Windows process security descriptors/restricted tokens/job objects;
   - Linux/macOS separate UID, namespace/container, sandbox, seccomp, or equivalent mechanisms;
   - watchdog/supervisor recovery if prevention cannot be made portable;
   - behavior for MCP/plugin-provided shell-like tools, which may bypass the built-in runner.
4. Decide what can be guaranteed cross-platform versus what is only best-effort defense in depth.
5. Produce an implementation plan with a testable threat model before coding the guard.

## Requirements / expected behavior

- Direct attempts to terminate the protected Code Puppy process or protected process tree are blocked before execution with a clear explanation.
[...]

Issue #454, opened 2026-06-10
destructive_command_guard and force_push_guard are whole-file structural clones — extract a shared command-guard framework


mpfaffenberger:
## Theme

`plugins/destructive_command_guard/register_callbacks.py` (152 lines) and `plugins/force_push_guard/register_callbacks.py` (148 lines) are the **same plugin twice** with different strings and a different detector. jscpd flagged only an 8-line window (83-91 ↔ 79-87), but a full-file diff shows the entire structure is cloned:

| Component | destructive_command_guard | force_push_guard |
|---|---|---|
| `_is_interactive()` | :23-28 | :19-24 (byte-identical) |
| main callback | `destructive_command_guard_callback` :31-67 | `force_push_guard_callback` :27-63 (identical flow: disabled-check → detect → TTY? prompt : block) |
| `_prompt_user_approval()` | :70-117 | :66-113 (identical panel construction, only title/strings differ) |
| `_block_command()` | :120-145 | :116-141 (identical dict shape, only message text differs) |
| `register()` + import-time auto-register | :148-152 | :144-148 |

Both consult the same `get_disable_dangerous_command_guard()` kill switch, both build the same `Text` panel (warning line, pattern name, description, `$ command`, consequence line), both return the same `{"blocked": True, "reasoning": ..., "error_message": ...}` contract.

## Why it matters

This is the template every future guard plugin will be copied from (the third copy is already half-born: `plugins/shell_safety/` implements the same hook with its own structure). Any fix to the approval flow — e.g. the non-interactive detection, the feedback plumbing, or the blocked-dict contract — must be applied N times and will be missed in at least one copy. The two prompts have already drifted cosmetically (consequence sentences, reasoning prefixes), which is harmless today but proves the copies aren't maintained in lockstep.

## Proposed shared abstraction

Add `code_puppy/plugins/_guard_framework.py` (or `code_puppy/guards.py`) with one parametrized factory:

```python
@dataclass(frozen=True)
class GuardSpec:
    title: str                  # "Force Push Guard "
    detected_label: str         # "Force push detected: "
    consequence: str            # "Force pushing rewrites remote history..."
    block_advice: str           # "If you *really* need to ... in your terminal ..."
    detect: Callable[[str], Optional[Any]]   # returns match with .pattern_name/.description

def make_shell_guard(spec: GuardSpec) -> Callable:
    async def guard(context, command, cwd=None, timeout=60):
        if get_disable_dangerous_command_guard():
[...]

PR #689, opened 2026-07-30
fix/feat: Issue 454 and 607, self_termination_guardrail

collin-kierce:
This PR aims to address issues #454 and #607. 
It refactors existing shell-command guards into a shared framework and adds a mandatory, defense-in-depth detector for direct process-termination commands targeting the active Code Puppy process or selected ancestors.

This is not complete protection against same-user arbitrary code execution. Interpreter indirection, external shell-like tools, delayed/background execution, and OS-level isolation remain follow-up work.
 
### Changes 
 1. Extract shared guard callback plumbing
 
 Adds code_puppy/plugins/guard_framework.py, which centralizes the common callback flow used by:
 
 - destructive_command_guard
 - force_push_guard
 - the new self_termination_guardrail
 
The shared framework handles disabled-guard policy, interactive approval prompts, non-interactive blocking, and consistent blocked-command responses. The self-termination guard opts out of the normal dangerous-command disable setting.
 
 2. Add `self_termination_guardrail`
 
Adds a guard intended to block direct shell commands that could terminate Code Puppy or selected parent processes. At plugin initialization, the detector walks the current process’s parent chain and builds a protected identifier set from process names and PIDs, along with static Code Puppy and shell aliases. 

Before a shell command is run, the detector:
 
 - identifies the executable and arguments, including several common command wrappers;
 - checks whether the executable is a recognized process-termination utility, such as kill, pkill, killall, taskkill, Stop-Process, or spps;
 - checks whether an argument targets a protected identifier.

When both conditions match, the command is blocked immediately without a user-approval prompt or the normal dangerous-command guard bypass.


thomwebb:

PR #691, opened 2026-07-31
feat: cost_tracker plugin — /cost session spend estimates from billed…

v3nr1ck:
## What

New built-in plugin: `/cost` shows estimated dollar spend for the current
session, broken down per model, with `/cost reset` to zero the meter.

```
💰 Session cost (estimated): $5.21
Last run: $0.525 (opus)

  opus: $5.21 (in 110k, out 25k, cache-read 500k, cache-write 50k; 2 runs) (priced as anthropic/claude-opus-4-0)

Estimates use bundled models.dev list prices; actual billing may differ (tiers, discounts, free quotas).
```

## Why

The models.dev integration already ships per-million-token prices
(`cost.input` / `cost.output` / `cost.cache_read` in the bundled
`models_dev_api.json`), and `_runtime` already extracts real billed usage
from `result.usage()` into the `agent_run_end` metadata
(`usage_input_tokens` etc.) — but nothing multiplies the two together.
For a tool whose README opens with anger at IDE pricing, spend visibility
felt like a natural fit.

## How

- `agent_run_end` callback folds each run's billed usage into a
  thread-safe process-lifetime accumulator. Sub-agent runs fire their own
  `agent_run_end` with their own `usage()`, so summing every event matches
  actual API billing — no double counting.
- Pricing resolves `config name → underlying model id → bundled models.dev
  entry`, preferring the provider implied by the config `type` when one id
  exists under several providers; `round_robin` models price via their
  first member. **Never touches the network** — bundled JSON only,
  matching the privacy commitment.
- Two cache-accounting styles, mirroring the split `model_factory` already
  uses: Anthropic-style (`input_tokens` excludes cache traffic; buckets
  summed at their own rates) vs OpenAI-style (cached portion subtracted
  from full-rate input and billed at the cache-read rate).
- Reasoning tokens are deliberately not billed separately (providers
  include them in billed output — adding them would double-charge).
- Models with no pricing data still get token tracking; they're listed
  separately and excluded from the total rather than guessed at.
- Failed/cancelled runs still count whatever was billed before the
  failure. Runs with all-zero usage (cancelled pre-billing) are skipped.
- `record_run` is wrapped so cost tracking can never break an agent run.

## Testing

- 26 new tests in `tests/plugins/test_cost_tracker.py`: cost math for both
  accounting styles, cache-rate fallback, bundled-DB resolution
  (provider preference, round-robin, date-suffix stripping, unknown
  models), accumulation/reset/thread-state, run-end callback including the
  failed-run path, `/cost` dispatch + `reset` + `--help` + unknown
  subcommand, formatting bands, index determinism.
- Full `tests/plugins/` suite passes (one pre-existing, unrelated failure
  in `test_chatgpt_oauth_core.py` reproduces on clean main).
- `ruff check` / `ruff format` clean under the CI-pinned 0.15.x.

## Notes / open questions

- Everything is labeled an estimate: list prices only, no tiers/batch/free
  quotas. Happy to add a config toggle to disable the plugin entirely if
  you'd prefer it opt-in.
- A statusline segment (running `$` next to the context indicator) would be
  a natural follow-up; kept out of this PR to keep it reviewable.

originally committed an old version, sorry about that homie :)

if you want changes or enhancements, i would gladly work on them.

---

*This contribution was developed with AI assistance (Claude Fable 5 on Max effort); all tests and
lint were run against current main before submission.*

### DIFF
--- a/code_puppy/plugins/cost_tracker/__init__.py
+++ b/code_puppy/plugins/cost_tracker/__init__.py
@@ -1,0 +1,1 @@
+"""Cost Tracker plugin - estimates dollar spend per session from billed token usage."""
--- a/code_puppy/plugins/cost_tracker/pricing.py
+++ b/code_puppy/plugins/cost_tracker/pricing.py
@@ -1,0 +1,308 @@
+"""Offline pricing lookup for the cost_tracker plugin.
+
+Resolves a Code Puppy model-config name (e.g. ``gpt-5``, ``qwen1``) to
+per-million-token prices using the **bundled** ``models_dev_api.json``
+database that ships with Code Puppy for the ``/add_model`` offline
+fallback.
+
+Design constraints:
+
+* **Never touches the network.** The project has a hard privacy
+  commitment; pricing lookups read only the bundled JSON that is already
+  on disk. If a model isn't in the bundle, its cost is simply reported
+  as *unknown* (tokens are still tracked).
+* **Deterministic.** The same model name always resolves to the same
+  price. When one model id exists under several providers with
+  different prices, the provider inferred from the model config's
+  ``type`` wins; otherwise the alphabetically-first provider id is used
+  and the result is (like everything here) labeled an estimate.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import pathlib
+import re
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple
+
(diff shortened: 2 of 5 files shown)


mpfaffenberger:
Plugins like this don't need to be merged to main. If you're inside Walmart deploy it to the plugin market.

PR #692, opened 2026-07-31
fix(mcp): a config can't be its own untrusted project twin (CWD == $HOME)

breedx:
Run `code-puppy` from your home directory and every launch warns about a file it already loaded:

```
> hi
Project MCP config '/home/you/.code_puppy/mcp_servers.json' is not trusted yet;
its servers are NOT loaded. Review it, then run '/mcp trust' to accept
```

### The two "sources" are the same file

When `CWD == $HOME`, the project candidate and the user-level config resolve to one path:

| source | path |
|---|---|
| user-level (implicitly trusted) | `~/.code_puppy/mcp_servers.json` |
| "project" (trust-gated) | `<CWD>/.code_puppy/mcp_servers.json` |

`get_project_mcp_servers_file()` returns that path as a project candidate, so the gate flags the user-level config as an untrusted stranger — while `load_mcp_server_configs()` has *already* merged its servers in via the user-level source. Measured:

```
load_mcp_server_configs()  → ['some-server']   ← loaded
get_trust_status(...)      → untrusted         ← warns "NOT loaded"
```

### Why this is worse than cosmetic noise

The warning is **false in the direction that costs trust**: it tells users their MCP tools are unavailable while they are working fine. Someone who believes it goes debugging a non-existent problem; someone who learns to ignore it has been trained to ignore the trust gate — which is the one message here that should carry weight.

It also fires on *every* launch, because the remedy never gets applied: `/mcp trust` is ceremony over a decision the user already made by writing the file, so nothing lands in `trusted_mcp.json` and the warning returns forever.

Not an exotic setup, incidentally — `cd ~ && code-puppy` is enough. Anything that runs the agent with `CWD=$HOME` (containers, sandboxes, headless/autonomous runs) hits it on every start, and an unattended run has nobody available to type `/mcp trust` anyway.

### Fix

`get_project_mcp_servers_file()` returns `None` when the candidate **is** the user-level config. There is no second, less-trusted source to gate, so there is nothing to ask about.

Compared with `samefile()` rather than path equality, so a symlink or bind-mount pointing at the user-level config is recognized too. The check targets `config.MCP_SERVERS_FILE` and never `$HOME`, so it stays correct when `XDG_CONFIG_HOME` relocates the user-level config out of `~/.code_puppy`.

Deliberately one chokepoint: `get_trust_status()`, `load_project_mcp_server_configs()` and `/mcp trust` all already handle "no project file" correctly, so a single guard fixes the whole path without adding a second code path.

### The real gate is untouched

A genuinely separate project config is still withheld until trusted — asserted in both directions, so the test fails if this fix ever widens into a bypass:

```python
assert cp_config.load_mcp_server_configs() == {"user_only": "u"}   # gated
assert pc.trust_project_mcp() is True
assert "proj_only" in cp_config.load_mcp_server_configs()          # honored
```

### Verification

- 4 tests added; **3 fail without the change** (14 → 17 in `tests/mcp/test_project_config.py`).
- Branch is cut from current `main`, so the diff is exactly these two files.

Context: `project_config.py` arrived in #615; the fix is against `if candidate.is_file(): return candidate` in the discovery helper.

### DIFF
--- a/code_puppy/mcp_/project_config.py
+++ b/code_puppy/mcp_/project_config.py
@@ -25,6 +25,11 @@ Trust model (identical philosophy to ``code_puppy/plugins/trust.py``):
   ``allow``).
 * **Fail closed**: an unreadable store, malformed JSON, or an unhashable file
   all resolve to "not trusted" → the project config is silently ignored.
+* **A file can't be its own less-trusted twin**: the project candidate can
+  resolve to the very same file as ``config.MCP_SERVERS_FILE`` — the usual
+  way being ``<CWD> == $HOME`` under the default config location. That file
+  is already loaded as user-level (implicitly trusted), so there is nothing
+  to gate and :func:`get_project_mcp_servers_file` returns ``None``.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


mpfaffenberger (approved):

mpfaffenberger:
Wow, weird edge case. Nice catch.

PR #696, opened 2026-08-03
feat(model-switcher): theme-style split-panel /model picker plugin

12britz:
## Summary

Adds a theme-style split-panel `/model` picker plugin (mirrors the `/theme` picker's look and feel), replacing the core single-pane model picker.

- **Left pane:** paginated, filterable model list
- **Right pane:** live preview — provider, context window, description, credential status, pinned agents
- **Interactions:** ↑/↓ navigate, PgUp/PgDn page, type to filter, Enter to apply, Ctrl+E to edit credentials, Esc to cancel
- **Commands:** `/model` (picker), `/model <name>` (switch by name), `/model show` (show active)

image

## Bug fix included

The plugin's `/model` override was silently replaced by the core handler at first command dispatch: `core_commands` is imported lazily on dispatch and re-registered the core `/model`, clobbering the plugin (registered at `cli_runner` import time). The fix forces `import code_puppy.command_line.core_commands` before the plugin registers, so the override wins for the whole session.

## Tests

- `tests/plugins/test_model_switcher_plugin.py` (16 tests): command registration/override, by-name switching, filter, pagination, navigation, rendering, empty-catalog handling
- `ruff check` + `ruff format` clean
- Full `tests/command_line/` suite (1940) still passes


thomwebb:

PR #697, opened 2026-08-03
Fix the highlights

adi07das:
Currently, the users face an issue with the highlight which hides the content being highlighted due the highlight color and the rendered content font colors. 

Current:
Screenshot 2026-08-03 at 2 17 33 PM

Screenshot 2026-08-03 at 2 19 20 PM

After the change:

Screenshot 2026-08-03 at 2 17 47 PM

Co-author: aditya.das@walmart.com

### DIFF
--- a/code_puppy/config.py
+++ b/code_puppy/config.py
@@ -110,6 +110,27 @@ def get_pack_agents_enabled() -> bool:
     return str(cfg_val).strip().lower() in {"1", "true", "yes", "on"}
 
 
+def get_selection_friendly_transcript() -> bool:
+    """Return True if transcript markdown should skip the theme fg override.
+
+    Default True. When True, ``_render_agent_response`` prints Markdown
+    WITHOUT forcing the theme's soft body foreground on every cell, so the
+    terminal's own text-selection highlight (typically a pale color) retains
+    contrast against the default terminal foreground.
+
+    When False, restores the pre-fix behavior of applying the theme fg to
+    all rendered markdown. This works around terminals that ignore OSC 10
+    (default-fg overrides), at the cost of low-contrast selection on any
+    terminal whose selection color is close in luminance to the theme fg.
+
+    Toggle via the ``selection_friendly_transcript`` config key.
+    """
+    cfg_val = get_value("selection_friendly_transcript")
+    if cfg_val is None:
+        return True  # Prefer selection contrast by default.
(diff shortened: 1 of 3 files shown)


adi07das:
@mpfaffenberger - please help review this

adi07das:
@thomwebb  - Please help review this. I've got the approvals on this: https://jira.walmart.com/browse/GECOSC-1556

thomwebb (changes requested):
Nice, targeted fix — the toggle design is clean and both branches are actually covered by the two new tests (`93 passed` on `tests/messaging/test_rich_renderer.py`). `ruff check` on the touched files shows nothing new introduced by this diff either.

Two things I'd like addressed before merge, though:

## 1. This flips a deliberate bug fix's default, silently

`git log -S "on_prompt_text_color()"` on this file turns up commit `98802318` ("Make terminal rendering fully theme aware"), whose commit message is explicit:

> "Some terminals ignore OSC 10, making Rich resets leak the terminal profile's white into otherwise themed responses."

That was a real fix for a real complaint. This PR's new default (`selection_friendly_transcript = True`) reopens exactly that leak for anyone on an OSC-10-blind terminal — they'll get white text bleeding through the theme again, with no indication that a config flag exists to get the old behavior back.

Selection contrast vs. theme leak is a legitimate trade-off, but flipping the *default* in one direction, silently, without any terminal-capability detection, deserves a call-out in the PR description at minimum so @mpfaffenberger can weigh in on whether that's an acceptable trade for most users. Would also be worth confirming this was tested against the original OSC-10 scenario the earlier commit fixed, not just the selection-contrast case this PR targets.

## 2. New config key isn't wired into the `/set` menu

Every other boolean/enum getter in `config.py` (`enable_pack_agents`, `enable_universal_constructor`, `enable_streaming`, the mcp toggles, etc.) is registered in `code_puppy/command_line/set_menu_catalog.py` as a `Setting(...)` entry, which is how they become discoverable/toggleable via the `/set` TUI. `get_selection_friendly_transcript()` isn't registered there. Given how consistently every other config key follows this pattern, this looks like an oversight rather than an intentional exception — right now the only way to opt back into the legacy behavior is to know the key exists and hand-edit the config file directly.

Please add a `Setting` entry (probably in the `_FEATURES` category, right next to the other bool toggles) so this is actually discoverable.

---
Happy to re-review once these are addressed — the core idea here is solid, just want the default-flip to be an explicit, informed call rather than a quiet side effect.

PR #703, opened 2026-08-05
Keep model catalog loads out of settings TUI repaints

breedx:
## Summary
- snapshot model configuration once per `/model_settings` session
- reuse the snapshot for names, descriptions, capability checks, defaults, and choices
- keep model-picker render and keypress paths free of config callbacks
- let `model_supports_setting` accept an optional preloaded catalog

## Reproduction
The model picker called `ModelFactory.load_config()` 16 times per repaint: once in the list and 15 times while deriving capabilities for the details panel. For network-backed model catalogs, a 0.9s config load made each arrow/Enter/Esc repaint take 14.89s. The inner settings view remained responsive because it reused its supported-settings list.

## Verification
- regression tests assert repeated picker repaints perform zero additional catalog loads
- internal `TypeError` propagation is covered (no exception-driven retry)
- focused model-settings suites: 131 passed on current upstream main
- Ruff and formatting clean

### DIFF
--- a/code_puppy/command_line/model_settings_menu.py
+++ b/code_puppy/command_line/model_settings_menu.py
@@ -251,10 +251,23 @@ def _format_custom_pairs(pairs: Dict) -> str:
     return "; ".join(f"{k}={_format_custom_value(v)}" for k, v in pairs.items())
 
 
-def _load_all_model_names() -> List[str]:
+def _load_all_model_names(models_config: Optional[dict] = None) -> List[str]:
     """Load all available model names from config."""
-    models_config = ModelFactory.load_config()
+    if models_config is None:
+        models_config = ModelFactory.load_config()
     return list(models_config.keys())
+
+
+def _supports_setting(
+    model_name: str, setting: str, models_config: Optional[dict] = None
+) -> bool:
+    """Check capability against one preloaded model-catalog snapshot."""
+    try:
+        return model_supports_setting(model_name, setting, models_config=models_config)
+    except TypeError:
+        # Preserve compatibility with older plugin overrides/test doubles that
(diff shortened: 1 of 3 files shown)


thomwebb (approved):
## Verification

The bug survived the 247 commits since this branched — nothing on main memoizes or threads the catalog, and the three `load_config()` sites are still there (`model_settings_menu.py:256, 347, 542`). It got worse: I counted loads per menu session with a patched `load_config` —

| | construct | 3 repaints | after selecting a model |
|---|---|---|---|
| main | 3 | 63 (20/repaint) | 101 |
| this branch | 3 | 3 | 3 |

20 per repaint against the 16 in the repro when this opened — `e787168d` (xhigh adaptive effort) added capability checks to the loop.

Merged current main in locally: zero conflicts, 261 passing across the new latency suite, menu coverage, and config/settings. The 2 `TestAPIKeys` failures fail identically on plain main (env vars on my box).

Snapshot staleness: not possible mid-session — the menu is modal with no add/import/save path that could mutate the catalog. Snapshot at construct, pure render and key paths after. Correct shape, merging as-is.

## Follow-ups (non-blocking)

- `__init__`'s `inspect.signature` check on `_load_all_model_names` can't fail on code this branch ships — the function is defined in this file, in this PR. Call it directly in a cleanup pass.
- `_supports_setting`'s reflection crosses a real boundary (`config.model_supports_setting` is monkeypatched elsewhere), but it runs per capability check inside the render loop this PR exists to make cheap. Hoist to an import-time bool, or drop the branch if nothing actually swaps the signature. The TypeError-propagation test suggests it's deliberate — naming that consumer in a comment would settle it.

Happy to take either follow-up myself.

PR #705, opened 2026-08-05
fix: preserve termflow table alignment columns

mattnico:
Fixes a bug where column alignment was being applied to the wrong columns by Termflow.


code_puppy/termflow_patches.py
```diff
@@ -0,0 +1,133 @@
...
+        if original is None:
+            raise AttributeError("termflow Parser._parse_table_alignments not found")
+        if getattr(original, _PATCH_MARKER, False):
+            return True
+
+        def _parse_table_alignments(self: Any, content: str) -> None:
+            self.table_alignments.clear()
+            for raw_cell in content.split("|"):
+                cell = raw_cell.strip()
+                if not cell:
```
thomwebb, line 106: Termflow accepts empty cells in a separator row (`_is_table_separator` skips empty cells while validating), so this `continue` still drops valid interior columns and can shift alignment metadata. For example, `|---||---:|` currently produces `("none", "right")`, but it must preserve the empty middle column as `("none", "none", "right")`. Please distinguish optional edge cells from empty interior cells and append `"none"` for the latter; add a regression test for this case.

PR #712, opened 2026-08-07
fix: isolate retry hook callback diagnostics

WSxDemise:
[CodePuppy Agent] - ## What

Add a synchronous `streaming_retry_display_suffix` callback seam to the streaming retry warning path.

## Why

Retry warnings can benefit from deployment-specific diagnostics without putting provider or enterprise details into the core harness.

## How

- Add a provider-neutral retry-warning suffix callback phase.
- Preserve generic warning output when no callback is registered.
- Collect additive suffixes in registration order.
- Isolate callback exceptions, invalid returns, and oversized values so plugins cannot interrupt retry recovery.
- Add `walk_exception_tree()` for shared traversal of wrapped causes, contexts, and `ExceptionGroup` members.
- Support callable objects safely in callback diagnostics rather than assuming callbacks always expose `__name__`.

## Testing

- `/pr-me` local gate: changed-file Ruff and formatter checks passed.
- `/pr-me` affected tests: 209 passed.
- `/pr-me` independent pre-PR review: clean.
- Full validation previously passed: `12,692 passed, 81 skipped, 1 xpassed`.
- i18n audit completed.

## Scope

Provider-neutral OSS infrastructure only. This PR contains no Walmart, Element, gateway response-body, or private diagnostic behavior.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -23,7 +23,7 @@ import signal
 import threading
 import uuid
 from contextlib import AsyncExitStack
-from typing import Any, Callable, Iterator, List, Optional, Sequence, Type, Union
+from typing import Any, Callable, List, Optional, Sequence, Type, Union
 
 import httpcore
 import httpx
@@ -59,7 +59,7 @@ try:  # pragma: no cover - pydantic-ai version dependent
 except ImportError:
     ModelAPIError = None  # type: ignore[misc,assignment]
 
-# Python 3.11+ builtin; graceful fallback for 3.10
+# Python 3.11+ builtin; graceful fallback for 3.10.
 try:
     from builtins import BaseExceptionGroup  # type: ignore[attr-defined]
 except ImportError:  # pragma: no cover - 3.10 only
@@ -81,6 +81,7 @@ from code_puppy.agents._run_signals import (
     sigint_should_cancel,
 )
 from code_puppy.agents.event_stream_handler import event_stream_handler
+from code_puppy.agents.exception_tree import walk_exception_tree
 from code_puppy.callbacks import (
     on_agent_exception,
     on_agent_run_cancel,
@@ -89,6 +90,7 @@ from code_puppy.callbacks import (
     on_agent_run_result,
     on_agent_run_start,
     on_should_skip_fallback_render,
+    on_streaming_retry_display_suffix,
     on_user_prompt_submit,
 )
 from code_puppy.config import (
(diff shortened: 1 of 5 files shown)


thomwebb:
@WSxDemise is this still relevant? if so can you resolve the merge conflicts? also my agent found the following:


Issues
──────

Blocking-ish: 208 commits behind main, and it shows

Both test files this PR modifies (tests/agents/test_streaming_retry_e2e.py, tests/test_callbacks_extended.py) no longer exist on current main — the retry tests were restructured into tests/agents/test_runtime_streaming_retry_classifier.py, and _runtime.py/callbacks.py have drifted substantially (~370/~240 diff lines). The "12,692 passed" claim is against a stale base; this will conflict hard on rebase and the new tests need re-homing. The good news: I checked, and _walk_cause_chain/_group_members still exist on main, so the refactor itself remains applicable.

Substantive: two seam-safety gaps

• Async-callback trap: if a plugin registers async def for this phase (a natural mistake given most phases here are async), callback(...) returns a coroutine → silently dropped as "invalid" plus coroutine was never awaited GC noise every retry. The sibling filter has the same trap, so it's consistent — but this dispatcher's docstring explicitly contracts failure behavior, so one inspect.iscoroutine(result) check with a targeted warning (and .close() on it) would close the loop. Or at least say "must be a plain def" in the docstring.
• Latency isn't isolated: exceptions/returns/size are contained, but time isn't. The most probable real consumer of this hook is a diagnostics plugin phoning an internal portal — i.e., blocking network I/O — executed synchronously before every retry sleep. "Must be quick and non-blocking" is a docstring plea, not a guarantee. A full watchdog is probably YAGNI for now, but flagging it because it will be the first way this gets misused.

Nits

• "".join(suffixes) relies on each plugin supplying its own leading space (tests use " [one]"). That convention is undocumented — one docstring sentence prevents (attempt 1, streak 1/3)FOO.
• _callback_name fixes the callable-object crash for the new phase, but on_thinking_display_filter and _chain_value_callbacks still use bare callback.__name__ inside except-handlers — a raising callable object there makes the logger itself throw and escape the dispatcher. That's a pre-existing bug this PR's helper would fix in 3 lines. Cheap DRY follow-up.
• walk_exception_tree currently has exactly one OSS consumer; the docstring's "underpins private provider adapters" tells us the second consumer lives in a fork. The extraction stands on its own merits (one-liner predicate), but it's YAGNI-adjacent until then — and maintainers should note the OSS repo inherits maintenance of a seam whose primary consumer won't live here.

PR #713, opened 2026-08-07
PUP-582: Add namespace_skill_search plugin: model-agnostic namespace + on-demand skill search

thomwebb:
**Jira:** [PUP-582](https://jira.walmart.com/browse/PUP-582)

## What

Adds a new builtin plugin, `namespace_skill_search`, that replaces the flat per-skill list injected into the system prompt with a compact namespace directory plus an on-demand `browse_skill_namespace` tool.

Also updates `docs/AGENT_SKILLS.md` in two places:
- A new "Skill Namespaces (Large Catalogs)" section documenting the plugin's behavior, the `browse_skill_namespace` tool's three modes, namespace sizing, and how to opt out.
- A small addition to the existing "Creating Your Own Skills" authoring guide: the `tags` field description and a callout now explain that a skill's **first** tag determines its namespace, so new skill authors know tag order matters instead of discovering it after shipping a skill that lands in a crowded or unhelpful namespace.

## Why

Frontier labs converge on one shape for large tool/skill catalogs: **one grouping layer + on-demand search**, instead of a flat list of every item injected upfront.

- Anthropic's shipped Tool Search Tool: flat on-demand search, `defer_loading: true`, ~85% token reduction, measurable accuracy gains on internal MCP evals (Anthropic engineering blog: "advanced-tool-use").
- OpenAI's shipped `tool_search`: same on-demand mechanic, plus an explicit **namespace** grouping layer above it, because "our models have primarily been trained to search those surfaces" (OpenAI's tool-use guide).
- Neither ships a deep multi-level tree — one grouping layer is the norm in production; true hierarchical routing is mostly academic (e.g. MCP-Zero, arXiv:2506.01056).

This PR copies OpenAI's namespace+search shape, but OpenAI's mechanism depends on a model-specific `defer_loading` API flag. This plugin gets the same effect through Code Puppy's existing model-agnostic plugin hooks, so it behaves identically on Claude, GPT, Gemini, and every other model Code Puppy supports — no provider-specific request fields required.

## How it works

1. **Namespace derivation** — a skill's namespace is its first `tags:` entry (untagged skills land in `General`). No new frontmatter field, no migration of existing `SKILL.md` files.
2. **Compact directory injection** via the `load_prompt` hook — one line per namespace instead of one line per skill. `load_prompt` fragments are simply newline-joined across plugins, so this is safely additive.
3. **Drill-down tool**, `browse_skill_namespace`: no-args directory listing, `namespace=` to list skills in one namespace, `query=` to keyword-search across all namespaces.
4. **Flat list turned off** on first load via the existing `set_frontmatter_in_system_prompt(False)` config API, so the built-in flat list and the new namespace directory don't both render. This is a one-time flip on a dedicated migration marker — a user who re-enables it via `/skills frontmatter on` afterward has that choice respected on every subsequent run, since the plugin only checks the migration marker (not the current frontmatter value) to decide whether to act.
5. **Docs**: authoring guidance now flags that tag order matters (see "What" above), so this isn't just a runtime behavior change discoverable only by reading source.

## What this deliberately avoids

- **No `get_model_system_prompt` callback.** That phase is last-write-wins across plugins (confirmed in `model_utils.prepare_prompt_for_model` — augmenter results are threaded sequentially and the last one processed wins on any shared key). A second plugin registering that hook could silently clobber this plugin's contribution, or vice versa. `load_prompt` is additive by construction, so that's what this plugin uses instead.
- No vector DB / embeddings — namespace + substring search is enough at this scale.
- No deep multi-level tree (namespace -> subnamespace -> skill) — no frontier lab ships that in production.
- No new tool duplicating `list_or_search_skills` — that tool is untouched and still works; `browse_skill_namespace` is additive.

## Testing

- 19 new unit tests in `tests/plugins/test_namespace_skill_search_plugin.py`, all passing.
- Verified no regressions against `test_plugins_init.py`, `test_plugin_meta.py`, and `test_builtin_plugin_lock.py` (118 tests, all passing).
- `ruff check` and `ruff format` clean.
- Manually verified against a real installed skill set (21 skills -> 14 namespaces) to confirm the grouping/rendering logic behaves correctly on real data, not just mocks.
- Docs-only follow-up commit (authoring guide tag-order callout) has no test coverage of its own — it's a text change with no executable behavior.

### DIFF
--- a/code_puppy/plugins/namespace_skill_search/README.md
+++ b/code_puppy/plugins/namespace_skill_search/README.md
@@ -13,6 +13,14 @@ every item injected into the system prompt upfront.
   `defer_loading: true`, results injected at end of context to preserve
   prompt caching. ~85% token reduction, measurable accuracy gains on
   internal MCP evals ([Anthropic engineering blog, Nov 2025](https://www.anthropic.com/engineering/advanced-tool-use)).
+  Note: our `load_prompt` fragment lands earlier in the assembled prompt
+  (layer 2 of `base_agent.py::get_full_system_prompt`) than the built-in
+  flat skill list it replaces did (layer 6, dead last). We prioritized
+  additive-safety (`load_prompt` can't collide with another plugin's
+  fragment; `get_model_system_prompt` can) over exactly matching
+  Anthropic's end-of-context caching position. If cache-hit rate on the
+  namespace block ever matters in practice, moving it later is a
+  follow-up, not a redesign.
 - OpenAI's shipped `tool_search`: same on-demand mechanic, PLUS an
   explicit **namespace** grouping layer above it
   (`{"type": "namespace", "name": "crm", "tools": [...]}`), because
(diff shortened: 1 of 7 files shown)


thomwebb:
## Adversarial self-review pass

Ran this through an independent adversarial code review before asking for a human look. First pass surfaced 3 blockers + 6 warnings, all real bugs (not style nits):

**Blockers found & fixed:**
1. `register_callbacks.py` was flipping `frontmatter_in_system_prompt` off as an *import-time* side effect — irreversible, untestable, and inconsistent with this repo's own `startup`-callback convention (see `theme._apply_default_theme_on_first_run`). Moved to a `startup` callback with a dedicated migration marker.
2. Namespace grouping was case-*sensitive* while the drill-down lookup was case-*insensitive* — skills tagged `finance`/`Finance`/`FINANCE` would silently fragment into separate namespace entries. Fixed: grouping now canonicalizes case at build time.
3. Blank/whitespace-only `query=""` silently returned **zero** results instead of "no filter" in both browse modes (classic `any([])  == False` trap). Fixed with a shared `_query_terms()` helper.

**Warnings found & fixed:** blocking filesystem I/O in the async tool wasn't getting thread-offloaded by pydantic-ai (verified against the actual installed package — async tools don't get the same `anyio.to_thread` treatment sync ones do), no graceful handling if the skill catalog read throws, no visibility into duplicate skill names across namespaces, and the README's caching-position claim didn't match where `load_prompt` actually lands in the prompt-assembly order.

Second pass confirmed every finding is actually fixed (not just reworded) — full review trail available on request. 11 new tests added (30 total, all passing), no regressions in `test_plugins_init`/`test_plugin_meta`/`test_builtin_plugin_lock`, `ruff check`/`format` clean.

WSxDemise (approved):
Still think @mpfaffenberger should give it the sniff test, if possible.

mpfaffenberger:
It's a plugin so I can just merge

PR #735, opened 2026-08-09
Decouple chatgpt_oauth plugin (#728)

### LINKED ISSUE #728: Decouple chatgpt_oauth plugin: replace core imports with register_model_type / register_model_providers + usage-status hook
## Context

## Goal

- `register_model_type` (already registered)
- `register_model_providers` (exists)
- `register_agent_tools`, `register_skills`, `custom_command` (already registered)

|---|---|
| `command_line/core_commands.py` | call-time `handle_tutorial_command` -> `.oauth_flow.run_oauth_flow` |
| `messaging/spinner/__init__.py` | call-time `_codex_usage_suffix` -> `.usage.get_usage_status` |

## Notes

- The spinner reads `get_usage_status()` (a cached Codex quota read). No
  `usage_status` hook exists yet; propose a small `load_usage_status` /
  `get_usage_status` callback so spinner.py queries the hook instead of the
  plugin module. (Keep it non-blocking / cache-only semantics.)
- OAuth flow (`run_oauth_flow`) should be reachable via the `custom_command`
  hook the plugin already registers; core's tutorial command should dispatch
  to that hook rather than import `.oauth_flow`.

## How to verify

- `grep -rn "code_puppy.plugins.chatgpt_oauth" code_puppy --include='*.py'` outside
  `code_puppy/plugins/chatgpt_oauth/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
Closes #728. Lazy oauth_flow imports + new usage_status callback hook.

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -23,6 +23,7 @@ PhaseType = Literal[
     "agent_reload",
     "custom_command",
     "custom_command_help",
+    "usage_status",
     "file_permission",
     "pre_tool_call",
     "post_tool_call",
@@ -86,6 +87,7 @@ _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "agent_reload": [],
     "custom_command": [],
     "custom_command_help": [],
+    "usage_status": [],
     "file_permission": [],
     "pre_tool_call": [],
     "post_tool_call": [],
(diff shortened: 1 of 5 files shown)


mpfaffenberger:

I verified both core imports are removed: `core_commands.py` no longer imports `chatgpt_oauth.oauth_flow`, and the spinner no longer lazily imports `chatgpt_oauth.usage`. The plugin owns `usage_status`; absent/disabled plugins produce `""""`, callback exceptions are isolated, registration is deduped by the callback registry, and the tutorial reuses the plugin's existing `/chatgpt-auth` handler (which also performs the prior model switch).

The spinner tests meaningfully cover provider, non-Codex, and no-provider behavior. The seam is sync-only by intent (cached rendering hot path), which is appropriate/YAGNI.

Integration note: this PR adds `PhaseType`/`_callbacks` entries in `callbacks.py` and will textually overlap #738, #741, #742, and #744. Resolve by retaining the union of phase names and registry keys; #743 also edits the same module near the registry.

PR #737, opened 2026-08-10
Decouple file_permission_handler plugin (#729)

### LINKED ISSUE #729: Decouple file_permission_handler plugin: core should consume the file_permission hook, not imports
## Context

## Goal

- `file_permission` (already registered!)

|---|---|
| `tools/common.py` | call-time `_get_user_approval_impl`, `_get_user_approval_async_impl` -> `.register_callbacks` |
| `tools/file_modifications.py` | call-time `_create_rejection_response`, `_emit_diff_message` -> `.register_callbacks` |

## Notes

- This is the cleanest one: the approval flow *is* a hook (`file_permission`).
  The plugin already registers it. Core should invoke approval through the
  hook mechanism and only fall back to its own logic if no handler answered.
- The exact symbols currently imported from `.register_callbacks` should move
  into a small public API the plugin exposes via the hook path, or core calls
  `file_permission` handlers directly.

## How to verify

- `grep -rn "code_puppy.plugins.file_permission_handler" code_puppy --include='*.py'` outside
  `code_puppy/plugins/file_permission_handler/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
Fixes #729.

Core consumes the existing `file_permission` hook for approval decisions and a registered file-permission state provider for feedback/diff UX state instead of importing the plugin.

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -159,6 +159,11 @@ def clear_loading_context() -> None:
     _current_loading_plugin = None
 
 
+def get_loading_context() -> Optional[str]:
+    """Return the plugin currently being loaded, if any."""
+    return _current_loading_plugin
+
+
 def get_callback_owner(func: CallbackFunc) -> Optional[str]:
     """Return the plugin name that registered *func*, or ``None``."""
     return _callback_owners.get(func)
@@ -225,6 +230,11 @@ def clear_callbacks(phase: Optional[PhaseType] = None) -> None:
             logger.debug(f"Cleared async callbacks for phase '{phase}'")
 
 
+def is_callback_owner_enabled(owner: Optional[str]) -> bool:
+    """Return whether callbacks and providers owned by *owner* are enabled."""
+    return owner is None or owner not in _get_disabled_plugins()
+
+
 def get_callbacks(
     phase: PhaseType, *, include_disabled: bool = False
 ) -> List[CallbackFunc]:
@@ -237,11 +247,11 @@ def get_callbacks(
     if include_disabled:
         return all_cbs
 
-    disabled = _get_disabled_plugins()
-    if not disabled:
-        return all_cbs
-
-    return [cb for cb in all_cbs if _callback_owners.get(cb) not in disabled]
+    return [
+        callback
+        for callback in all_cbs
+        if is_callback_owner_enabled(_callback_owners.get(callback))
+    ]
 
 
 def count_callbacks(phase: Optional[PhaseType] = None) -> int:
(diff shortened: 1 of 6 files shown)


mpfaffenberger:

The four lazy core imports of `file_permission_handler.register_callbacks` are removed and no-provider defaults preserve the old ImportError fallback. The focused provider tests cover delegation, replacement, and fallback.

Blocking disabled-plugin/load-lifecycle issue: `register_file_permission_state_provider(...)` installs five process-global function references outside the callback registry (`code_puppy/plugins/file_permission_handler/register_callbacks.py:563-573`; `code_puppy/tools/file_permission_state.py:32-61`). Consequently callback ownership filtering does not apply: after this plugin is loaded and then disabled, core still calls its state provider. There is also no unregister/reset API for unload/reload, and re-registration can mix lifecycle expectations even though it atomically replaces the five references.

Recommended fix: make this provider ownership-aware and consult `get_disabled_plugins`, add an unregister token/provider object, or route these accessors through dedicated callbacks so existing disabled filtering/idempotency applies. Add an integration test that loads/registers the provider, disables `file_permission_handler`, and verifies core falls back to `False`/`None`/no-op. Sync semantics are otherwise appropriate because these are thread-local accessors. This PR does not edit `callbacks.py`, so it avoids the phase-key conflicts among #735/#738/#741/#742/#744.

mpfaffenberger:
Addressed reviewer clone #2's lifecycle blocker in ff41a21f.

The state provider now:
- captures the plugin loader ownership context automatically;
- uses the same `is_callback_owner_enabled` predicate as callback filtering, so loaded-then-disabled plugins immediately fall back to no-provider behavior;
- returns an identity registration token with explicit unregister support;
- protects reloads from stale unload tokens removing the replacement provider.

Added tests for disabled ownership, explicit unregister, reload/re-registration, loading-context ownership, and no-plugin fallback.

Validation:
- Ruff format + format check: passed
- Ruff check --fix + check: passed
- git diff --check: passed
- targeted permission/callback suite: 225 passed
- full tests/plugins suite: 1812 passed (1 existing warning)

No core imports of `file_permission_handler` remain.

mpfaffenberger:
Independent reviewer clone #2 re-review of `ff41a21f`: **clean / approve (COMMENT review)**.

The lifecycle blocker is resolved. Registration captures the loader owner, `_get_active_provider()` uses the same enabled-owner predicate as callback filtering, and disabled owners fall back without touching stale plugin state. Registration returns an identity token; explicit unregister restores fallback; replacement on reload is atomic; stale tokens cannot unregister the replacement. Tests cover all of those cases plus loading-context capture and absence.

No core `file_permission_handler` imports remain. The provider accessors remain correctly synchronous/thread-local.

I independently ran Ruff check/format on all 6 changed Python files (pass) and `tests/tools/test_file_permission_state.py`: **8 passed**. Claimed broader suites were not rerun in full.

Overlap: this adds generic loading-context/owner helpers in `callbacks.py`, while #735/#738/#741/#742/#744 add phase keys there. Preserve these helpers plus the union of phase keys during sequential rebases; the changes are semantically compatible.

PR #738, opened 2026-08-10
Decouple claude_code_oauth plugin (#727)

### LINKED ISSUE #727: Decouple claude_code_oauth plugin: replace core imports with register_model_type / register_model_providers hooks
## Context

## Goal

- `register_model_type` (already registered)
- `register_model_providers` (exists in callbacks.py)
- `custom_command`, `custom_command_help` (already registered)
- `agent_run_start` / `agent_run_end` / `prepare_model_prompt` (already registered)

|---|---|
| `claude_cache_client.py` | call-time `_check_stored_token_expiry`, `_refresh_claude_oauth_token` -> `.utils` |
| `model_factory.py` | call-time `load_config` / `get_model` -> `.utils` |
| `command_line/core_commands.py` | call-time `handle_tutorial_command` -> `.register_callbacks` |

## Notes

- The token-refresh / model-construction helpers in `.utils` are the real
  coupling. `register_model_type` already self-registers the model type; the
  remaining `.utils` imports should move behind a provider registry the
  plugin owns, consumed via `register_model_providers`.

## How to verify

- `grep -rn "code_puppy.plugins.claude_code_oauth" code_puppy --include='*.py'` outside
  `code_puppy/plugins/claude_code_oauth/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
Fixes #727.

Removes core-to-plugin imports by routing token expiry checks, token refresh, filtered model loading, and tutorial authentication through callback hooks registered by the claude_code_oauth plugin.

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -332,6 +332,10 @@ def _trigger_callbacks_sync(
                     logger.warning(
                         f"Async callback {callback.__name__} called from async context in sync trigger"
                     )
+                    # The sync dispatcher cannot await while its event loop is
+                    # already running. Explicitly close the coroutine so this
+                    # safe fallback never leaks an unawaited-coroutine warning.
+                    result.close()
                     results.append(None)
                     continue
                 except RuntimeError:
@@ -1210,6 +1214,11 @@ def on_check_claude_oauth_token_expiry() -> List[Any]:
     return _trigger_callbacks_sync("check_claude_oauth_token_expiry")
 
 
+async def on_check_claude_oauth_token_expiry_async() -> List[Any]:
+    """Async variant for consumers already running inside an event loop."""
+    return await _trigger_callbacks("check_claude_oauth_token_expiry")
+
+
 def on_refresh_claude_oauth_token() -> List[Any]:
     """Ask the claude_code_oauth plugin to force a refresh-token exchange.
 
@@ -1218,6 +1227,11 @@ def on_refresh_claude_oauth_token() -> List[Any]:
         registered callbacks; empty when the plugin is not loaded.
     """
     return _trigger_callbacks_sync("refresh_claude_oauth_token")
+
+
+async def on_refresh_claude_oauth_token_async() -> List[Any]:
+    """Async variant for consumers already running inside an event loop."""
+    return await _trigger_callbacks("refresh_claude_oauth_token")
 
 
 def on_load_claude_oauth_models() -> List[Any]:
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -49,6 +49,10 @@ PhaseType = Literal[
     "register_browser_types",
     "register_model_providers",
     "register_completion_provider",
+    "check_claude_oauth_token_expiry",
+    "refresh_claude_oauth_token",
+    "load_claude_oauth_models",
+    "claude_oauth_authenticate",
     "message_history_processor_start",
     "message_history_processor_end",
     "on_message",
@@ -129,6 +133,10 @@ _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "register_browser_types": [],
     "register_model_providers": [],
     "register_completion_provider": [],
+    "check_claude_oauth_token_expiry": [],
+    "refresh_claude_oauth_token": [],
+    "load_claude_oauth_models": [],
+    "claude_oauth_authenticate": [],
     "message_history_processor_start": [],
     "message_history_processor_end": [],
     "on_message": [],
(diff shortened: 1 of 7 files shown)


mpfaffenberger:

The four intended core→`claude_code_oauth` imports are removed, including lazy imports. No-plugin/disabled behavior is mostly safe because `get_callbacks` filters callback owners and wrappers return empty lists.

Blocking seam/API findings:
1. `code_puppy/model_factory.py:585-587` bypasses the public `on_load_claude_oauth_models()` wrapper and directly invokes `get_callbacks(...)[0]()`. That loses normal callback error isolation and sync/async handling, hard-codes first-provider/load-order semantics in a consumer, and leaves the newly added wrapper unused. Call the wrapper and select the first valid dict result (or define a singular provider API explicitly).
2. This PR changes four runtime seams but adds no tests. Please cover absent/disabled plugin fallback, token-expiry and refresh result semantics, tutorial auth success/failure, filtered-model provider failure, repeated registration, and sync/async behavior (or explicitly reject async providers).

Also, `on_claude_oauth_authenticate()` returns `[None]` on successful `_hook_authenticate`, so `if on_claude_oauth_authenticate()` works only because the container is non-empty—not because auth reported success. Prefer a boolean provider result and test it; otherwise a swallowed callback exception and no provider are indistinguishable.

Integration: `callbacks.py` PhaseType/_callbacks edits conflict with #735/#741/#742/#744; retain the union when rebasing. #743 edits the same file near the registry too.

mpfaffenberger:
Addressed reviewer clone #2 findings in c26dba03:

- ModelFactory now consumes `on_load_claude_oauth_models()` exclusively, preserving callback error isolation and sync/async handling; it selects the first valid dict result and falls back to plain JSON for absent, disabled, or failed providers.
- Authentication now reports a real boolean: `_perform_authentication()` and `_hook_authenticate()` return success/failure, and tutorial model switching requires an explicit `True` result (so `[None]` is not success).
- Token expiry and refresh consumers now select validated results rather than assuming callback index 0.
- Added a focused four-seam regression matrix covering no provider, provider exception, disabled plugin owner, and valid provider paths, plus duplicate registration and sync/async dispatcher behavior.

Validation:
- Ruff format + format check: passed on every changed Python file
- Ruff check --fix + check: passed on every changed Python file
- git diff --check: passed
- Targeted OAuth/model-factory/core tests: 286 passed, 2 skipped
- Full tests/plugins suite: 1,842 passed (2 pre-existing warnings)

mpfaffenberger:
Independent reviewer clone #2 re-review of `c26dba03`: **remaining changes needed**.

Resolved: ModelFactory now uses `on_load_claude_oauth_models()` and selects a validated dict; dispatcher exceptions fall back safely; token results are type-validated; authentication returns an explicit bool and model switching requires `result is True`; substantial absent/disabled/error/valid seam tests were added.

One blocker remains: the claimed async-provider safety is not valid in the actual async-client context. All four public wrappers call `_trigger_callbacks_sync`. When an async provider is invoked while an event loop is already running, that dispatcher appends `None` and leaves the coroutine unawaited. I reproduced on this head:

```
async def provider(): return True
# register for check_claude_oauth_token_expiry, then call inside asyncio.run(...)
result == [None]
RuntimeWarning: coroutine 'provider' was never awaited
```

This matters because `ClaudeCacheAsyncClient.send()` calls `_should_refresh_token()` / `_check_stored_token_expiry()` and `_refresh_claude_oauth_token()` while its event loop is running. The new `test_dispatchers_support_sync_and_async_providers` is a synchronous test, so `_trigger_callbacks_sync` uses `asyncio.run()` and misses the production case.

Recommended fix: either make these provider contracts explicitly sync-only and remove the async-support claim/test, closing unawaited coroutine objects safely, or add async dispatchers and await them from async call sites. Add a test invoking an async provider from inside a running loop with warnings treated as errors.

I independently ran Ruff check/format on all 6 changed files (pass) and the new seam suite (**30 passed**), plus the reproduction above. `callbacks.py` still has the known PhaseType/_callbacks overlap with #735/#741/#742/#744; preserve all keys on rebase.

mpfaffenberger:
Addressed the remaining async-provider blocker in cd9f7fc2.

Changes:
- Added async expiry and refresh dispatchers backed by the normal async callback executor (`_trigger_callbacks`), preserving error isolation and disabled/no-provider filtering.
- `ClaudeCacheAsyncClient.send()` now awaits both the expiry provider and token-refresh provider in its running event loop.
- Sync consumers retain the sync dispatchers. If a sync dispatcher is called from a running loop with an async provider, it explicitly closes the coroutine before returning the safe `[None]` fallback, preventing an unawaited-coroutine leak.
- Refactored JWT decision/result application so sync and async paths share semantics.
- Fixed case-preserving auth-header replacement exposed by the real-send test (avoids combined old/new Authorization values).

Regression evidence:
- New real async `ClaudeCacheAsyncClient.send()` tests cover async provider success and provider-exception fallback, assert the outgoing token, and assert zero `RuntimeWarning`.
- Added running-loop sync-dispatch regression asserting the coroutine is closed and no `RuntimeWarning` occurs.
- Seam suite with `-W error::RuntimeWarning`: 33 passed.
- Targeted Claude cache/model tests: 132 passed, 2 skipped.
- Full plugin suite: 1,845 passed (one unrelated pre-existing AsyncMock warning elsewhere in the suite; the seam suite is warning-clean under RuntimeWarning-as-error).
- Ruff check --fix, Ruff format, Ruff check, Ruff format --check: passed on all changed Python files.
- git diff --check: passed.

mpfaffenberger:
Independent reviewer clone #2 final re-review of `cd9f7fc2`: **clean / approve (COMMENT review)**.

The running-event-loop blocker is resolved. `ClaudeCacheAsyncClient.send()` now awaits `_should_refresh_token_async()` and `_refresh_claude_oauth_token_async()`; those use async callback dispatchers backed by `_trigger_callbacks`, so sync and async providers are awaited normally with callback error isolation and disabled/no-provider filtering. Provider exceptions safely produce the old-token/no-refresh fallback. The synchronous wrappers remain available for genuinely sync callers; if misused inside a running loop with an async provider, the coroutine is explicitly closed before returning `[None]`, eliminating the previous leak.

The new real-`send()` tests exercise async expiry and refresh providers inside the active loop, verify refreshed versus fallback outgoing Authorization headers, force collection/yield, and assert no `RuntimeWarning`. I independently ran:

- Ruff check + format check on all 7 changed Python files: pass
- `tests/plugins/test_claude_oauth_callback_seams.py` with `-W error::RuntimeWarning`: **33 passed**
- seam + Claude cache tests: **86 passed**

Authentication still requires an explicit `True`, model loading uses the public dispatcher, and error/no-provider semantics remain correct.

Integration note: this PR still adds PhaseType/_callbacks entries in `callbacks.py`; when sequencing with #735/#741/#742/#744 and the ownership helpers from #737/#739, preserve the union of keys and helpers. No remaining blocker found.

PR #736, opened 2026-08-10
Decouple agent_skills plugin (#724)

mpfaffenberger:
Fixes #724.

Core imports were already lazy/function-local at the branch base. The agent_skills plugin now self-registers its enabled filesystem skills through the existing `register_skills` hook without recursing through discovery or changing filesystem-skill precedence.

Tests:
- `uv run ruff check --fix code_puppy/plugins/agent_skills/register_callbacks.py code_puppy/tools/skills_tools.py code_puppy/command_line/skills_completion.py`
- `uv run ruff format --check code_puppy/plugins/agent_skills/register_callbacks.py code_puppy/tools/skills_tools.py code_puppy/command_line/skills_completion.py`
- `git diff --check`
- `uv run pytest tests/test_skills_tools_full_coverage.py tests/command_line/test_skills_completion.py tests/plugins/test_agent_skills.py tests/plugins/test_agent_skills_callbacks_coverage.py -q`
- `uv run pytest tests/plugins/ -q`
- `uv run pytest tests/ -q -p no:cacheprovider`

PR #739, opened 2026-08-10
Decouple universal_constructor plugin (#726)

### LINKED ISSUE #726: Decouple universal_constructor plugin: replace core imports with register_tools / registry hooks
## Context

## Goal

- `register_tools` (core `tools/__init__.py` already calls `register_universal_constructor`)
- `register_agent_tools`
- `startup` (already registered)

|---|---|
| `tools/universal_constructor.py` | top-level import of `.models`; call-time `.registry.get_registry`, `.sandbox`, `USER_UC_DIR` |
| `tools/__init__.py` | call-time `_register_uc_tool_wrapper` -> `.registry.get_registry` |
| `agents/json_agent.py` | call-time `get_available_tools` -> `.registry.get_registry` |
| `agents/agent_creator_agent.py` | call-time `get_system_prompt` / `get_available_tools` -> `.registry.get_registry` |
| `command_line/uc_menu.py` | **top-level** imports of `.models.UCToolInfo` and `.registry.get_registry` |

## Notes

- This is the most-imported plugin (5 core files), so it's the highest value.
- The registry (`registry.py`) is the real interface; consider exposing it via
  a module-level lazy accessor so import-time coupling disappears.
- `uc_menu.py`'s top-level imports of `UCToolInfo` / `get_registry` must be
  moved inside functions (they're only used when the menu opens).

## How to verify

- `grep -rn "code_puppy.plugins.universal_constructor" code_puppy --include='*.py'` outside
  `code_puppy/plugins/universal_constructor/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
Fixes #726.

Replaces core->plugin imports with lazy access + hook-driven registration.

### DIFF
--- a/code_puppy/agents/agent_creator_agent.py
+++ b/code_puppy/agents/agent_creator_agent.py
@@ -33,9 +33,9 @@ class AgentCreatorAgent(BaseAgent):
         # Also get Universal Constructor tools (custom tools created by users)
         uc_tools_info = []
         try:
-            from code_puppy.plugins.universal_constructor.registry import get_registry
+            from code_puppy.tools.universal_constructor import get_uc_registry
 
-            registry = get_registry()
+            registry = get_uc_registry()
             uc_tools = registry.list_tools(include_disabled=True)
             for tool in uc_tools:
                 status = "✅" if tool.meta.enabled else "❌"
--- a/code_puppy/agents/agent_creator_agent.py
+++ b/code_puppy/agents/agent_creator_agent.py
@@ -33,10 +33,12 @@ class AgentCreatorAgent(BaseAgent):
         # Also get Universal Constructor tools (custom tools created by users)
         uc_tools_info = []
         try:
-            from code_puppy.tools.universal_constructor import get_uc_registry
+            from code_puppy.universal_constructor_provider import (
+                get_universal_constructor_provider,
+            )
 
-            registry = get_uc_registry()
-            uc_tools = registry.list_tools(include_disabled=True)
+            provider = get_universal_constructor_provider()
+            uc_tools = provider.list_tools(include_disabled=True) if provider else []
             for tool in uc_tools:
                 status = "✅" if tool.meta.enabled else "❌"
                 uc_tools_info.append(
--- a/code_puppy/agents/json_agent.py
+++ b/code_puppy/agents/json_agent.py
@@ -126,10 +126,12 @@ class JSONAgent(BaseAgent):
         # Also get UC tool names
         uc_tool_names = set()
         try:
-            from code_puppy.tools.universal_constructor import get_uc_registry
+            from code_puppy.universal_constructor_provider import (
+                get_universal_constructor_provider,
+            )
 
-            registry = get_uc_registry()
-            for tool in registry.list_tools():
+            provider = get_universal_constructor_provider()
+            for tool in provider.list_tools() if provider else []:
                 if tool.meta.enabled:
                     uc_tool_names.add(tool.full_name)
         except ImportError:
--- a/code_puppy/agents/json_agent.py
+++ b/code_puppy/agents/json_agent.py
@@ -126,9 +126,9 @@ class JSONAgent(BaseAgent):
         # Also get UC tool names
         uc_tool_names = set()
         try:
-            from code_puppy.plugins.universal_constructor.registry import get_registry
+            from code_puppy.tools.universal_constructor import get_uc_registry
 
-            registry = get_registry()
+            registry = get_uc_registry()
             for tool in registry.list_tools():
                 if tool.meta.enabled:
                     uc_tool_names.add(tool.full_name)
--- a/code_puppy/command_line/uc_menu.py
+++ b/code_puppy/command_line/uc_menu.py
@@ -4,11 +4,13 @@ Provides a split-panel interface for browsing and managing UC tools
 with live preview of tool details and inline source code viewing.
 """
 
+from __future__ import annotations
+
 import asyncio
 import sys
 import unicodedata
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
 from prompt_toolkit.application import Application
 from prompt_toolkit.key_binding import KeyBindings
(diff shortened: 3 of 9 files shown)


mpfaffenberger:

Blocking: this does not decouple core from `universal_constructor`; it moves several imports behind a core shim that still imports the plugin. `code_puppy/tools/universal_constructor.py:18-28` (`get_uc_registry`) lazily imports `plugins.universal_constructor.registry`, `_get_uc_models()` at ~41 imports plugin models, and create/update paths still import `USER_UC_DIR` and `sandbox` (around new lines 460 and 645). `code_puppy/command_line/uc_menu.py` also retains a lazy `USER_UC_DIR` plugin import. The requirement includes lazy imports, so core remains structurally dependent on the plugin.

The new `register_tools` callback only decouples tool registration; it does not provide registry/models/storage/sandbox capabilities. `TYPE_CHECKING = Any` plus deferred Pydantic rebuild adds substantial complexity without removing the dependency, and no tests were changed to validate importability without the plugin, fallback behavior, disabled-plugin behavior, schema rebuild concurrency/idempotency, or callback registration.

Recommended fix: either move the UC implementation and neutral response contracts out of `plugins/` into core (if it is genuinely core-owned), or define a cohesive plugin provider interface for registry/actions and keep plugin model types out of core annotations. Add a test that physically makes the plugin unavailable and imports/uses core surfaces. No `callbacks.py` key collision here, but its new `register_tools` contribution overlaps semantically with existing tool providers and #736's registration approach; verify deterministic ordering/deduplication.

mpfaffenberger:
Addressed the blocking review in 28c77da6.

- Added a core `UniversalConstructorProvider` protocol/singleton seam; core imports no `code_puppy.plugins.universal_constructor` module, lazily or otherwise.
- Added the plugin-owned provider implementation for registry, storage, and sandbox capabilities.
- Replaced deferred plugin-model Pydantic rebuilding with neutral core response contracts.
- Removed core static tool registration; the plugin `register_tools` callback is now the sole deterministic contributor.
- Added subprocess coverage that physically blocks every UC plugin import while importing core surfaces, asserts no plugin module loads/static tool registration exists, plus no-provider and registered-provider seam tests.

Validation:
- Ruff format + format --check: passed on all changed Python files
- Ruff check: passed on all changed Python files
- `git diff --check`: passed
- Targeted suite: 684 passed, 3 warnings
- `tests/plugins/`: 1812 passed, 1 existing AsyncMock warning

mpfaffenberger:
Independent reviewer clone #2 re-review of `28c77da6`: **remaining changes needed**.

Resolved: resulting core code has zero imports of `code_puppy.plugins.universal_constructor`, including lazy imports. Neutral response contracts replaced deferred plugin-model rebuilding. The subprocess test genuinely blocks all UC plugin imports and proves core surfaces import without it; no-provider tool behavior and provider-driven list behavior are tested. Ruff and focused tests pass.

One lifecycle blocker remains: `code_puppy/universal_constructor_provider.py:44-59` stores and returns a bare process-global provider with no loader owner or enabled check. Once the plugin registers, disabling `universal_constructor` does not remove/filter it; core continues exposing UC tools and operations. I reproduced on this head by registering a provider, making `get_disabled_plugins()` return `{"universal_constructor"}`, and observing `get_universal_constructor_provider()` still returns the provider. `clear_universal_constructor_provider()` exists but is only described as primarily for unload/tests and is never wired to plugin disable/unload; replacement also has no identity token to protect against stale unloads.

Recommended fix: use the ownership-aware provider lifecycle introduced in #737 (capture loading context, filter with `is_callback_owner_enabled`, identity-token unregister/reload safety), or expose UC solely through an owned callback provider. Add disabled-owner and stale-unregister/reload tests. This is especially important because `_register_tools` is callback-filtered while the global provider is not, creating inconsistent disabled behavior.

I independently ran Ruff check/format on all 9 changed Python files (pass) and `tests/test_universal_constructor_provider.py`: **5 passed**. No `callbacks.py` edit here, but adopting the shared ownership helpers from #737 creates a dependency/rebase consideration rather than a PhaseType-key conflict.

mpfaffenberger:
Addressed the remaining lifecycle blocker in 1173128b, following #737's ownership-aware provider design.

- Added shared `get_loading_context()` / `is_callback_owner_enabled()` helpers and made callback filtering use the same owner predicate.
- `register_universal_constructor_provider()` now returns an immutable identity token containing provider + owner and captures loader ownership when not explicit.
- `get_universal_constructor_provider()` returns the provider only while its owner is enabled. A loaded-then-disabled `universal_constructor` plugin therefore becomes inactive without losing its registration and reactivates when enabled.
- Added explicit identity-based unregister. Re-registration replaces the provider; stale tokens cannot remove a newer provider.
- Plugin registration is explicitly owned by `universal_constructor`, and its unload helper captures its registration token safely across module reloads.
- Added tests for loader-owner capture, disabled/re-enabled owner behavior, explicit unregister/no-provider fallback, replacement, and stale-unregister safety.
- Core still has zero imports of `code_puppy.plugins.universal_constructor`.

Validation:
- Ruff check --fix, format, check, format --check: passed across all 10 PR-changed Python files
- `git diff --check`: passed
- Targeted UC/tools/agents tests: 688 passed, 3 warnings
- `tests/plugins/`: 1812 passed, 1 existing AsyncMock warning

mpfaffenberger:
Independent reviewer clone #2 final re-review of `1173128b`: **clean / approve (COMMENT review)**.

The provider lifecycle blocker is resolved. Registration captures or accepts an owner and returns an immutable identity token. Lookup applies the same `is_callback_owner_enabled()` predicate as callback dispatch, so a loaded provider disappears while `universal_constructor` is disabled and reappears when re-enabled. Explicit unregister restores the no-provider fallback; re-registration atomically replaces the provider; stale tokens cannot remove the replacement. Plugin registration retains its token and exposes an unregister helper.

I also re-verified the resulting core tree has **zero** imports of `code_puppy.plugins.universal_constructor` (top-level or lazy). The subprocess test physically blocks that namespace while importing all relevant core surfaces, and no-provider/provider-driven behavior remains covered.

Independent validation:

- Ruff check + format check on all 10 changed Python files: pass
- `tests/test_universal_constructor_provider.py`: **9 passed**
- Direct ownership/disable/re-enable/re-register/stale-token lifecycle reproduction: pass
- Core import grep excluding plugin code/tests: none

Integration note: this adds the same generic callback ownership helpers as #737. Resolve sequencing by keeping one copy of those helpers and retaining the PhaseType/_callbacks union from the other decoupling PRs. No remaining blocker found.

PR #740, opened 2026-08-10
Decouple ask-user-question theme integration

### LINKED ISSUE #734: Decouple theme plugin: consume styles via prompt_toolkit_style / termflow_style hooks
## Context

## Goal

- `prompt_toolkit_style` (already registered)
- `termflow_style`, `termflow_highlighter`, `prompt_text_color` (already registered)
- `custom_command`, `startup` (already registered)

|---|---|
| `tools/ask_user_question/theme.py` | call-time `get_rich_colors` -> `.prompt_toolkit_theme.get_style_rules` |

## Notes

- The theme plugin already self-registers all the style hooks. The single
  remaining core import (`get_style_rules`) should be served through the
  `prompt_toolkit_style` hook / registry rather than a direct import.

## How to verify

- `grep -rn "code_puppy.plugins.theme" code_puppy --include='*.py'` outside
  `code_puppy/plugins/theme/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
## Summary
- resolve ask-user-question Rich colors through the public `prompt_toolkit_style` callback chain
- remove the direct dependency on the theme plugin style-rule implementation
- preserve default colors when no plugin-provided style is available
- add focused callback integration tests

## Tests
- `uv run ruff check code_puppy/tools/ask_user_question/theme.py tests/tools/test_ask_user_question/test_theme.py`
- `uv run pytest -q tests/tools/test_ask_user_question/test_theme.py tests/tools/test_ask_user_question tests/test_ask_user_question_full_coverage.py tests/plugins/test_prompt_toolkit_theme.py tests/plugins/test_theme_plugin.py` (226 passed)
- `uv run pytest -q tests/plugins` (1812 passed)
- `git diff --check`

Fixes #734

### DIFF
--- a/code_puppy/tools/ask_user_question/theme.py
+++ b/code_puppy/tools/ask_user_question/theme.py
@@ -126,12 +126,15 @@ _DEFAULT_RICH = RichColors()
 
 def get_rich_colors() -> RichColors:
     """Return Rich styles backed by the shared prompt-toolkit palette."""
-    from code_puppy.plugins.theme.prompt_toolkit_theme import get_style_rules
+    from code_puppy.callbacks import on_prompt_toolkit_style
 
-    muted_rule = get_style_rules().get("tui.muted", "").removeprefix("fg:")
-    if not muted_rule:
+    style = on_prompt_toolkit_style()
+    if style is None:
         return _DEFAULT_RICH
-    muted = muted_rule.split(maxsplit=1)[0]
+
+    muted = style.get_attrs_for_style_str("class:tui.muted").color
+    if not muted or muted == "default":
+        return _DEFAULT_RICH
 
     muted_style = f"{muted} italic"
     return _DEFAULT_RICH._replace(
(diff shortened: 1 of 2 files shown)


mpfaffenberger:

Verified the only core→theme-plugin import (the lazy `get_style_rules` import in `ask_user_question/theme.py`) is removed. It reuses the existing `prompt_toolkit_style` chain, so callback ownership already handles disabled plugins, callback failures preserve the prior/default value, load order follows documented chain semantics, and no new registration/idempotency mechanism is invented. `None` and unresolved/default colors correctly fall back to `_DEFAULT_RICH`.

The focused tests cover provider and no-provider behavior and exercise the public seam. Sync use is appropriate for style resolution. SOLID/DRY/YAGNI look good.

No new `PhaseType`/`_callbacks` edits, so no direct conflict with #735/#738/#741/#742/#744; it only consumes the pre-existing style callback.

PR #741, opened 2026-08-10
Decouple quick resume from statusline

### LINKED ISSUE #731: Decouple statusline plugin: git-branch detection should come from a prompt/status provider hook
## Context

## Goal

- `startup`, `custom_command`, `custom_command_help` (already registered)
- `load_prompt` / `get_model_system_prompt` (available for status injection)

|---|---|
| `config.py` | call-time `get_quick_resume_location` -> `.payload.detect_git_branch` |

## Notes

- Core's config asks the statusline plugin for `detect_git_branch` (used in
  quick-resume location). This is a *data consumer* — introduce a small
  status/prompt provider hook (or reuse `load_prompt`) that the plugin
  registers, and have config query the hook with a None fallback.
- Keep `get_quick_resume_location` behavior identical when plugin is active.

## How to verify

- `grep -rn "code_puppy.plugins.statusline" code_puppy --include='*.py'` outside
  `code_puppy/plugins/statusline/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
## Summary
- add a narrow git-branch provider hook to the existing callback registry
- have the statusline plugin self-register its existing branch detector
- make quick-resume consume the provider with a safe no-plugin fallback
- preserve branch-scoped quick-resume behavior when statusline is loaded

## Tests
- `uv run ruff check code_puppy/callbacks.py code_puppy/config.py code_puppy/plugins/statusline/register_callbacks.py tests/test_quick_resume.py`
- `git diff --check`
- `uv run pytest -q tests/test_quick_resume.py tests/plugins/test_statusline_plugin.py tests/test_config_full_coverage.py` (247 passed)
- `uv run pytest -q tests/plugins` (1812 passed, 1 pre-existing RuntimeWarning)

Fixes #731

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -63,6 +63,7 @@ PhaseType = Literal[
     "post_autosave",
     "notification",
     "awaiting_user_input",
+    "git_branch_provider",
 ]
 CallbackFunc = Callable[..., Any]
 
@@ -126,6 +127,7 @@ _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "post_autosave": [],
     "notification": [],
     "awaiting_user_input": [],
+    "git_branch_provider": [],
 }
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


mpfaffenberger:

Verified the lazy `config.py` import of `plugins.statusline.payload.detect_git_branch` is removed. The plugin self-registers the provider; no/disabled plugin returns `None` via normal ownership filtering; first non-empty provider gives deterministic precedence; callback errors are isolated; duplicate function registration is deduped. Keeping this sync is suitable for quick-resume path calculation.

Tests meaningfully cover provider precedence, absent provider, and the core quick-resume seam. The narrow provider is SOLID and not overbuilt.

Integration note: `callbacks.py` PhaseType/_callbacks additions textually conflict with #735/#738/#742/#744 (and #743 edits nearby). Rebase/resolve by retaining all phase names and registry entries.

PR #742, opened 2026-08-10
Decouple Ollama setup completion provider

### LINKED ISSUE #733: Decouple ollama_setup plugin: expose completer via a registration hook (needs small indirection)
## Context

## Goal

- `custom_command`, `custom_command_help` (already registered)
- `register_cli_args` (available for CLI registration)
- **New-ish:** a completion-provider registration mechanism (or a lazy lookup owned by the plugin)

|---|---|
| `command_line/prompt_toolkit_completion.py` | call-time `get_input_with_combined_completion`, `get_completions` -> `.completer.OllamaSetupCompleter` |
| `messaging/editor_completion.py` | call-time `build_completer` -> `.completer.OllamaSetupCompleter` |

## Notes

- The coupling is a **class** (`OllamaSetupCompleter`). Introduce a
  `register_completion_provider` hook (or reuse a provider registry) that the
  plugin registers its completer class into; core builds completers from the
  registry with a lazy import fallback if absent.
- This needs one small new hook — flag it as the acceptable exception.

## How to verify

- `grep -rn "code_puppy.plugins.ollama_setup" code_puppy --include='*.py'` outside
  `code_puppy/plugins/ollama_setup/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
## Summary
- add a small generic callback seam for plugin completion providers
- have `ollama_setup` self-register its completer
- preserve classic prompt and persistent editor completion without direct Ollama imports
- cover provider registration and absent-plugin fallback

## Tests
- `uv run ruff check code_puppy/callbacks.py code_puppy/command_line/prompt_toolkit_completion.py code_puppy/messaging/editor_completion.py code_puppy/plugins/ollama_setup/register_callbacks.py tests/plugins/test_completion_provider.py`
- `uv run pytest -q --no-cov tests/plugins/test_completion_provider.py tests/messaging/test_editor_completion.py tests/test_prompt_toolkit_completion.py` (71 passed, 1 xpassed)
- `uv run pytest -q --no-cov tests/plugins` (1815 passed)
- `git diff --check`

Fixes #733

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -47,6 +47,7 @@ PhaseType = Literal[
     "register_mcp_catalog_servers",
     "register_browser_types",
     "register_model_providers",
+    "register_completion_provider",
     "message_history_processor_start",
     "message_history_processor_end",
     "on_message",
@@ -110,6 +111,7 @@ _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "register_mcp_catalog_servers": [],
     "register_browser_types": [],
     "register_model_providers": [],
+    "register_completion_provider": [],
     "message_history_processor_start": [],
     "message_history_processor_end": [],
     "on_message": [],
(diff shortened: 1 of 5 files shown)


mpfaffenberger:

Verified both lazy core imports of `plugins.ollama_setup.completer` are removed. The plugin creates its completer through a generic provider; no/disabled plugin yields the base completer list through callback ownership filtering; provider failures are isolated; `None` is ignored; registration dedupes on repeated module loading. Both completion construction paths consume the same seam (nice DRY win).

Tests cover no provider, a generic provider, and actual Ollama registration/completion. The provider is implicitly synchronous; that is appropriate because completer construction is synchronous even when called from the async prompt path. I would keep that contract sync rather than invent async machinery.

Integration note: `callbacks.py` PhaseType/_callbacks additions conflict textually with #735/#738/#741/#744, with #743 nearby. Preserve the union on rebase.

PR #743, opened 2026-08-10
Decouple customizable command dispatch

### LINKED ISSUE #730: Decouple customizable_commands plugin: dispatch through custom_command hook instead of direct import
## Context

## Goal

- `custom_command` (already registered)
- `custom_command_help` (already registered)

|---|---|
| `command_line/command_handler.py` | call-time `_is_namespaced_custom_command`, `handle_command` -> `.register_callbacks` |

## Notes

- The custom command subsystem already self-registers via `custom_command` /
  `custom_command_help`. command_handler should check/dispatch through those
  hooks and not import the plugin module.
- Watch for the `_ensure_plugins_loaded` path — core must guarantee plugins
  are loaded before dispatching to hooks (load_plugins() is idempotent).

## How to verify

- `grep -rn "code_puppy.plugins.customizable_commands" code_puppy --include='*.py'` outside
  `code_puppy/plugins/customizable_commands/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
## Summary
- route custom command execution through the registered `custom_command` callback
- preserve namespaced custom commands while allowing file paths to fall through
- move the custom command input result contract into core callbacks while retaining the plugin alias
- remove all customizable_commands plugin imports from the core command handler

Fixes #730

## Tests
- `uv run ruff check code_puppy/command_line/command_handler.py code_puppy/callbacks.py code_puppy/plugins/customizable_commands/register_callbacks.py tests/command_line/test_command_handler_coverage.py`
- `git diff --check`
- `uv run pytest -q tests/test_command_handler.py tests/command_line/test_command_handler_coverage.py tests/command_line/test_remaining_coverage.py tests/plugins/test_flux_bootstrap.py tests/plugins/test_remaining_coverage.py` (242 passed)
- `uv run pytest -q tests/plugins` (1812 passed, 1 warning)

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -66,6 +66,20 @@ PhaseType = Literal[
 ]
 CallbackFunc = Callable[..., Any]
 
+
+class CustomCommandResult:
+    """Custom command content that should be processed as user input."""
+
+    def __init__(self, content: str):
+        self.content = content
+
+    def __str__(self) -> str:
+        return self.content
+
+    def __repr__(self) -> str:
+        return f"CustomCommandResult({len(self.content)} chars)"
+
+
 _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "startup": [],
     "shutdown": [],
--- a/code_puppy/command_line/command_handler.py
+++ b/code_puppy/command_line/command_handler.py
@@ -8,23 +8,6 @@ import code_puppy.command_line.uc_menu  # noqa: F401
 _PLUGINS_LOADED = False
 
 
-def _is_namespaced_custom_command(first_token: str) -> bool:
-    """True if ``first_token`` (e.g. ``/flux/status``) is a loaded custom command.
-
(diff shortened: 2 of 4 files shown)


mpfaffenberger:

Verified both lazy imports of `plugins.customizable_commands.register_callbacks` are removed. The result marker is now a neutral core callback contract while the plugin keeps a compatibility alias. Namespaced commands and ordinary slash commands use one dispatch function; absent/disabled plugins fall through via normal callback filtering; errors are isolated at both dispatcher layers; repeated registration remains registry-deduped.

Tests meaningfully cover namespaced handled/fallthrough behavior, and existing command-handler tests cover ordinary markdown/custom results. Sync semantics match the existing command callback contract. This is a sensible DIP fix without unnecessary abstraction.

Integration note: this adds `CustomCommandResult` in `callbacks.py` and may conflict contextually with the PhaseType/_callbacks edits in #735/#738/#741/#742/#744. Preserve both the class and the union of their phase keys when rebasing.

PR #744, opened 2026-08-10
Decouple DBOS state from set menu catalog

### LINKED ISSUE #732: Decouple dbos_durable_exec plugin: feature-enabled check should come from a feature registry, not an import
## Context

## Goal

- `startup`, `shutdown`, `agent_run_context`, `agent_run_cancel`, `should_skip_fallback_render`, `custom_command` (already registered)
- `register_cli_args` (available)

|---|---|
| `command_line/set_menu_catalog.py` | **top-level** import-time `from code_puppy.plugins.dbos_durable_exec.config import is_enabled` |

## Notes

- The one core file imports `is_enabled` at module top level. Move the
  enabled-check behind a lazily-populated feature registry the plugin writes
  at `startup`, and make set_menu_catalog query that registry (guarded, lazy)
  instead of importing the module.

## How to verify

- `grep -rn "code_puppy.plugins.dbos_durable_exec" code_puppy --include='*.py'` outside
  `code_puppy/plugins/dbos_durable_exec/` shows only hook lookups / guarded lazy imports,
  no top-level imports.

mpfaffenberger:
## Summary
- remove the set-menu catalog's top-level import of DBOS plugin config
- expose DBOS enabled state through a generic plugin capability callback with a safe false fallback
- preserve configured enabled state even when the optional `dbos` package is unavailable

## Tests
- `uv run ruff format --check code_puppy/callbacks.py code_puppy/command_line/set_menu_catalog.py code_puppy/plugins/dbos_durable_exec/config.py code_puppy/plugins/dbos_durable_exec/register_callbacks.py tests/command_line/test_set_menu.py tests/plugins/test_dbos_durable_exec.py`
- `uv run ruff check code_puppy/callbacks.py code_puppy/command_line/set_menu_catalog.py code_puppy/plugins/dbos_durable_exec/config.py code_puppy/plugins/dbos_durable_exec/register_callbacks.py tests/command_line/test_set_menu.py tests/plugins/test_dbos_durable_exec.py`
- `uv run pytest -q tests/command_line/test_set_menu.py tests/plugins/test_dbos_durable_exec.py tests/plugins/test_dbos_startup_lock.py --no-cov` (89 passed)
- `uv run pytest -q tests/plugins --no-cov` (1812 passed)

Fixes #732

### DIFF
--- a/code_puppy/callbacks.py
+++ b/code_puppy/callbacks.py
@@ -66,6 +66,7 @@ PhaseType = Literal[
     "notification",
     "awaiting_user_input",
     "git_branch_provider",
+    "feature_capability",
 ]
 CallbackFunc = Callable[..., Any]
 
@@ -146,6 +147,7 @@ _callbacks: Dict[PhaseType, List[CallbackFunc]] = {
     "notification": [],
     "awaiting_user_input": [],
     "git_branch_provider": [],
+    "feature_capability": [],
 }
 
 logger = logging.getLogger(__name__)
@@ -291,6 +293,14 @@ def count_callbacks(phase: Optional[PhaseType] = None) -> int:
     if phase is None:
         return sum(len(callbacks) for callbacks in _callbacks.values())
     return len(_callbacks.get(phase, []))
+
+
+def get_feature_capability(name: str) -> bool:
+    """Return the last plugin-provided state for *name*, or safely default false."""
+    results = _trigger_callbacks_sync("feature_capability", name)
+    return next(
+        (result for result in reversed(results) if isinstance(result, bool)), False
+    )
 
 
 def _trigger_callbacks_sync(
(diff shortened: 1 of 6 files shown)


mpfaffenberger:

Verified the top-level `set_menu_catalog.py` import of `plugins.dbos_durable_exec.config` is removed and no other core DBOS import remains. The generic capability callback safely defaults false, callback ownership filters disabled plugins, non-matching providers decline with `None`, bool-only selection avoids accidental truthiness, and last bool gives explicit override precedence. Registration is deduped and the capability is intentionally available even when optional `dbos` is missing, accurately reflecting configured state.

Tests cover the catalog seam, disabled DBOS, enabled-with-package-missing, enabled-with-package-present, unknown capability, and reload idempotency. Sync is appropriate for a config getter. API is generic but still small enough to avoid bespoke callbacks per setting.

Integration note: `callbacks.py` PhaseType/_callbacks additions conflict textually with #735/#738/#741/#742 (and #743 edits nearby). Preserve the union during rebase.

PR #748, opened 2026-08-10
Render recent history when resuming interactively

avargaskun:
## Summary

- re-render recent conversation history when using `-r` / `--resume`
- only render for interactive TTY sessions
- keep headless and piped output unchanged

## Why

This enables tmux-resurrect to restore active Code Puppy panes by resuming a specific session with `-r`, since `--quick-resume` can select the wrong one when multiple sessions share a working directory.

## Testing

- `pytest -q tests/test_cli_runner_resume_render.py` — 2 passed
- `ruff check` — passed
- `ruff format --check` — passed

### DIFF
--- a/code_puppy/cli_runner.py
+++ b/code_puppy/cli_runner.py
@@ -486,6 +486,25 @@ async def main():
                         session=session_name,
                     )
                 )
+                # Re-render the recent conversation so an interactive resume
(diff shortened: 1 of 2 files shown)


thomwebb (approved):
looks good

PR #750, opened 2026-08-10
fix: patch OpenAI response validation for non-compliant providers

kkaileni:
## Problem

Compaction fails **every time** with:

```
Compaction failed: [SummarizationError] LLM call failed during summarization:
[UnexpectedModelBehavior] Invalid response from openai chat completions endpoint:
2 validation errors for ChatCompletion
  choices.0.index - Input should be a valid integer [input_value=None]
  object - Input should be chat.completion [input_value=None]
```

## Root Cause

Some OpenAI-compatible providers (GitHub Copilot GHE proxy, LM Studio, vLLM, etc.) return chat completion responses with `null` for fields the OpenAI SDK marks as required:

- `choices[i].index` (must be `int`, gets `None`)
- `object` (must be `"chat.completion"`, gets `None`)

The OpenAI SDK builds response objects via `model_construct()` (skipping validation), so `None` values pass silently. But pydantic-ai's `_validate_completion` re-validates with `model_validate()`, which explodes.

## Fix

Added `patch_openai_response_defaults()` in `pydantic_patches.py` — patches `OpenAIChatModel._validate_completion` to fill in sensible defaults before validation:

- `object = None` → `"chat.completion"`
- `choices[i].index = None` → `i` (positional index)

Follows the same pattern as pydantic-ai's existing Ollama workaround for `None` finish_reason.

## Testing

- Reproduced the exact error without the patch 
- Patch fixes the validation — defaults applied correctly 
- Full test suite: **7015 passed, 0 failures**

### DIFF
--- a/code_puppy/pydantic_patches.py
+++ b/code_puppy/pydantic_patches.py
@@ -413,6 +413,116 @@ def patch_tool_call_callbacks() -> bool:
         )
 
 
+def patch_openai_response_defaults() -> bool:
+    """Patch OpenAIChatModel._validate_completion to handle missing fields.
+
+    Some OpenAI-compatible providers (GitHub Copilot, LM Studio, Ollama,
+    vLLM, etc.) return chat completion responses with ``None`` for fields
+    the OpenAI SDK marks as required — most commonly:
+
+    - ``choices[i].index`` (must be ``int``, gets ``None``)
+    - ``object`` (must be ``"chat.completion"``, gets ``None``)
+
+    The OpenAI SDK builds the response object via ``model_construct()``
+    (skipping validation), so these ``None`` values pass silently. But
+    pydantic-ai's ``_validate_completion`` re-validates with
+    ``model_validate()``, which explodes on the missing fields.
+
+    pydantic-ai already patches ``finish_reason`` for Ollama in
+    ``_process_response``. This patch extends that philosophy to the
+    other commonly-missing fields, applied generically so *every*
+    OpenAI-compatible backend benefits.
+    """
+    try:
+        from pydantic_ai.models.openai import OpenAIChatModel
+
+        _original_validate = OpenAIChatModel._validate_completion
+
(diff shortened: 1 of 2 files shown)


mpfaffenberger:
I have not experienced this with vLLM or LMStudio - can you tell me exactly how to reproduce?

kkaileni:
> I have not experienced this with vLLM or LMStudio - can you tell me exactly how to reproduce?

i am facing this on all compactions when using the github copilot based models , here is how it looks : 
image

mpfaffenberger:
If this is isolated to only Github Copilot models, we should fix it in that plugin.

mpfaffenberger:
Other question for you - can you verify if the bug is still present after the very recent bump to latest PydanticAI?

kkaileni:
1.verified this is happening even on the latest pydantic ai version bump
2. It looks like the same missing-field issue (choices[i].index = None, object = None) will be triggered  by any non-compliant OpenAI-compatible provider routed through OpenAIChatModel._validate_completion that includes custom_openai configs (LM Studio, Ollama, vLLM) along with the copilot model

kkaileni:
for LMstudio vs copliot 

The code path is identical (OpenAIChatModel._validate_completion → model_validate()), but the HTTP response payload differs. LM Studio returns spec-compliant values ("object": "chat.completion", "index": 0), so validation passes silently. Copilot returns null for both fields, which trips pydantic's strict validation.

The patch is a no-op for compliant providers — the getattr(...) is None guards short-circuit. It lives in pydantic_patches.py because OpenAIChatModel is a shared class;

kkaileni:
@mpfaffenberger can you have a look at this again , thanks.

mpfaffenberger:
> @mpfaffenberger can you have a look at this again , thanks.

yes

john3parker:
I am also experiencing the compaction issue when logged in via Copilot and using Copilot models.

PR #752, opened 2026-08-11
feat(hooks): let a PostToolUse hook withhold a tool's output

tamirkiviti13:
`PostToolUse` currently cannot affect anything. `on_post_tool_call` fires from the `finally` of the `_call_tool` wrapper — after `return result` has already decided the value — and its return is discarded. A hook can see a secret in a tool's output and has no way to keep it out of the model.

This makes a `{"blocked": True, "reason": ...}` verdict from a `post_tool_call` callback substitute a notice for the result.

## What this does and does not do

The tool has **already run** by the time `PostToolUse` fires, and this does not change that. Side effects have happened — the file was read, the command executed. What the verdict controls is where the output *goes*: it stays out of the model's context, out of the message history, and out of your provider's logs. `PreToolUse` remains the way to stop a call from happening at all.

That distinction is stated in the docs rather than left for someone to discover.

## Implementation

The callback moves from the `finally` onto the success path, before the result is handed back. The mechanism is the one already used a few lines above to prepend `PreToolUse` stdout to a tool result, so this isn't new plumbing — just the same rewrite point, with a verdict attached.

The `finally` still fires on the exception path, where there is no result to withhold and the notification stays observational. A flag prevents a successful call from notifying twice.

**Additive for existing consumers.** All eight `post_tool_call` registrants — run stats, subagent panel, ACP bridge, timestamp heartbeat, frontend emitter, quick-resume, herdr, and the hooks bridge — are observational and return nothing, so none of them change behaviour. `_run_post_tool_call` never raises: a hook that throws leaves the result untouched.

## Honest framing

Unlike the other hook work I've sent, this is **not** a Claude Code parity fix. Claude Code's `PostToolUse` cannot withhold a result either — exit 2 there feeds stderr back to the model *alongside* the result. This is a capability request, so it's a separate PR rather than bundled with the compat fixes, and easy to decline on its own merits.

## Stacking

This sits on top of the three-commit hooks-blocking branch (`fix/user-prompt-submit-block-and-session-id`), because it reuses the `_block_reason` helper introduced there. Until that lands, this PR's diff shows those commits too; the commit unique to this PR is the last one. Both branches are rebased onto current `main`.

## Testing

Six new tests: the bridge verdict, the runtime substitution, pass-through when allowed, and a hook that throws.

Full suite on current `main` + this branch: **7124 passed, 8 skipped**. `ruff check` clean; `ruff format --check` flags only two files this branch doesn't touch (`agent_creator_agent.py` and its test, already unformatted on `main`).

Verified end-to-end against a live agent and a real model, on 0.0.702:

- `cat` a file containing a marker → hook blocks → the agent replies *"I can't see what's in that file"*, where before this change it echoed the secret straight back.
- The `Stop` payload confirms the secret is absent from the agent's final response, so it never entered the transcript either.
- `PostToolUse` still receives the real `tool_result`, which is what makes it scannable in the first place — only what flows onward is replaced.

The withheld output is still printed in the operator's own terminal by code puppy's shell display, which runs before the hook. That is expected: this protects the model context, not the local screen.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -99,6 +99,7 @@ from code_puppy.config import (
 )
 from code_puppy.keymap import sigint_fallback_cancels
 from code_puppy.messaging import emit_error, emit_info, emit_warning
+from code_puppy.session_context import get_session_id, set_session_id
 from code_puppy.tools.command_runner import is_awaiting_user_input
 
 # ---- Streaming retry helpers ------------------------------------------------
@@ -642,6 +643,10 @@ async def run_with_mcp(
     global _active_run_depth
     is_nested_run = _active_run_depth > 0
     _active_run_depth += 1
+    # Restored on the way out so a nested run hands its parent's id back, and
+    # the outermost run leaves no stale id behind for code that runs between
+    # turns. The impl publishes this run's own id once it mints one.
+    previous_session_id = get_session_id()
     try:
         return await _run_with_mcp_impl(
             agent,
@@ -654,6 +659,7 @@ async def run_with_mcp(
         )
     finally:
         _active_run_depth -= 1
+        set_session_id(previous_session_id)
 
 
 async def _run_with_mcp_impl(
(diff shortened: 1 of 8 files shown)


tamirkiviti13:
Folding this into #747 rather than keeping it separate.

I originally split it out on the grounds that withholding tool output is a capability Claude Code doesn't have, so it deserved its own debate. That was the wrong yardstick. Code puppy's own `docs/HOOKS.md` documents `exit 1 = Block` as a single contract across all hook events, and the engine resolves a `blocked` verdict for every one of them — `PostToolUse` was simply a second consumer discarding it, exactly like `UserPromptSubmit`. Same defect, same fix, so it belongs in the same PR.

#747 now carries all four commits and has been retitled and rewritten around that. Nothing is dropped.

PR #755, opened 2026-08-11
PUP-585: fix(subagent): fall back instead of hard-failing when a pinned model is unavailable

thomwebb:
### Jira
[PUP-585](https://jira.walmart.com/browse/PUP-585)

### Problem
Code Puppy lets you pin a model to an agent (`/pin <agent> <model>`, or the `"model"` field in a JSON agent file). If that pinned model later disappears from config (deleted entry, unsupported type, or fails to instantiate — e.g. expired API key), two code paths handled it inconsistently:

1. The main interactive agent build path (`_builder.py::build_pydantic_agent`) already degraded gracefully via `load_model_with_fallback()`: warn, then fall back to the global default model, then any other configured model.
2. Sub-agent invocation (`subagent_invocation.py::_invoke_agent_impl`) did **not** use that helper — it hand-rolled a check and raised a bare `ValueError`, hard-failing the whole `invoke_agent`/`invoke_agent_with_model` tool call with zero fallback.

Additionally, once the fallback kicked in, the same warning re-fired on every agent rebuild forever, since a stale pin is never auto-cleared (by design — some failures like an expired API key are transient, so silently deleting a user's explicit pin would be actively harmful).

### Fix (1st commit)
- `subagent_invocation.py` now routes the ambient/pinned model path through the same `load_model_with_fallback()` helper the main agent uses.
- `_builder.py::load_model_with_fallback()` gained a warned-combo dedup, `reset_model_fallback_warnings()`, and fix-instruction text appended to the warning. The pin itself is **never auto-cleared** — that stays a deliberate, human-initiated action.
- `/clear` resets the dedup, so a persistently broken pin gets re-flagged in a fresh conversation instead of going silent forever.

### Self-review round 1 (2nd commit)
1. **`invoke_agent_with_model`'s explicit override now hard-fails instead of silently substituting.** A bad explicit override is now a hard, per-call `ValueError` naming the requested + available models. The ambient/pinned path is unaffected.
2. **ACP multi-session warning suppression** (attempted fix): `_make_session` started resetting the warned-combo set on every session, since it previously only reset via the CLI's `/clear`.

### Self-review round 2 (3rd commit)
Round 1's ACP fix for #2 traded one bug for another: the dedup key had no conversation identity (`(agent_name, model)` only), so a blanket reset on every new session wiped *other live sessions'* already-earned warnings too. Fix: `load_model_with_fallback()` gained an explicit `conversation_scope` folded into the key as `(conversation_scope, agent_name, model)`, sourced (at the time) from `get_session_context()`.

### Self-review round 3 (4th commit)
`get_session_context()` turned out to be the wrong primitive: a plain, lock-protected instance attribute on a process-wide singleton, **not** a `contextvars.ContextVar` — zero logical isolation between concurrent asyncio tasks. Fixed by adding `set_conversation_root_id`/`get_conversation_root_id` to `subagent_context.py`, backed by a real `ContextVar`, set once at the true conversation root (ACP's `session.py` prompt handler). Concurrent tasks now correctly get isolated copies of the ContextVar; nested sub-agent invocations correctly inherit the same root. Also added `close_session` purging its own session-scoped bucket via `reset_model_fallback_warnings(scope=session_id)`.

### Self-review round 4 (5th commit)
Round 4 review independently traced the ContextVar propagation claim against the real pydantic-ai/acp library internals (task creation, tool dispatch, parallel-tool-call batching) and confirmed it holds — **verdict: merge-ready, no blockers.** It flagged two non-blocking follow-ups, both closed here:
1. **Narrow close/in-flight-run race**: `close_session`'s bare `session.cancel()` only *requests* cancellation (delivered at the task's next await point, not immediately), then immediately purged the warn-dedup bucket — a still-in-flight run past its `load_model_with_fallback` write could race the purge and leak one `(session_id, agent_name, model)` tuple forever. Fixed with `ACPSession.cancel_and_wait()` (cancel + await the task, bounded by a timeout), wired into `close_session`. New regression test races `close_session` against a genuinely in-flight run.
2. **Doc drift**: `_make_session`'s docstring still described round-2/3 phrasing ("parent conversation's session id"); updated to name the round-3 ContextVar concept correctly.

### Tests
- `tests/test_pinned_model_fallback.py`: ambient-path fallback + warn-once-per-combo + fix instructions + `/clear` re-arm; explicit-override hard-fail; independent conversations don't share warning state; scoped reset doesn't wipe a different scope; direct unit-test class on `load_model_with_fallback`'s scoping/reset semantics.
- `tests/tools/test_subagent_context_conversation_root.py`: real (unmocked) `asyncio.gather`-based tests proving concurrent tasks never cross-contaminate the ContextVar, and nested calls inherit the same root.
- `tests/command_line/test_session_commands.py`: `/clear` calls `reset_model_fallback_warnings()`.
- `tests/test_acp_plugin.py`: `new_session` resets the shared/unscoped bucket; `close_session` purges its own session-scoped bucket; `close_session` waits for a genuinely in-flight run before purging (new race regression test).
- Full suite (7041 tests) passes; `ruff check` clean.

### Acceptance criteria
- [x] A sub-agent pinned to a model no longer in config completes its run on the fallback model instead of erroring out (ambient/pinned path only).
- [x] An explicit `invoke_agent_with_model` override that's unavailable fails loudly and immediately, naming the model and available alternatives — never silently runs a different model than requested.
- [x] The fallback warning names the dead model + agent and gives actionable `/pin`/`/unpin`/`/model` instructions.
- [x] The same (agent, dead-model) combo doesn't repeat the warning within one conversation — true even across concurrent conversations sharing a process, and across nested sub-agent chains within one conversation, backed by a real ContextVar rather than a shared mutable global.
- [x] `/clear` re-arms the warning for a new CLI conversation; ACP session close purges exactly that session's own warning bucket (waiting for any in-flight run to finish first), never another live session's.
- [x] The pinned model config value is never silently modified/cleared by this fallback logic.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -307,15 +307,52 @@ def reload_mcp_servers(agent_name: Optional[str] = None) -> List[Any]:
     return manager.get_servers_for_agent(agent_name=agent_name)
 
 
+# (agent_name, requested_model_name) combos we've already warned about this
+# conversation. Deliberately NOT cleared when a model finally loads again --
+# only reset_model_fallback_warnings() (wired to /clear) resets it, so a
+# still-broken pin doesn't re-nag on every agent rebuild within one
+# conversation, but a fresh conversation gets the reminder again. We
+# deliberately do NOT auto-clear the pin itself -- see
+# config.clear_agent_pinned_model; that stays a human decision.
+_warned_model_fallbacks: Set[Tuple[Optional[str], str]] = set()
+
+
+def reset_model_fallback_warnings() -> None:
+    """Forget which (agent, model) fallback warnings have already fired.
+
+    Called when the user starts a fresh conversation (/clear) so a
+    persistently broken pin gets re-flagged instead of going silent forever.
+    """
+    _warned_model_fallbacks.clear()
+
+
+def _model_fallback_fix_hint(agent_name: Optional[str]) -> str:
+    """Build the how-to-fix tail appended to model fallback warnings."""
+    if agent_name:
+        return (
+            f"Fix it with `/pin {agent_name} <model>` once a working model is "
+            f"available, or `/unpin {agent_name}` to just track the global "
+            "default from now on. Run `/model` to see configured models."
+        )
+    return "Set a valid model with `/model`, or check your models configuration."
+
+
 def load_model_with_fallback(
     requested_model_name: str,
     models_config: Dict[str, Any],
     message_group: str,
+    agent_name: Optional[str] = None,
 ) -> Tuple[Any, str]:
     """Load the requested model, or fall back to a sensible alternative.
 
     Falls back in order: the globally configured model, then any other
     configured model. Raises ``ValueError`` only if nothing loads.
+
+    ``agent_name``, when given, scopes the model-unavailable warning to fire
+    once per (agent, requested model) combo per conversation (see
+    ``reset_model_fallback_warnings``) and tailors the fix instructions to
+    that agent's ``/pin``/``/unpin`` commands. The agent's pinned model is
+    never auto-cleared -- that stays a deliberate, human-initiated action.
     """
     try:
         model = ModelFactory.get_model(requested_model_name, models_config)
@@ -330,30 +367,38 @@ def load_model_with_fallback(
         available_str = (
             ", ".join(sorted(available)) if available else "no configured models"
         )
+        warn_key = (agent_name, requested_model_name)
+        already_warned = warn_key in _warned_model_fallbacks
+        _warned_model_fallbacks.add(warn_key)
+        fix_hint = _model_fallback_fix_hint(agent_name)
+
         # Distinguish between "key missing", "type unsupported", and "creation failed"
         exc_msg = str(exc)
-        if "not found in configuration" in exc_msg:
+        if already_warned:
+            pass
+        elif "not found in configuration" in exc_msg:
             emit_warning(
-                f"Model '{requested_model_name}' not found. Available models: {available_str}",
+                f"Model '{requested_model_name}' not found. Available models: "
+                f"{available_str}. {fix_hint}",
                 message_group=message_group,
             )
         elif "Unsupported model type" in exc_msg:
             model_type = models_config.get(requested_model_name, {}).get("type", "?")
             emit_warning(
                 f"Model type '{model_type}' is not supported (model '{requested_model_name}'). "
-                f"Available models: {available_str}",
+                f"Available models: {available_str}. {fix_hint}",
                 message_group=message_group,
             )
         elif "could not be instantiated" in exc_msg:
             emit_warning(
                 f"Model '{requested_model_name}' could not be instantiated. "
-                f"Available models: {available_str}",
+                f"Available models: {available_str}. {fix_hint}",
                 message_group=message_group,
             )
         else:
             emit_warning(
                 f"Model '{requested_model_name}' failed: {exc_msg}. "
-                f"Available models: {available_str}",
+                f"Available models: {available_str}. {fix_hint}",
                 message_group=message_group,
             )
 
@@ -526,7 +571,10 @@ def build_pydantic_agent(
 
     models_config = ModelFactory.load_config()
     model, resolved_model_name = load_model_with_fallback(
-        agent.get_model_name(), models_config, message_group
+        agent.get_model_name(),
+        models_config,
+        message_group,
+        agent_name=getattr(agent, "name", None),
     )
     instructions = _assemble_instructions(agent, resolved_model_name)
     mcp_servers = load_mcp_servers(agent_name=getattr(agent, "name", None))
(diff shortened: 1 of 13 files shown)


thomwebb:
Ran this through an adversarial self-review before asking for human eyes. It found two real gaps in the first commit, both fixed in the follow-up commit (`642f9d65`):

1. `invoke_agent_with_model`'s explicit override was silently falling back to a different model instead of failing loudly \u2014 that broke the tool's "run exactly this model" contract for automated callers. Now hard-fails with a clear error naming the requested + available models. The ambient/pinned-model path (`invoke_agent`, no override) is untouched and still degrades gracefully, which is correct there.
2. The once-per-conversation warning dedup only ever reset via the CLI's `/clear`, so a long-running ACP server with multiple concurrent client sessions would silently suppress a dead-pin warning for every session after the first one that hit it \u2014 including sessions that never saw the original warning. `new_session`/`load_session`/`resume_session`/`fork_session` now all reset the dedup set via their shared `_make_session` choke point.

4 new regression tests added for both; full suite (7025 tests) green, ruff clean. PR description updated to reflect both rounds.

thomwebb:
Round 2 adversarial self-review update (commit `b4ad491f`): round 1's ACP fix for cross-session warning suppression was itself flawed \u2014 it reset the dedup set globally on every `new_session`, which fixed the silent-suppression bug but introduced a new one: an unrelated session B's creation could wipe session A's already-earned warning mid-conversation (and made `new_session` a trivial spam vector).

Root cause: the dedup key had no conversation identity at all, just `(agent_name, model)`. Real fix: `load_model_with_fallback()` now takes an explicit `conversation_scope`, keyed as `(conversation_scope, agent_name, model)`. Sub-agent invocation scopes it to the parent conversation's session id, so independent conversations (concurrent ACP sessions) get naturally separate buckets and can't cross-contaminate \u2014 no reset-on-session-creation needed for that path at all. `reset_model_fallback_warnings()` now also accepts an optional `scope` so ACP's remaining reset (for the main-agent-build path, which still shares one bucket by design) only clears its own bucket, never another live session's.

7 more regression tests (a direct unit-test class on the scoping semantics + 2 integration tests proving cross-session isolation). Full suite (7032 tests) green, ruff clean. PR description updated again.

thomwebb:
Round 3 adversarial self-review update (commit `e7c81c5e`): round 2's `conversation_scope` was itself built on the wrong primitive. `get_session_context()` (`code_puppy/messaging/bus.py`) is a plain, lock-protected instance attribute on a process-wide singleton, **not** a `contextvars.ContextVar` — the lock stops torn reads/writes but gives zero logical isolation between concurrent asyncio tasks. That meant two residual bugs: (1) concurrent conversations (concurrent ACP sessions, or parallel sibling tool calls under `yolo_mode`) could still read each other's in-flight session id, and (2) nested sub-agent invocations (A invokes B) each mint a fresh transient session id, so a nested call's scope was its parent's transient id rather than the true conversation's identity — degrading "warn once per conversation" to "warn once per parent call."

Real fix: `code_puppy/tools/subagent_context.py` (which already solves this exact class of problem for sub-agent depth/name/chain tracking) gains `set_conversation_root_id`/`get_conversation_root_id` backed by a genuine `ContextVar`, set once at the true conversation root (ACP's `session.py` prompt handler, next to the existing `set_working_directory` ContextVar pattern) and left `None` for the CLI. `subagent_invocation.py` now reads that instead of `get_session_context()`. Proved both properties (no cross-contamination between concurrent tasks, correct inheritance through nested calls) with real, unmocked `asyncio.gather` tests in a new `tests/tools/test_subagent_context_conversation_root.py` — not just mock-consistency tests.

Also closed a leak the same review flagged: per-session warning buckets now never got cleaned up on session close. `close_session` now purges exactly that session's own bucket via `reset_model_fallback_warnings(scope=session_id)`.

Full suite (7040 tests) green, ruff clean. PR description updated again to reflect all 4 commits' history.

thomwebb:
Round 4 adversarial self-review (commit `52efd402`): the reviewer independently traced round 3's ContextVar propagation claim against the real `pydantic-ai`/`acp` library internals (task creation via `asyncio.ensure_future`/`create_task`, async tool dispatch, parallel-tool-call batching) rather than trusting the new unmocked test alone, and confirmed it holds correctly for this codebase's actual call shape. **Verdict: merge-ready, no blockers.**

Two non-blocking follow-ups were flagged and closed in this commit anyway since they were cheap:

1. **Narrow race**: `close_session` cancelled an in-flight run via the bare, fire-and-forget `ACPSession.cancel()` — `task.cancel()` only *requests* cancellation, delivered at the task's next await point, not immediately — then immediately purged that session's warn-dedup bucket. A run still mid-flight past its `load_model_with_fallback` write could race the purge and leak exactly one `(session_id, agent_name, model)` tuple forever (nothing else will ever purge that session's bucket once closed). Fixed with `ACPSession.cancel_and_wait()` (cancel + await the task, bounded by a timeout so a stuck run can't hang session close forever), wired into `close_session`. Added a regression test that races `close_session` against a genuinely in-flight slow run and confirms it doesn't return until the run has actually finished unwinding.
2. **Doc drift**: `_make_session`'s docstring still described round-2/3 phrasing ("parent conversation's session id"); updated to correctly name the round-3 ContextVar-based "conversation root id" concept.

Full suite (7041 tests) green, ruff clean. PR description updated to reflect all 5 commits.

AndrewTilson (approved):

PR #757, opened 2026-08-12
fix: recover from corrupted config files

thomwebb:
**Jira:** [PUP-605](https://jira.walmart.com/browse/PUP-605)

## Problem

A user hit an uncaught `MemoryError` traceback on startup:

```
_apply_default_theme_on_first_run -> _apply_theme -> set_config_value
  -> configparser.ConfigParser().read(CONFIG_FILE) -> MemoryError
```

`config.py` had ~10 call sites doing raw `configparser.ConfigParser(); config.read(CONFIG_FILE)` / `open(CONFIG_FILE, "w")` with zero error handling. A malformed or pathologically-shaped `puppy.cfg` (e.g. one giant line with no newline -- plausible from AV/EDR interference or a flaky network-mounted home directory, which the reporter's VDI-style hostname suggests) makes stdlib `configparser`'s buffered line reader balloon memory instead of raising a clean parse error, crashing whatever caller touches config next.

## Fix

New `code_puppy/config_file.py` centralizes all config I/O behind two entry points; `config.py` no longer opens `CONFIG_FILE` directly anywhere:

- **`load_config()`** -- reads are byte-bounded *before* parsing, so a pathological file can never reach the unbounded stdlib reader in the first place. Only confirmed parse/decode/oversize corruption is quarantined (renamed aside with a collision-resistant name, never deleted, never overwrites a prior backup). Ordinary `OSError` (permissions, transient network-mount hiccups) propagates untouched instead of being misclassified as corruption and silently discarded. A second read happens under a cross-process lock before anything is quarantined, closing the read/quarantine TOCTOU race -- if another process already fixed the file, its healthy version wins.
- **`mutate_config()`** -- every write (`set_config_value`, `reset_value`, `set_model_name`, `clear_model_settings`, `ensure_config_exists`) is now one locked read-modify-write transaction using a same-directory temp file + `fsync` + atomic `os.replace`, so concurrent writers can't lose updates and a crash mid-write can't reintroduce corruption.

This design went through an internal adversarial review pass; the first draft (a simpler catch-and-quarantine wrapper) was rejected for treating *any* `OSError` as corruption, having a read/quarantine race, and using collision-prone quarantine filenames. All three are addressed above.

## Tests

`tests/test_config_corruption_resilience.py` covers the original crash chain plus the hardening: bounded reads (including a "lying" file-size case), oversized-file handling, quarantine collision safety, transient-I/O propagation, the TOCTOU recovery race, atomic-write failure safety, and lock-timeout behavior.

Full suite: **7120 passed, 0 failed** (`uv run pytest tests/ -q`).

### DIFF
--- a/code_puppy/config.py
+++ b/code_puppy/config.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 from typing import Any, Optional
 
+from code_puppy.config_file import load_config, mutate_config
 from code_puppy.session_storage import save_session
 
 logger = logging.getLogger(__name__)
@@ -307,6 +308,11 @@ _default_vision_model_cache = None
 _warned_no_model = False
 
 
+def _load_config() -> configparser.ConfigParser:
+    """Load ``CONFIG_FILE`` through the bounded, recoverable I/O layer."""
+    return load_config(CONFIG_FILE)
+
+
 def ensure_config_exists():
     """
     Ensure that XDG directories and puppy.cfg exist, prompting if needed.
@@ -317,15 +323,17 @@ def ensure_config_exists():
         if not os.path.exists(directory):
             os.makedirs(directory, mode=0o700, exist_ok=True)
     exists = os.path.isfile(CONFIG_FILE)
-    config = configparser.ConfigParser()
-    if exists:
-        config.read(CONFIG_FILE)
+    # Skip the read entirely when we already know there's nothing to read --
+    # matches configparser's own no-op-on-missing-file behavior and avoids an
+    # unnecessary open() attempt during first-run setup.
+    config = _load_config() if exists else configparser.ConfigParser()
     missing = []
     if DEFAULT_SECTION not in config:
         config[DEFAULT_SECTION] = {}
     for key in REQUIRED_KEYS:
         if not config[DEFAULT_SECTION].get(key):
             missing.append(key)
+    prompted_values: dict[str, str] = {}
     if missing:
         # Note: Using sys.stdout here for initial setup before messaging system is available
         import sys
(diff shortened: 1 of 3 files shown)


AndrewTilson:
Nice patch — the design is tight and the tests pin exactly the right invariants (bounded read before parse, transient-`OSError` propagation, quarantine collision retry, TOCTOU under lock). Approving.

One thing worth capturing before it evaporates: while auditing whether anywhere *else* in the codebase still bypasses this new layer, I confirmed the INI side is fully covered (no module outside `config.py` / `config_file.py` imports `configparser`, and the only remaining `ConfigParser()` outside `config_file.py` is the intentional short-circuit in `ensure_config_exists`). But several **adjacent JSON config files** have the same three failure modes this PR fixes (unbounded read, torn writes, no cross-process lock), and are arguably more exposed since users hand-edit them:

**Tier 1 — `mcp_servers.json` (hand-edited, four concurrent writers, zero atomicity):**
- `command_line/mcp/wizard_utils.py:300–316`
- `command_line/mcp/custom_server_form.py:435–459`
- `command_line/mcp/custom_server_installer.py:172–187`
- `command_line/mcp/remove_command.py:78–88`

All four do raw `open("r") → json.load → mutate → open("w") → json.dump`. Two wizard flows in different terminals (or wizard + `/mcp remove`) will lose each other's updates, and a giant hand-edited `mcp_servers.json` will balloon `json.load` the same way `configparser` ballooned in PUP-605.

**Tier 2 — `extra_models.json` (three plugins racing the same file, three different write patterns):**
- `plugins/aws_bedrock/utils.py` `save_extra_models` — closest to correct (temp+replace) but no `fsync`, no lock
- `plugins/ollama_setup/register_callbacks.py:107–150` — no atomicity, no lock
- `command_line/add_model_menu.py:757–808` — raw `open("w")`

**Tier 3 — cheap fixes:** `plugins/puppy_spinner/spinners.py:319, 439` (`spinners.json`); `~/.code_puppy/agents/*.json` reads at startup are also unbounded.

**Already correct, don't touch:** `secret_store.py` (`_write_fallback` already does `mkstemp` + `chmod 0o600` + `fsync` + `os.replace` + cross-process lock).

If it turns out to be worth doing, the natural shape is to promote the three primitives already living in `config_file.py` — bounded-read, atomic-write, `fcntl`/`msvcrt` path lock — into a shared `code_puppy/atomic_io.py`, then layer today's `load_config`/`mutate_config` on top for INI and a parallel `atomic_json.load_json`/`mutate_json` for JSON. The four MCP writers then collapse to `mutate_json(MCP_SERVERS_FILE, lambda d: ...)` one-liners sharing one lock.

Explicitly **not** asking for any of this in #757 — the current scope is tight, the fix is complete on its own terms, and expanding it would delay a clean crash fix. Just worth a follow-up ticket so the pattern gets pushed out to the rest of the config surface before someone hits the same `MemoryError` on `mcp_servers.json` next.

AndrewTilson (approved):
Approving — clean, well-tested fix for the PUP-605 crash. Design deliberately avoids each of the classic traps (catch-all OSError → data loss, read/quarantine race, collision-prone backups, torn writes), and the regression suite pins both the fix and each finding from the internal adversarial review. Left a separate comment noting several adjacent JSON config files (`mcp_servers.json`, `extra_models.json`, `spinners.json`) that have the same class of bugs and would benefit from a follow-up applying this pattern — explicitly out of scope for this PR.

PR #760, opened 2026-08-12
feat: add model_select hook for per-run model routing

bporterfielddsc:
## Summary

Adds a new `model_select` callback phase that lets a plugin choose the model for
**each run** — e.g. a small model for trivial turns and a frontier model when it
matters (cost control, latency SLAs, and provider failover are other natural
uses). Purely additive: with no plugin registered, behaviour is byte-for-byte
unchanged and the cost is one dict lookup + an empty-list loop per run.

Motivation: today a single agent uses one model for every turn, so trivial edits
burn the same frontier tokens as hard design work. The plumbing to swap the
per-run model already exists (`_runtime_model_name_override`), and
`RoundRobinModel` already exists specifically for "rate-limit mitigation" — but
there's no hook that *decides* which model a turn should use. This fills that gap.

## What changed

| File | Change |
|---|---|
| `callbacks.py` | New `model_select` phase + `on_model_select(...)` trigger (returns the first non-empty model name; `None` = defer). |
| `agents/base_agent.py` | New `_auto_model_override` slot + getter/setter; `get_model_name()` precedence now `runtime override > model_select > pinned > global`. |
| `agents/json_agent.py` | Its `get_model_name()` now honours the auto override **above** the JSON `model` field, so per-turn routing works for JSON agents too. |
| `model_switching.py` | New `resolve_run_model_selection(agent, messages, session_id)`: resets any prior turn's choice, respects an explicit runtime override, calls the hook, and invalidates the cached pydantic agent when the model changes. Never raises. |
| `agents/_runtime.py` | Calls `resolve_run_model_selection(...)` once at the start of every run, before the pydantic agent is (re)built. |
| `tests/test_model_select_hook.py` | 8 tests: hook contract, precedence, cache invalidation, reset-per-run, and error isolation. |

Net: **6 files, +272/-1.**

## Design & precedence

```
explicit runtime override  >  model_select hook  >  pinned / JSON model  >  global
```

- Fires **once per run**, before the build, so the chosen model is actually used.
- Self-cleaning: the auto choice is reset at the start of every run, so it never
  leaks into a later turn — no fragile try/finally around the run body.
- Error-isolated: a selector that raises is swallowed and the run proceeds on its
  normal model (a plugin must never crash the app).

## Hook contract

```python
def on_model_select(
    *,
    agent_name: str,
    current_model: str | None,   # what would be used absent any hook
    messages: list,              # message history for this run
    session_id: str | None,
) -> str | None:                 # a model name for this run, or None to defer
```

## Backward compatibility

Additive only. No existing behaviour changes when the hook is unregistered
(the common case), and the phase short-circuits on an empty callback list.

## Testing

- `tests/test_model_select_hook.py` — **8 passed** (hook + resolver).
- Regression: `tests/test_pinned_model_fallback.py tests/agents/` +
  the new file — **360 passed, 12 skipped, 0 failed**.
- `ruff check` and `ruff format --check` clean on all touched files.

## Decision points I'd love a call on (why this is a Draft)

1. **Hook name** — `model_select` vs `select_model` vs `route_model`?
2. **Precedence** — is `model_select` above pinned/JSON the behaviour you want,
   or should a pinned model win over the hook?
3. **Fire location** — start of `_run_with_mcp_impl` (current) vs inside
   `build_pydantic_agent` (the model choke point, but it isn't called every turn
   because the agent is cached)?
4. **Multiple selectors** — first-non-None wins now; want last-wins or a
   documented priority instead?

## Follow-up (not in this PR)

With this hook, an "auto" model mode becomes a tiny plugin that reuses a
complexity heuristic to return `small`/`large`. I have a working reference
plugin that consumes this hook end-to-end and can follow as a second PR once the
hook shape lands.

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -696,6 +696,20 @@ async def _run_with_mcp_impl(
         # Hook failures must never block the run.
         pass
 
+    # Let a ``model_select`` hook route THIS turn to a different model (e.g.
+    # small-vs-large by complexity) before the pydantic agent is built. This
+    # resets any prior turn's auto choice, respects an explicit runtime
+    # override, and invalidates the cached agent when the model changes so the
+    # build below picks it up. No-op (and near-zero cost) if no plugin
+    # registered the hook.
+    try:
+        from code_puppy.model_switching import resolve_run_model_selection
+
+        resolve_run_model_selection(agent, agent._message_history, group_id)
+    except Exception:
+        # Selection must never block a run.
+        pass
+
     if agent._code_generation_agent is None:
         build_pydantic_agent(agent)
     pydantic_agent = agent._code_generation_agent
--- a/code_puppy/agents/base_agent.py
+++ b/code_puppy/agents/base_agent.py
@@ -73,6 +73,11 @@ class BaseAgent(ABC):
         self._code_generation_agent: Any = None
         self._last_model_name: Optional[str] = None
         self._runtime_model_name_override: Optional[str] = None
+        # Model chosen by a ``model_select`` hook for the current run. Slots
+        # BELOW an explicit runtime override but ABOVE pinned/JSON/global, and
+        # is reset at the start of every run (see resolve_run_model_selection),
+        # so it never leaks across turns.
+        self._auto_model_override: Optional[str] = None
         self._puppy_rules: Optional[str] = None
         self._mcp_servers: List[Any] = []
         self.cur_model: Optional[pydantic_ai.models.Model] = None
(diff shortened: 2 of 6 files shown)


kvandre12-commits:
I refreshed this proposal onto current `main` in companion PR #818 because the original branch is now conflict-stale.

The refresh preserves `b09e033e` as a commit authored by Blayne Porterfield; the current-main compatibility changes are a separate commit. The only merge conflict was the newer runtime system-prompt state in `BaseAgent`, and both states are preserved.

The companion also addresses one contract gap discovered during the refresh: selectors now receive the current prompt after `user_prompt_submit` rewriting, rather than only prior history. This makes the stated trivial-vs-complex turn routing use case implementable. First-non-empty callback precedence now genuinely short-circuits lower-priority selectors, and the hook is documented in `AGENTS.md`.

Validation: 410 agent/pinned/model-select tests passed; focused callback/model tests report 48 passed; Ruff and `git diff --check` are clean.

Companion: https://github.com/mpfaffenberger/code_puppy/pull/818

kvandre12-commits:
Scope correction for #818: I removed the proposed shared callback-dispatch short-circuit change after review. The companion is now limited to refreshing this hook on current `main`, exposing the current rewritten prompt required by the routing contract, and documenting the hook. Existing callback execution semantics are unchanged.

The narrowed branch was rerun successfully: 410 agent/model tests and 48 focused callback/model tests pass.

PR #761, opened 2026-08-12
fix: extend atomic, bounded, locked config I/O to JSON config surfaces

thomwebb:
Jira: [PUP-607](https://jira.walmart.com/browse/PUP-607)

## Follow-up to #757

Andrew Tilson's review comment on #757 approved the INI config fix but flagged that several adjacent JSON config files share the same three failure modes the original crash exposed: unbounded reads, torn writes, and no cross-process lock -- and noted these are arguably more exposed since users hand-edit them directly.

This PR closes that gap.

### What changed

**Extracted the shared primitives.** `config_file.py`'s bounded-read / cross-process-lock / atomic-write logic is now a generic module, `code_puppy/atomic_io.py`. `config_file.py` is rebuilt on top of it with no behavior change -- the existing 17 regression tests from #757 pass unmodified against the new internals.

**Added the JSON counterpart.** `code_puppy/atomic_json.py` provides `load_json` / `mutate_json`, mirroring `config_file.py`'s API. One deliberate difference: a hand-edited JSON file that fails to parse is surfaced to the caller as `JsonFileCorrupt` rather than auto-quarantined. A user's one-typo mistake is far more likely than bitrot here, and silently resetting the file would destroy the rest of their hand-edited config -- a worse outcome than the crash this is meant to prevent.

**Migrated every site from the review, by tier:**

- **Tier 1 -- `mcp_servers.json`** (4 concurrent writers): the install wizard, the custom-server form (including the rename-in-place case), the custom-server installer, and `/mcp remove` all used to hand-roll `open->json.load->mutate->open->json.dump`. Collapsed into one shared `code_puppy/command_line/mcp/mcp_servers_store.py` built on `mutate_json`, so two wizard flows (or a wizard racing `/mcp remove`) can no longer lose each other's writes.
- **Tier 2 -- `extra_models.json`** (3 different write patterns across `aws_bedrock`, `azure_foundry`, `ollama_setup`, and `add_model_menu.py`): all four now go through `load_json` / `mutate_json` / locked `atomic_write_bytes`, closing the "closest to correct but no fsync/lock" gap and the two fully-unlocked writers.
- **Tier 3 -- cheap fixes**: `puppy_spinner`'s `load_user_spinners` / `save_interval_tweak` / `write_template`, and `JSONAgent._load_config`'s startup read of `~/.code_puppy/agents/*.json` (an unbounded read at every CLI startup).

### Post-review fixes

An adversarial self-review pass (see PR comments) caught a lost-update race: `aws_bedrock` and `azure_foundry`'s add/remove functions were splitting the read and write across two unlocked calls, so a concurrent writer to the same `extra_models.json` could have its update silently overwritten. All 5 affected functions now go through one locked `atomic_json.mutate_json()` transaction, with real-thread concurrency tests pinning the fix. Also tightened a `null`-vs-missing ambiguity in `atomic_json.load_json`/`mutate_json` (private sentinel) and aligned `azure_foundry`'s exception handling with `aws_bedrock`'s.

### Testing

- `tests/test_atomic_io.py` and `tests/test_atomic_json.py`: direct coverage of the primitives (bounded reads, oversize handling even when `stat` lies, quarantine collision safety, atomic write failure/rollback, lock contention/timeout, concurrent-writer safety).
- `tests/agents/test_json_agent_corruption_resilience.py`: pins the JSON-agent bounded-read fix specifically.
- Updated `tests/test_config_corruption_resilience.py` and `tests/command_line/mcp/test_custom_server_form.py` where internals moved (patches now target `atomic_io`/`mcp_servers_store` instead of the old raw `os`/`open` call sites); all previously-covered behavior still holds.
- Full suite: **7182 passed, 23 skipped, 1 xpassed**, no regressions.
- `ruff check` and `ruff format --check` both clean.

cc @AndrewTilson -- thanks for the thorough review, this should close out the loose ends you flagged.

### DIFF
--- a/code_puppy/agents/json_agent.py
+++ b/code_puppy/agents/json_agent.py
@@ -1,9 +1,10 @@
 """JSON-based agent configuration system."""
 
-import json
 import logging
 from pathlib import Path
 from typing import Any, Dict, List, Optional
+
+from code_puppy import atomic_json
 
 from .base_agent import BaseAgent
 
@@ -25,14 +26,24 @@ class JSONAgent(BaseAgent):
         self._validate_config()
 
     def _load_config(self) -> Dict:
-        """Load configuration from JSON file."""
+        """Load configuration from JSON file.
+
+        Bounded via :mod:`code_puppy.atomic_json` -- a pathologically large
+        agent file dropped in ``~/.code_puppy/agents/`` can no longer
+        balloon memory the way the original ``configparser`` bug did.
+        """
         try:
-            with open(self.json_path, "r", encoding="utf-8") as f:
-                return json.load(f)
-        except (json.JSONDecodeError, FileNotFoundError) as e:
+            data = atomic_json.load_json(self.json_path)
+        except (atomic_json.JsonFileCorrupt, OSError) as e:
             raise ValueError(
                 f"Failed to load JSON agent config from {self.json_path}: {e}"
             ) from e
+        if data is None:
+            raise ValueError(
+                f"Failed to load JSON agent config from {self.json_path}: "
+                "file not found or empty"
+            )
+        return data
 
     def _validate_config(self) -> None:
         """Validate required fields in configuration."""
--- a/code_puppy/atomic_io.py
+++ b/code_puppy/atomic_io.py
@@ -1,0 +1,201 @@
+"""Generic, cross-process-safe primitives for user-editable config files.
+
+Extracted from :mod:`code_puppy.config_file` (the PUP-605 config-corruption
+fix) so every corner of the config surface -- not just the INI file -- gets
+the same three guarantees for free:
+
+* **bounded reads** -- a pathological/oversized file can never balloon
+  memory the way it did in the original ``MemoryError`` field report;
+* **one cross-process lock per path** -- concurrent readers/writers on the
+  same file (two terminals, a wizard + a slash-command) serialize instead
+  of racing; and
+* **atomic writes** -- same-directory temp file + ``fsync`` + ``os.replace``,
+  so a crash mid-write can never leave a half-written file behind.
+
+:mod:`code_puppy.config_file` (INI) and :mod:`code_puppy.atomic_json` (JSON)
+both build their format-specific ``load_x``/``mutate_x`` helpers on top of
+these primitives. New config-file-shaped state should use one of those two
+modules rather than hand-rolling ``open()``/``json.load()`` again.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+import os
+import tempfile
+import time
+import uuid
+from collections.abc import Iterator
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_MAX_BYTES = 10 * 1024 * 1024
+DEFAULT_LOCK_TIMEOUT_SECONDS = 30.0
+_LOCK_POLL_SECONDS = 0.05
+
+try:  # POSIX
+    import fcntl
+except ImportError:  # pragma: no cover - Windows
+    fcntl = None  # type: ignore[assignment]
+
+try:  # Windows
+    import msvcrt
+except ImportError:  # pragma: no cover - POSIX
+    msvcrt = None  # type: ignore[assignment]
+
+
+class ContentTooLarge(Exception):
+    """A file exceeded the caller's byte budget, before or during a read."""
+
+
+class LockTimeout(TimeoutError):
+    """Another process held the path lock beyond the bounded wait."""
+
+
+def _try_lock(fd: int) -> bool:
+    if fcntl is not None:
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            return True
+        except BlockingIOError:
+            return False
+    if msvcrt is not None:
+        try:
+            os.lseek(fd, 0, os.SEEK_SET)
+            msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+            return True
+        except OSError:
+            return False
+    return True
+
+
+def _unlock(fd: int) -> None:
+    if fcntl is not None:
+        fcntl.flock(fd, fcntl.LOCK_UN)
+    elif msvcrt is not None:
+        os.lseek(fd, 0, os.SEEK_SET)
+        msvcrt.locking(fd, msvcrt.LK_UNLCK, 1)
+
+
+@contextlib.contextmanager
+def path_lock(
+    path: str, timeout: float = DEFAULT_LOCK_TIMEOUT_SECONDS
+) -> Iterator[None]:
+    """Cross-process advisory lock scoped to a ``{path}.lock`` sidecar file.
+
+    Uses ``fcntl.flock`` on POSIX and ``msvcrt.locking`` on Windows. Raises
+    :class:`LockTimeout` rather than blocking forever if another process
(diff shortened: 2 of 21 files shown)


thomwebb:
Ran an adversarial self-review pass on this before merge and it caught a real bug:

**Blocker found:** `add_bedrock_models_to_config`, `remove_bedrock_models_from_config`, `add_foundry_models_to_config`, `add_discovered_models_to_config`, and `remove_foundry_models_from_config` all split the read and write across two separate, unlocked calls (`load_extra_models()` ... `save_extra_models(models)`). A concurrent writer to the same `extra_models.json` (ollama-setup, add-model-menu, or the sibling bedrock/foundry plugin) could have its update silently overwritten -- exactly the class of bug this PR exists to eliminate. `ollama_setup`'s `_register_model` already did this correctly; the other two plugins didn't.

**Fixed** in the latest commit: all 5 functions now build their new/removed entries locally and commit via one locked `atomic_json.mutate_json()` transaction. Added real-thread concurrency tests in both plugins' test files that fail deterministically against the old code and pass against the fix (verified by reverting and re-running).

Also tightened two secondary nits: `atomic_json.load_json`/`mutate_json` now distinguish a missing file from one containing literal JSON `null` via a proper sentinel (previously both collapsed to the caller's `default`), and `azure_foundry`'s exception handling is narrowed to match `aws_bedrock`'s.

Full suite: 7182 passed. Ready for review.

WSxDemise (approved):

AndrewTilson:
*(Edited: an emoji/unicode filter on my side mangled several sentences in the first version of this comment. Reposted below in clean ASCII. Apologies for the noise.)*

Thanks for turning this around so fast, and for taking the `atomic_io` / `atomic_json` layering suggestion further than I'd sketched it. A few things worth calling out as genuinely well done before the findings:

- **The JSON-vs-INI asymmetry is the right call and well-argued.** Surfacing `JsonFileCorrupt` instead of auto-quarantining is correct: INI is Code-Puppy-owned and safe to reset, but a hand-edited `mcp_servers.json` that fails to parse is far more likely a typo the user just made than bitrot, and silently resetting it to `{}` would be a worse outcome than the crash. The module docstring explains it properly.
- **The `_Missing` sentinel.** Using `None` there would have conflated "file absent" with "file contains literal `null`". Easy to get wrong, and `TestNullIsNotConflatedWithMissing` pins it.
- **`upsert_mcp_server(..., replace_name=...)`** keeping rename-in-place as one locked transaction rather than two writes.
- **Self-catching the lost-update race in 744c22fa.** That read/write split is exactly the bug this kind of migration tends to introduce.
- `config_file.py` shrinking to a thin adapter with no behaviour drift, still guarded by the #757 suite.

Two gaps though, and the first one matters.

---

### HIGH: `load_mcp_server_configs()` is still an unbounded read, and it is on the startup path

`code_puppy/config.py:601`:

```python
with open(MCP_SERVERS_FILE, "r", encoding="utf-8") as f:
    configs.update(_parse_mcp_servers_mapping(f.read()))
```

Reached at startup via `MCPManager.__init__` -> `sync_from_config()` (`mcp_/manager.py:146,167`) -> `load_mcp_server_configs()`, and again from `agents/_builder.py:306`.

This is the PUP-605 crash shape, unfixed, on a startup path, for the very file that `mcp_servers_store.py`'s own docstring describes as able to "balloon `json.load` the same way `configparser` ballooned in the original PUP-605 crash". The PR correctly bounded the other startup read (`json_agent.py`), so this one looks like an oversight rather than a deliberate carve-out.

One note on the existing `except Exception` at line 602: it would catch `MemoryError` (it is an `Exception` subclass), but only after the allocation already happened, and `emit_error` on a memory-exhausted interpreter is not a happy path. Bounding before the read is the actual fix; catching afterwards is not.

```python
from code_puppy import atomic_io
raw = atomic_io.read_bounded_bytes(MCP_SERVERS_FILE)
configs.update(_parse_mcp_servers_mapping(raw.decode("utf-8")))
```

---

### MEDIUM-HIGH: there is a fifth `mcp_servers.json` writer, and leaving it unlocked degrades the Tier 1 fix

My #757 comment listed four writers. I missed this one too: **`code_puppy/mcp_/config_wizard.py:506-521`**, inside `run_add_wizard()`, which is public API (exported from `mcp_/__init__.py:9,57`).

```python
if os.path.exists(MCP_SERVERS_FILE):
    with open(MCP_SERVERS_FILE, "r") as f:
        data = json.load(f)                       # unbounded
...
temp_path = Path(MCP_SERVERS_FILE).with_suffix(".tmp")
with open(temp_path, "w", encoding="utf-8") as f:
    json.dump(data, f, indent=2)
temp_path.replace(MCP_SERVERS_FILE)               # no lock, no fsync
```

The locking issue is the real problem here: **a lock only buys mutual exclusion if every writer participates.** Now that the other four serialise on `{path}.lock` and this one does not, this writer can read-modify-write straight through the middle of another writer's locked transaction and clobber it. Before this PR all five raced equally; after it, four are safe and the fifth silently voids their guarantee. That is arguably a worse failure mode than before, because the other four now look safe.

Two smaller problems in the same block:

- `Path(MCP_SERVERS_FILE).with_suffix(".tmp")` yields a single fixed `mcp_servers.tmp`, so two concurrent wizards collide on the same temp path. This is exactly what `mkstemp` in `atomic_write_bytes` avoids.
- No `fsync` before the replace, so a crash can still leave an empty or partial file.

Collapses to the same two-liner the other four now use:

```python
from code_puppy.command_line.mcp.mcp_servers_store import upsert_mcp_server
upsert_mcp_server(config.name, config.config)
```

---

### Remaining unbounded reads (non-blocking)

- `command_line/mcp/edit_command.py:108` - read-only, interactive, not startup. Consistency fix.
- `mcp_/project_config.py:256` and `command_line/mcp/trust_command.py:178` - project-level `.code_puppy/mcp_servers.json`. Mitigated by the trust gate at `project_config.py:247` (untrusted repos return before the read), but still unbounded for trusted ones, and this is repo-supplied input, i.e. whatever repo you `cd` into. These are the two I would most want bounded after the blocker.
- `mcp_/project_config.py:144`, `plugins/trust.py:50` - trust store, Code-Puppy-owned, low risk.

### Nits

1. **`stat.S_IMODE` carried forward from my #757 review.** `atomic_io.atomic_write_bytes` still does `os.chmod(temp_path, mode)` where `mode = os.stat(target).st_mode` includes `S_IFREG` type bits. It works (chmod masks them), but now that this is a shared primitive rather than one call site, `stat.S_IMODE(mode)` is worth the two characters.
2. **`save_extra_models()` is dead code** in both `aws_bedrock` and `azure_foundry` - no production callers, only definitions and tests. It got hardened with `path_lock` anyway. Either drop it or note in the docstring that it is retained as plugin-facing API.
3. **`path_lock` is not reentrant.** `flock` is per-open-file-description, so calling `mutate_json(p)` from inside a `path_lock(p)` block self-deadlocks for the full 30s and then raises `LockTimeout`. Nothing does this today (I checked all nine `mutate_json` call sites), but one sentence in the docstring would keep it that way: "not reentrant - do not call `mutate_*` on the same path from within a held lock".

---

The architecture here is right and the extraction is clean. The two findings are both small fixes on top of solid work. Happy to re-review once the `config.py:601` bound and the `config_wizard.py` writer are in. Nice catch on your own race in the second commit.

PR #764, opened 2026-08-13
fix: recover broken SSE streams without bypassing rate-limit backoff

s0s01o8:
## Root cause\n\nThe \`Turn interrupted mid-stream\` banner masked two distinct transient failures:\n\n- malformed SSE/protocol payloads (for example, \`extra JSON data in SSE payload\`)\n- provider rate limiting / upstream 5xx responses\n\nThe runtime previously only had a generic retry classification. Retrying malformed SSE through the same streaming transport repeats the broken protocol, while immediately issuing a non-streaming retry for rate limits adds avoidable pressure during the hot window.\n\n## Change\n\n- Add a narrow \`should_fallback_to_non_streaming\` classifier for corrupted/early-ended SSE and protocol errors.\n- Switch to non-streaming once only for those transport failures.\n- Preserve normal retry-profile backoff for rate limits and provider 5xx responses.\n- Keep streaming disabled for follow-up calls in a recovered turn.\n\n## Validation\n\n- \`pytest tests/agents/test_streaming_retry.py -q\` -> 84 passed\n- \`ruff check code_puppy/agents/_runtime.py tests/agents/test_streaming_retry.py\` -> passed\n\nThe PR intentionally contains one commit and excludes unrelated local work.


s0s01o8:
Follow-up commit `671ab89e` removes the misleading terminal banner entirely. Retry output now reports the classified real cause and wait time: provider rate limits show the provider message, while malformed SSE reports a streaming protocol error. Added regression tests ensuring the old \`Turn interrupted mid-stream\` / \`last completed step\` wording cannot return. Focused suite: 86 passed; Ruff passed.

PR #768, opened 2026-08-15
feat!: report subagent token usage per call and add Model Judge

AndrewTilson:
Sub-agent runs invoked through `invoke_agent_with_model` now report token
usage that can actually be turned into a cost.

image

## Diff breakdown

`+1243 / -234` looks big; most of it isn't logic.

| Category | Lines | % |
|---|---:|---:|
| Implementation | 153 | 12% |
| Tests | 721 | 58% |
| Agent prompt text (English, not code) | 128 | 10% |
| Docstrings & comments | 27 | 2% |
| Blank | 214 | 17% |
| **Total** | **1243** | |

The behavioural change is ~150 lines across four functions. Tests outweigh
implementation roughly 5:1, which is deliberate: every fix here is a
"the number was silently wrong" bug, and each is pinned by a test that fails
when the fix is reverted. The prompt text is the `model-judge` persona -- prose
that happens to live in a `.py` file.

## What changed

**Per-field availability.** A provider that reported one counter but omitted
another had the omission exposed as a measured `0`. Availability is now decided
per field, so an unreported counter stays `None`. Input is sourced from
attributes only: Anthropic's `details["input_tokens"]` is base-only and would
double-subtract the cache buckets.

**`total_tokens` removed.** Input, cache read, cache write and output are billed
at different rates, so a blended sum cannot be converted into a cost. The four
buckets are reported separately and never aggregated.

> **Breaking, two ways.** `total_tokens` shipped on `main` and is gone -- no
> in-tree consumers were found; sum the buckets yourself if you only need a
> rough size. Separately, a counter the provider didn't report is now `None`
> where it used to be `0`, so arithmetic that assumed a number will raise
> instead of silently under-reporting. That second one is the point of the
> change, but it is the wider break surface.

**`per_request_usage` — one entry per model call.** Even correct run totals
cannot be priced when the rate depends on a single call's context length:
`3x60k` and `1x180k` both sum to 180,000 and bill very differently. Upstream
agrees, warning on `RequestUsage.__add__` that it "CANNOT be used to sum
multiple requests without breaking some pricing calculations". Each entry
carries its own buckets plus the `model_name` that served it, which also stops
a run that switched models mid-flight from being summed across two price sheets.

**`final_context_tokens`.** How much context was live when the run finished:
the last call's raw combined input plus its output. Deliberately not built from
the billing buckets — those subtract cached tokens out so the categories don't
overlap, but cached tokens still occupy the window, and reusing them undercounts
occupancy roughly threefold.

## Two traps worth flagging for review

`RunUsage.total_tokens` is an inherited `@property`, not a dataclass field, so
it is invisible to `dataclasses.fields()`. Any adapter that trusted it reported
a confident total alongside an unmeasured output. Totals are always derived now.

`all_messages()` includes messages from *older runs*. Extracting per-request
usage from it made a resumed session re-report every earlier call as freshly
billable — reproduced at 3.8x overstatement by turn 4, compounding without
bound. Extraction reads `new_messages()`; `all_messages()` is still used for
session persistence, which needs it.

## Provider caveat

Not every provider reports `cache_creation_input_tokens`, so it stays `None`
there. `None` means "not reported", never "zero" — the gap stays visible rather
than being silently priced as free.

Whether an unreported write is also *unbilled* depends on the price sheet, not
the provider name: no OpenAI model prices cache writes, and nor does Gemini
served directly by Google — but the same Gemini via OpenRouter does. The judge
prompt is written to check the rate for the exact model+provider rather than
infer it from the family name.

## Model Judge agent

Also adds a `model-judge` builtin: runs one prompt across several models and
reports a side-by-side comparison with a ranked verdict. It is bundled here
rather than sent separately because it consumes the fields above -- it reads
`per_request_usage` so cost is computed per call, and would reference
non-existent fields on today's `main`.

Its persona encodes the same rules as the schema: buckets are never summed
into a blended total, and an unreported figure stays `None` rather than being
shown as a measured zero.

It also documents how to recover cache writes on models that never report them.
OpenAI and Gemini leave `cache_creation_input_tokens` empty -- and for them that
is not a hidden cost: their price sheets carry no cache-write rate at all
(0 of 71 OpenAI models; every Google entry with one is Claude-on-Vertex, not a
`gemini-*`). The volume is still a useful behavioural signal, so the prompt
teaches how to recover it while forbidding pricing it for those providers.
With prefix caching, what one call writes is what the
next call reads -- and because reads are cumulative, the estimate is the
*difference* between consecutive reads, not the next call's raw read:

    estimated_write(call i) = cache_read(call i+1) - cache_read(call i)

Verified against a run whose true writes were 1000/300/250: the delta form
recovers them exactly, while the raw-read form yields 1000/1300/1550. The
estimate must be labelled as such, cannot be computed for the final call, and
is abandoned when compaction or a model switch breaks the cached prefix.

A test pins every metric named in the prompt to a field that really exists on
`AgentInvokeWithModelOutput` -- the persona had already fallen behind the schema
once, and telling the judge to read a missing field produces confident nonsense.

## Tests

7262 passing, including a 4-turn resumed-session regression, real `RunUsage` /
`RequestUsage` objects rather than stubs, and a `-W error::DeprecationWarning`
run. Each fix was mutation-tested: reverting it fails the suite.

Leave-one-out mutation testing was also used to prune the suite -- a test was
removed only after confirming it caught nothing another test already caught.
That exercise found a test asserting an alias it never actually exercised (it
set both an attribute and a detail key, and the attribute always won), plus
three detail-key aliases that appear nowhere in pydantic-ai or genai-prices.
The dead aliases are gone, and every alias that remains now fails a test when
deleted.


AndrewTilson:
Self-review completed. Removed the message-length token estimator because billing metrics must remain provider-reported. Also fixed cache undercounting in mixed-provider/multi-request runs by preferring pydantic-ai's normalized aggregate cache fields over individual detail aliases. Added regression coverage for provider-zero usage, fully cached input, explicit zero cache details, and normalized aggregate precedence.

AndrewTilson:
Thanks for tightening the usage accounting. I found two semantics that should be addressed before merge:

1. **Handle availability per token field, not globally** (`code_puppy/tools/subagent_usage_metrics.py`, `_safe_usage_metrics`).

   `provider_reported_tokens` becomes true when *any* token field is positive. Once that happens, zero-valued `input_tokens`, `output_tokens`, and `total_tokens` are all preserved. However, pydantic-ai's `RunUsage` defaults each counter independently to `0`, so a provider that supplies input usage but omits output usage can expose `output_tokens=0` as if the provider explicitly reported it.

   Please either track availability per field or establish/document an adapter invariant that input and output are always reported together. Add a regression test for partially populated usage, such as positive input with an omitted/default-zero output field.

2. **Clarify or change the `total_tokens` contract** (`code_puppy/tools/subagent_usage_metrics.py`, `_extract_usage_metrics`).

   When `usage.total_tokens` is absent, the implementation computes it by summing the token buckets. In the pinned pydantic-ai version, `RunUsage` does not define `total_tokens`, so this is normally a derived value rather than a provider-supplied field. That conflicts with the PR summary and `AgentInvokeWithModelOutput` documentation saying only provider-reported counts are used and unavailable values remain `None`.

   Please choose one explicit contract:
   - provider-only: leave `total_tokens=None` when no provider/adapter total is exposed; or
   - exact derived total: retain the sum but document `total_tokens` as derived and adjust the PR's provider-reported-only wording.

The normalized cache aggregate preference and focused regression coverage otherwise look good, and current GitHub CI checks are passing.

mpfaffenberger (approved):

PR #772, opened 2026-08-16
feat: terminal tool-result processing and per-agent spill hooks

thomwebb:
Adds the app half of the coordinated spill feature. Counterpart: mpfaffenberger/code_puppy_core_plugins#7 (merge that first — see *Release ordering*).

## What

**New tool-result pipeline** (pydantic-ai 2.x public APIs — `ToolManager.validate_tool_call` / `execute_tool_call` / `validate_output_tool_call`, plus a call-aware `CallToolsNode._handle_tool_calls` seam):

- `pre_tool_call` → blocking policy → hook-context envelope (composed off-loop in a worker) → `post_tool_call` → new `final_tool_result` phase → reserved terminal boundary.
- **Fail-closed blocking**: a latched block decision survives diagnostic failures (`emit_warning` raising, hostile `blocked`/reason values never evaluate callback-owned truthiness); blocked calls balance already-started lifecycle callbacks via a structured blocked result, increment `RunUsage.tool_calls`, and update `succeeded_tools` only under `wrap_validation_errors`.
- **Faithful delegation**: deferred (`ApprovalRequired`/`CallDeferred`), args-invalid, unknown-tool, and external calls bypass hooks entirely and surface pydantic-ai's exact exception.
- **History integrity**: hooks mutate an isolated deep copy; changes publish as detached execution + model-history snapshots (JSON-mode via `pydantic_core`). No-ops preserve provider bytes byte-for-byte (prompt-cache friendly), `args=None` stays `None`, unserializable mutations block instead of diverging history, and partial publication rolls back both views. Annotation-level custom serializers (e.g. `Annotated[range, PlainSerializer(...)]`) execute unharmed.
- **Claude Code `cp_` normalization** is per-call provider-aware (`ToolCallPart.provider_name`, run-provider fallback, exact-tool lookup first) instead of process-global model state — correct under sub-agents and concurrent runs. Output tools get the same normalization and malformed-JSON repair (partial streaming validation untouched); JSON repair runs before policy hooks so repaired arguments can be inspected, mutated, or blocked.
- Per-seam idempotent patching (a partially perturbed patch set is repaired without double-wrapping), and JSON-repair/callback markers survive wrappers via `functools.wraps`.

**Per-agent execution context**: async-safe `ContextVar` scoping of the executing agent, so plugins (like core's spill) can resolve per-agent config correctly across concurrent main agents and sub-agents.

**Docs**: README documents `tools_config.spill` per-agent controls, eligibility, budget semantics, the 128-field cap, and the exact single-key `{"error": ...}` exception.

## CI / validation

- Coordinated core-plugins suite against this commit: `2,004 passed, 2 skipped`
- Focused app set (callbacks/patches/subagents/idempotence/custom-serializer/rollback): `176 passed` / `225 passed` in independent re-runs
- Whole-repo Ruff format/check clean; patched modules split under 600 lines (`patch_support.py`, `tool_call_patches.py`, `pydantic_patches.py` re-exports)
- Survived ten rounds of adversarial review (final round PASS_NO_NITS), including independent reproduction of the custom-serializer, partial-patch, hostile-diagnostics, and history-divergence cases.

Note: this branch was cut from the staging fork's main, so it carries three fork-bookkeeping commits (two `chore: bump version` + the fork's merge commit) alongside the feature commits. Intentionally not rebased — core-plugins CI pins this branch by immutable SHA (`5525ad5d`), and history rewrites would break that. Happy to produce a rebased variant if maintainers prefer, with the pin updated in lockstep.

## Release ordering (stacked)

PyPI currently ships core-plugins `0.0.8`, which predates the coordinated phases. This PR's `pyproject.toml`/`uv.lock` therefore still resolve `>=0.0.8` **honestly**; the README gates the new controls on `>=0.0.9`. Once #7 merges (its publish workflow auto-bumps `0.0.8 → 0.0.9`), this PR raises the floor to `>=0.0.9` and regenerates the lock before merging. No lockfile fiction in the meantime.

## Also included: split-out test fixes

- `tests/test_http_utils_full_coverage.py` — the port-occupancy test fails deterministically on macOS on `main` (`SO_REUSEADDR` double-bind); fixed here and also proposed standalone in #773.
- `tests/integration/test_dbos_enabled.py` — hard-failed without the optional `durable` extra; now skips via `find_spec` (assertions stay hard when DBOS is installed). Also in #773.

Once #773 merges, both hunks drop from this PR's diff (identical content, no conflict). Kept only here: removal of a stale non-strict `xfail` on the Alt+M binding test (passes on this branch's pinned prompt_toolkit; version-dependent, so not proposed standalone), and a few import-order-only changes forced by the repo's isort pre-commit hook during merges.


thomwebb:
Closing to restart with a much smaller scope: only the per-agent execution context needed for core-plugins per-agent spill controls. Test fixes split out of this PR live in #773 and stay open. The larger tool-result work remains on the `feat/spill-basemodel-results` branch for future, separately-scoped PRs.

PR #790, opened 2026-08-18
Fix interrupted parallel tool-call pruning + assorted robustness fixes

### LINKED ISSUE #782: prune_interrupted_tool_calls deletes whole messages, erasing completed parallel tool calls on cancellation

**Where:** `code_puppy/agents/_history.py` → `prune_interrupted_tool_calls`

**What.** The pruner computes `tool_call_ids.symmetric_difference(tool_return_ids)` and then drops every *message* containing any part with a mismatched id. A single ModelResponse routinely carries several ToolCallParts plus a TextPart, so one unanswered call condemns the whole assistant turn including successful sibling calls. It runs at the top of `run_agent_task`, in its `finally`, and inside `filter_huge_messages` during compaction — and `_checkpoint_cancelled_history` explicitly documents relying on the finally-prune to 'normalize any dangling tool calls afterwards', so the partial state it works to recover from the RunCancelled snapshot is destroyed immediately after.

**Impact.** Cancel during a parallel edit of three files: two edits succeeded and their returns are in history, the third was interrupted. After the prune the conversation contains no record that the agent said what it was doing or that it edited a.py and b.py at all — the surviving returns are then orphaned and pydantic-ai's own `_drop_orphaned_tool_results` deletes those too. On the next turn the model re-plans from scratch and can redo or double-apply the completed edits. Same erasure via `filter_huge_messages` when one oversized return is dropped from a multi-call step.

**Suggested fix.** Operate at part granularity, or delete the pass entirely and let pydantic-ai handle it. For a dangling call, append a synthesized ToolReturnPart ('interrupted by user') to the following ModelRequest; for an orphaned return, drop only that part. Keep `has_pending_tool_calls` as-is — it is read-only and correct.

dsfaccini:
> This pull request was posted by Grok Build using grok-4.6 on behalf of David.

Fixes #782.

Primary fix: `prune_interrupted_tool_calls` now repairs unmatched tool calls at **part** granularity. A cancelled call in a parallel batch is closed with a synthesized interrupted return; completed sibling calls and their returns stay in history. Previously one unanswered id deleted the whole assistant turn (and then pydantic-ai dropped the surviving orphan returns), so cancel-during-parallel-edit lost the work that had already finished.

Also included — small, independently-tested robustness fixes:
- **grep**: honor `-A/-B/-C` context events, clustered short flags (`-iw`, `-C3`), and an explicit `-t/--type` without the tool's default `--type=all` defeating it; error if a value-taking flag is last so the target directory is not eaten.
- **hooks**: interpolate `$var` / `${var}` as literal text in the quoting context the template established, so metacharacters in values stay inert in the hook's own shell.
- **child processes and hooks**: remove only the agent's own provider credentials from the inherited environment; user tokens (`GITHUB_TOKEN`, AWS keys, etc.) still reach `gh` / `git` / `npm`.
- **project `.env`**: import known API-key names; Azure's endpoint hydrates from `puppy.cfg` only.
- **legacy sessions**: treat `builtins.globals` / `builtins.locals` as surrogates, same as the existing `vars` / `getattr` block.
- **project plugins**: load trusted plugins through a source-only finder so reviewed source is what executes.
- **compaction**: shape a summarization slice without fabricating assistant/user filler turns (leading-user repair, lossless same-role merge, trailing user folded into the instruction).

All changes ship with tests; the full suite passes.


thomwebb (changes requested):
Ran the tests, then went hunting for holes since a fair chunk of this is agentic. Found two concrete, reproducible bugs plus a couple of things I'd like explicit confirmation on before this merges.

## Blocking: grep flag-cluster expansion corrupts value-taking flags' arguments

<details>
<summary>details</summary>

`_expand_short_flag_cluster()` runs on every token *before* the code knows which tokens are flags vs. values for a preceding value-taking flag. That breaks the tool's own documented escape hatch for patterns that start with `-`:

```python
>>> _build_grep_args("-e -si")   # intent: search for the literal pattern "-si"
(['-e', '-s', '-i'], None)       # "-e" takes "-s" as its value; "-i" becomes a real flag
>>> _build_grep_args("-t -tw")
(['-t', '-t', 'w'], None)
```

Same bug independently in `_build_backend_matcher` (`-e -in"` errors with `"flag '-n' is not supported"` instead of treating `-in` as the literal pattern). The tool's docstring explicitly tells the model to use `-e '-pattern'` for this case, so this is a regression on a documented, encouraged code path — e.g. searching a script for `grep -in` usage.

Suggest interleaving cluster-expansion into the index-walking loop (only expand a token when you're about to treat it *as* a flag) rather than pre-expanding the whole token list.

</details>

## Blocking: `stringify_part` (a hashing/token-estimation helper) reused as LLM-facing prompt text

<details>
<summary>details</summary>

In `_compaction.py`, `_request_text()` falls back to `stringify_part()` when a trailing user turn's content isn't a plain string (e.g. multi-part content with an image). Reproduced:

```python
UserPromptPart(content=["check this screenshot", BinaryContent(...)])
→ _request_text(...) → 'user-prompt|content=check this screenshot|BinaryContent=b9e043d71a5f5a32'
```

That hash-digest string gets folded directly into the summarization prompt. It's also a regression vs. current `main`: today the trailing user turn (images included) stays intact in `message_history` sent to the summarizer; this PR peels it off and lossily flattens it, so the actual image content never reaches the summarizer.

Suggest either keeping multi-part trailing requests in `pruned` (same treatment the tool-return branch already gets) or writing a real prompt-rendering helper instead of repurposing the hash helper.

</details>

## Please confirm intentional: credential-scrub model got narrower

<details>
<summary>details</summary>

`_child_process_env()` used to strip *any* env var ending in `_API_KEY`/`_TOKEN`/`_SECRET` (broad denylist — this is what broke `GITHUB_TOKEN`/`NPM_TOKEN` for `gh`/`npm`, understood why that needed fixing). The new `environment_without_credentials()` only strips known provider creds (hardcoded list + whatever's referenced in the user's current `models.json`). That's a reasonable trade to unbreak `gh`/`npm`, but it does mean an unrelated secret-shaped env var (`STRIPE_API_KEY`, an internal `SLACK_TOKEN`, etc.) that used to get defensively stripped from every spawned child process now passes straight through. Given code-puppy already runs arbitrary LLM-directed shell commands, that's a real widening of the exfiltration surface — want to make sure this was a deliberate, discussed trade and not a side effect of the refactor.

Related: `credential_env_var_names()` is `functools.lru_cache(maxsize=1)`, so a credential added to a custom model mid-session (e.g. via the config wizard) won't be scrubbed from child/hook environments until restart. `_load_merged_model_config()` already re-reads from disk each call, so recomputing per-spawn is probably cheap enough to drop the permanent cache — but flagging since this is a security-relevant path.

</details>

## Worth a test either way: builtin/native tool calls are never repaired

<details>
<summary>details</summary>

`prune_interrupted_tool_calls` deliberately skips synthesizing a return for a dangling `builtin-tool-call`, reasoning that "results arrive from the provider in a later turn." If call+result always land atomically in the same `ModelResponse`, that's fine. But if a run is interrupted mid a native/builtin tool call and never resumes with that provider, `has_pending_tool_calls` will see that ID as pending forever — which is the exact "one orphaned call defers summarization forever" bug class this PR sets out to fix, just relocated to the builtin lane. Would like a regression test proving whether this is actually reachable before merge, one way or the other.

</details>

## What's solid

<details>
<summary>details</summary>

- The Azure `AZURE_OPENAI_ENDPOINT` / project-`.env` isolation is well-scoped (endpoint = redirect target, not a credential, correctly excluded from untrusted repo `.env` loading).
- The hook `$var`/`${var}` interpolation rewrite is a real fix (old code re-quoted a value already inside the author's own single quotes) and is tested by actually executing the substituted string through `/bin/sh` and asserting no injection — good test design for this bug class.
- Replacing the global `sys.pycache_prefix` mutation (a genuine cross-thread hazard) with a scoped meta-path finder is a solid improvement, tested against a deliberately-planted unchecked-hash `.pyc`.
- Full suite (7390 tests) passes; all new tests pass.

</details>

## Process ask, not specific to this PR

<details>
<summary>details</summary>

Could we try to keep future PRs scoped to one fix rather than "primary fix + assorted robustness fixes" bundling several unrelated subsystems (grep, hooks, credentials, plugins, compaction, unpickler all touched here alongside the actual #782 fix)? Not asking to split this one up — just a heads-up for next time. It makes review slower and means an unrelated regression (like the grep one above) blocks an otherwise-solid primary fix from merging on its own timeline.

</details>

dsfaccini:
> This comment was posted by Claude Code using claude-fable-5 on behalf of David. David has not yet reviewed this reply.

Point-by-point:

<details>
<summary><b>grep flag-cluster expansion</b> — already fixed in <code>6e99525d</code></summary>

Real at the commit you reviewed (`2d5c00d4`); fixed in `6e99525d`, which landed right after your review. Cluster expansion now runs inside the token walk, only for a token in flag position, so a value token is forwarded verbatim: `-e -si` → `['-e', '-si']`, `-t -tw` → `['-t', '-tw']`, `_build_backend_matcher("-e -in")` → pattern `-in`. Guards: `test_build_grep_args_value_not_shredded`, `test_build_grep_args_value_keeps_cluster_literal`, `test_build_backend_matcher_value_not_shredded` in `tests/tools/test_grep_flag_validation.py`.

</details>

<details>
<summary><b><code>stringify_part</code> as prompt text</b> — confirmed, fixed in <code>889248ca</code></summary>

A trailing request that isn't pure plain-text now takes the same keep-in-history branch as tool returns (the real message stays in `pruned`, the one-token break restores role alternation), so the image reaches the summarizer intact and no digest string enters the prompt. `_request_text` lost the `stringify_part` fallback entirely. Regression tests in `tests/agents/test_compaction_summarization_ordering.py` assert the image arrives and the prompt stays bare instructions — both fail against the old code with exactly the digest string you reproduced.

</details>

<details>
<summary><b>credential-scrub narrowing</b> — deliberate</summary>

Stated in the PR body ("user tokens (`GITHUB_TOKEN`, AWS keys, etc.) still reach `gh` / `git` / `npm`") and in the module docstrings. Worth adding why the suffix rule wasn't kept as a baseline: it was never a coherent secrets filter — it broke real tooling (`GITHUB_TOKEN`, `NPM_TOKEN`, `AWS_SESSION_TOKEN`) while passing comparably sensitive vars (`AWS_SECRET_ACCESS_KEY`, `GOOGLE_APPLICATION_CREDENTIALS`, `DATABASE_URL`, `SSH_*`). The new model scopes scrubbing to the credentials code_puppy itself manages (18 well-known provider keys + every catalog-referenced `api_key` var) and also newly scrubs the hook lane — on `main`, hooks inherit every provider key unscrubbed. The wider exfiltration surface for unrelated secrets is real; a coherent general filter would be allowlist-shaped ("only these names reach children"), which is its own design discussion — happy to open an issue for that if you want the defense-in-depth.

</details>

<details>
<summary><b><code>lru_cache</code> staleness</b> — cache dropped in <code>6c1b7595</code></summary>

Took your suggestion: the scrub set is recomputed per spawn (`_load_merged_model_config` already re-reads disk each call; the cost is negligible next to a process spawn). The two in-app writers already invalidated (`save_credential`, the `/add_model` writer); recomputing closes the remaining hand-edit path and deletes the invalidation coupling.

</details>

<details>
<summary><b>builtin/native tool calls</b> — reachable but inert here; the regression test already exists</summary>

A dangling builtin call in persisted history *is* reachable (Ctrl-C mid Anthropic server-tool block; Anthropic can also split a `tool_search` call/result across turns with no interruption at all). That's exactly why `6e99525d` excludes builtin parts from the pending/pairing sets: on this branch the dangling call is inert — `has_pending_tool_calls` ignores it and `prune_interrupted_tool_calls` is an identity. The regression test exists: `test_builtin_tool_call_left_alone` in `tests/agents/test_history_tool_call_repair.py` asserts both properties. The forever-defer you describe is `main`'s current behavior, which this PR removes. Residual: what happens when a persisted orphaned native call is re-sent lives in pydantic-ai's provider serializers (it drops orphaned `tool_search` calls; the direct API tolerates the rest), upstream of this PR.

</details>

<details>
<summary><b>process ask</b> — fair, taken</summary>

Future PRs stay scoped to one fix.

</details>

Also merged `main` into the branch (one conflict in `add_model_menu.py`: i18n strings vs. the since-removed `cache_clear` block). Full suite passes.

*(I also collapsed the review's sections into `<details>` blocks to keep the thread compact — content untouched.)*

thomwebb (approved):
Re-verified against `6c1b7595` (didn't just take the write-up's word for it — re-ran my exact repros plus a few new adversarial cases):

- **grep flag-cluster shredding** — fixed. `_build_grep_args("-e -si")` now correctly returns `['-e', '-si']`, `_build_grep_args("-t -tw")` returns `['-t', '-tw']`, and `_build_backend_matcher("-e -in")` compiles pattern `-in`. Also threw a few extra cases at it that weren't in the new tests (`-g *.py -e -si`, `--type py -e -tpy`, `-e -e`) — all correct, values after any value-taking flag stay literal even when they look like a legit cluster themselves.
- **`stringify_part` in compaction prompts** — fixed. Confirmed the `stringify_part` fallback is gone from `_request_text` entirely; a trailing multi-modal turn now stays in real message history (`_is_plain_text_request` gate) instead of being flattened into a hash-digest string.
- **`lru_cache` staleness** — fixed, confirmed the decorator's gone and the scrub set recomputes per call.
- **builtin/native tool-call dangling** — confirmed `has_pending_tool_calls` returns `False` for a dangling native call now, so it's inert rather than deferring summarization forever.
- **credential-scrub narrowing** — appreciate the writeup on why the old suffix rule wasn't a coherent filter to begin with. Agreed this is a separate design discussion (allowlist-shaped defense-in-depth) rather than a blocker on this PR — please do open that issue if you get a chance, happy to weigh in there.

Full suite passes on my end too (7458 passed after the fixup commits, up from 7390 at the commit I originally reviewed). Approving.

PR #797, opened 2026-08-19
[experimental, do not merge] Run CI against the pydantic-ai httpx2 commit

dsfaccini:
> This pull request was posted by Claude Code using claude-opus-5 on behalf of David.

## Experimental — not for merge

This PR exists to run CI against an unreleased pydantic-ai commit. It should not be merged; it will be closed once the question below is answered.

## What it changes

`pydantic-ai-slim[openai,anthropic,mcp]` moves from the `==2.31.0` release to a git pin on [`pydantic/pydantic-ai@507c8bf`](https://github.com/pydantic/pydantic-ai/commit/507c8bf87679a1cee11c6d85fa583a781decc6a4) — the merge commit of [pydantic/pydantic-ai#7351](https://github.com/pydantic/pydantic-ai/pull/7351), which switches pydantic-ai-owned HTTP and the OpenAI/Google/Mistral SDK boundaries to `httpx2`.

Lock effects: `httpx2` 2.10.0 enters the tree, `openai` resolves to 3.1.0 (the migration's floor is `openai>=3.0.0`). Legacy `httpx` stays — code_puppy depends on it directly, and the Anthropic SDK boundary is still on it.

## What it is checking

`code_puppy/model_factory.py` hands a caller-owned legacy `httpx.AsyncClient` to the OpenAI-family providers. Under #7351 that path is still accepted but emits a v3 deprecation warning. The run is looking for two things:

1. Does the suite still pass against httpx2-backed clients?
2. What warnings does it emit — deprecation or otherwise?

If CI is green, a follow-up commit turns warnings into errors under pytest to surface anything the passing run hid.

## Result

Baseline run was green. With warnings escalated to errors, 18 tests failed on a single deprecation — four call sites hand a legacy `httpx.AsyncClient` to an OpenAI-compatible provider ([details](https://github.com/mpfaffenberger/code_puppy/pull/797#issuecomment-5336885269)).

Migrating Code Puppy's own boundaries to `httpx2` takes that to 11 ([details](https://github.com/mpfaffenberger/code_puppy/pull/797#issuecomment-5337214631)). The rest live in `code-puppy-core-plugins` and need a companion change there — flipping the shared client factory from this side breaks that package's `httpx.Auth` and `httpx.AsyncByteStream` subclasses.


dsfaccini:
> This comment was posted by Claude Code using claude-opus-5 on behalf of David.

## Results

**Run 1 — httpx2 pin, warnings left alone: green.** All three jobs passed at [`a351763`](https://github.com/mpfaffenberger/code_puppy/commit/a3517632). 7468 passed. So nothing in the suite actually breaks against httpx2-backed pydantic-ai; the tree resolves `httpx2` alongside legacy `httpx`, and `openai` moves to 3.x.

**Run 2 — warnings as errors: 18 failures, one root cause.** Every failure is the same deprecation:

```
PydanticAIDeprecationWarning: `httpx.AsyncClient` support for OpenAI-compatible providers
is deprecated and will be removed in v3; use `httpx2.AsyncClient` instead.
```

Four call sites hand a caller-owned legacy `httpx.AsyncClient` to an OpenAI-compatible provider:

| Call site | Package |
| --- | --- |
| `code_puppy/provider_identity.py` — `AliasedOpenAIProvider` | this repo |
| `code_puppy/model_factory.py` — `CerebrasProvider` construction | this repo |
| `copilot_auth/register_callbacks.py` | `code-puppy-core-plugins` |
| `ollama/register_callbacks.py` | `code-puppy-core-plugins` |

Failures by file: 8 in `tests/plugins/test_copilot_auth_model.py`, 4 in `tests/test_model_factory.py`, 3 in `tests/test_model_factory_basics.py`, 3 in `tests/plugins/test_ollama_model.py`.

One of those is worth a second look independently of httpx2: the three `test_ollama_model.py` failures do not surface the deprecation, they fail on `assert None is not None`. The ollama callback catches broadly around model creation and logs `Failed to create ollama model 'ollama-qwen3': ...`, so an exception raised inside provider construction becomes a silent `None` return rather than a visible error.

## Packaging notes for installing pydantic-ai from git

Two changes were needed beyond swapping the requirement, both unrelated to httpx2 itself:

- `tool.hatch.metadata.allow-direct-references = true` — hatchling refuses a PEP 508 direct reference otherwise, which broke every job at install time.
- an explicit `pydantic-graph` pin at the same commit — a git-built `pydantic-ai-slim` requires the exact sibling dev version (`2.31.2.dev12+507c8bf8`), which is not on PyPI, so plain `pip install .` in the `quality` job could not resolve it.

## Not for merge

This branch stays experimental. The finding to carry forward is the four call sites: they keep working until pydantic-ai v3, and migrating them to `httpx2.AsyncClient` is what removes the deprecation.

dsfaccini:
> This comment was posted by Claude Code using claude-opus-5 on behalf of David.

## httpx2 at Code Puppy's own provider boundaries: 18 failures → 11

[`3be280e`](https://github.com/mpfaffenberger/code_puppy/commit/3be280ed) adds `httpx2` as a direct dependency and a `create_httpx2_async_client` factory in `code_puppy/http_utils.py` alongside the existing one. The retry/backoff behaviour is now a `_RetryingSendMixin` shared by `RetryingAsyncClient` (legacy) and `RetryingHttpx2AsyncClient`, since each stack defines its own connection-error classes.

Migrated to httpx2 — these hand a client to a pydantic-ai OpenAI-compatible provider:

- `model_factory.py` custom-OpenAI path (including the `isinstance` gate that decides whether `http_client` is passed at all — it would have silently dropped the client had it kept checking `httpx.AsyncClient`)
- `model_factory.py` Cerebras path
- `chatgpt_codex_client.py` — used by the `chatgpt_oauth` plugin, which passes it straight to `OpenAIProvider`

Deliberately left on legacy `httpx`, because their consumers require it:

- `mcp_/managed_server.py` — pydantic-ai types `MCPServer.http_client` as `httpx.AsyncClient`, and the stable path is FastMCP 3 / MCP 1 on legacy httpx
- `claude_cache_client.py` — the Anthropic SDK boundary is legacy in the migration matrix
- `model_factory.py` custom-Gemini path — feeds Code Puppy's own `GeminiModel` REST client, not a pydantic-ai provider

## The remaining 11 are in `code-puppy-core-plugins`, not here

8 copilot + 3 ollama failures come from `copilot_auth` and `ollama` calling `code_puppy.http_utils.create_async_client` and passing the result to `OpenAIProvider`.

Flipping that shared factory to httpx2 would fix them in one line and break two things in the plugins package, both verified against `httpx2` 2.10:

1. `copilot_auth/register_callbacks.py` sets `client.auth = _CopilotAuth(...)` where `_CopilotAuth` subclasses `httpx.Auth`. httpx2's `_build_auth` raises `TypeError: Invalid "auth" argument` for an auth object from the other stack.
2. `copilot_auth/reasoning_client.py` assigns `response.stream = _OpaqueCapturingStream(...)`, which subclasses `httpx.AsyncByteStream`. `httpx2.Response.aclose()` isinstance-checks its own `AsyncByteStream` and raises `RuntimeError: Attempted to call an async close on a sync stream`.

So clearing those 11 needs a companion change in [code_puppy_core_plugins](https://github.com/mpfaffenberger/code_puppy_core_plugins) — `_CopilotAuth` on `httpx2.Auth`, `_OpaqueCapturingStream` on `httpx2.AsyncByteStream`, and the plugins switching to `create_httpx2_async_client`. `tests/plugins/test_ollama_model.py` also builds its own `httpx.AsyncClient` in the patch it injects, so it needs updating here at the same time.

None of this is urgent: every remaining call site keeps working until pydantic-ai v3. The failures are only visible because this branch escalates warnings to errors.

mpfaffenberger:
Couple failing tests but not terrible?

dsfaccini:
yeah I'm unsure whether to turn this into a PR PR or scrap it and open a new PR that is just an httpx2 migration

could wait for the other SDKs to wake up as well, though

PR #802, opened 2026-08-19
fix: bound remaining mcp_servers.json reads and close fifth unlocked writer

thomwebb:
Jira: [PUP-610](https://jira.walmart.com/browse/PUP-610)

## Second follow-up to PUP-605, per Andrew Tilson's post-merge review of #761

After #761 merged (extending bounded/atomic/locked I/O to `mcp_servers.json`, `extra_models.json`, `spinners.json`, and agent JSON files), Andrew left a further comment identifying two remaining gaps specific to `mcp_servers.json` -- the exact file this whole line of fixes exists to protect.

### HIGH: `config.py`'s `load_mcp_server_configs()` was still an unbounded read, and it's on the startup path

Reached at every CLI startup via `MCPManager.__init__` -> `sync_from_config()`, and again from `agents/_builder.py`. This is the original PUP-605 `MemoryError` crash shape, unfixed, on a startup path, for the very file `mcp_servers_store.py`'s own docstring already warned could "balloon `json.load` the same way `configparser` ballooned in the original PUP-605 crash." The sibling `JSONAgent` startup read was correctly bounded in #761 -- this one looked like a straightforward oversight.

**Fixed** via `atomic_io.read_bounded_bytes`, matching every other call site.

### MEDIUM-HIGH: a fifth, previously unidentified `mcp_servers.json` writer

`mcp_/config_wizard.py::run_add_wizard()` (public API, exported from `mcp_/__init__.py`) was still doing an unlocked, unbounded read-modify-write with a fixed (collision-prone) temp filename and no `fsync`. #761 correctly collapsed the four writers flagged in #757's original review into one locked `mcp_servers_store.py` module -- but this fifth writer was missed. A lock only buys mutual exclusion if *every* writer participates; this one unlocked writer could silently void the safety guarantee the other four now have, which is arguably worse than before the fix, since the other four now *look* safe.

**Fixed** by migrating to the same shared `mcp_servers_store.upsert_mcp_server()` the other four use.

### Also from the same review comment

- Bounded two lower-priority project-level `mcp_servers.json` reads (`mcp_/project_config.py`, `command_line/mcp/trust_command.py`) -- project config is repo-supplied input (whatever repo you `cd` into), so these were flagged as next-most-wanted after the two items above.
- `atomic_io.atomic_write_bytes` now masks stat mode bits via `stat.S_IMODE` before `chmod`, rather than passing `os.stat()`'s raw `st_mode` (which includes file-type bits) straight through. Harmless today since `chmod` masks it anyway, but worth doing properly now that it's a shared primitive.
- `path_lock`'s docstring now documents that it is **not reentrant** (`flock` is scoped per open-file-description, not per-process/thread) -- a sentence to keep it that way rather than a real bug found today.

One nit from the same comment -- documenting or dropping `aws_bedrock`/`azure_foundry`'s now-dead `save_extra_models()` -- turned out to be moot: those plugins were relocated to the separate `code_puppy_core_plugins` package in the interim (`27efe8f5`) and no longer exist in this repo.

### Testing

- Added regression tests pinning both bounded-read fixes (`tests/test_config_full_coverage.py`, `tests/mcp/test_project_config.py`).
- Rewrote the wizard-save test in `tests/mcp/test_config_wizard.py` to use a real `tmp_path` file instead of mocking `open()`/`os.path.exists()`/`json.dump` internals that no longer exist post-refactor -- that stale mocking had been silently letting a real file write leak to `/tmp/mcp_servers.json` on every test run (cleaned up).
- Full suite: **7461 passed, 28 skipped, 1 xpassed** (rebased onto the current `main` tip, including the `code_puppy_core_plugins` extraction and the session-persistence migration).
- `ruff check` and `ruff format --check` both clean.

cc @AndrewTilson -- both flagged gaps should be closed now.

### DIFF
--- a/code_puppy/atomic_io.py
+++ b/code_puppy/atomic_io.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 import contextlib
 import logging
 import os
+import stat
 import tempfile
 import time
 import uuid
@@ -87,6 +88,15 @@ def path_lock(
     Uses ``fcntl.flock`` on POSIX and ``msvcrt.locking`` on Windows. Raises
     :class:`LockTimeout` rather than blocking forever if another process
     (or another lock held by this same process) doesn't release in time.
+
+    Not reentrant: ``flock`` is scoped per *open file description*, not per
+    process or thread, so a second ``path_lock(path)`` (directly, or via
+    ``mutate_config``/``mutate_json`` on the same ``path``) taken from
+    *inside* an already-held lock on that same path will not recognize the
+    outer lock as "ours" -- it will block for the full ``timeout`` and then
+    raise :class:`LockTimeout`, rather than deadlock forever, but still not
+    proceed. Do not call ``mutate_config``/``mutate_json`` for a path from
+    within a ``mutation``/``mutate`` callback already locking that path.
     """
     lock_path = f"{path}.lock"
     os.makedirs(os.path.dirname(os.path.abspath(lock_path)), exist_ok=True)
@@ -179,7 +189,7 @@ def atomic_write_bytes(path: str, data: bytes) -> None:
 
     mode = None
     try:
-        mode = os.stat(target).st_mode
+        mode = stat.S_IMODE(os.stat(target).st_mode)
     except OSError:
         pass
 
--- a/code_puppy/command_line/mcp/trust_command.py
+++ b/code_puppy/command_line/mcp/trust_command.py
@@ -173,9 +173,11 @@ class TrustCommand(MCPCommandBase):
     def _safe_load_servers(config_file) -> dict:
         """Parse the project config for display only; never raise."""
         try:
+            from code_puppy import atomic_io
             from code_puppy.config import _parse_mcp_servers_mapping
 
-            return _parse_mcp_servers_mapping(config_file.read_text(encoding="utf-8"))
+            raw = atomic_io.read_bounded_bytes(str(config_file))
+            return _parse_mcp_servers_mapping(raw.decode("utf-8"))
         except Exception:
             return {}
 
--- a/code_puppy/config.py
+++ b/code_puppy/config.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 from typing import Any, Optional
 
+from code_puppy import atomic_io
 from code_puppy.config_file import load_config, mutate_config
 from code_puppy.session_storage import save_session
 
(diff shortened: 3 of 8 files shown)


WSxDemise:

### PR Summary
The wizard now uses the shared MCP JSON writer and the remaining direct readers use bounded I/O.

### Qodo Merge Triage
Not configured for OSS/public GitHub.

### Code Review Findings
#### Really Should Fix These
- [P0] [AC NOT MET] Trusted project configs bypass the size bound — `code_puppy/mcp_/project_config.py:125`
  - Impact: A trusted project file is fully buffered by `read_bytes()` during startup hash verification before the new bounded parser runs, preserving the memory-exhaustion path.
  - Fix: Stream the SHA-256 calculation with the same byte budget (or reject by size first) and test a genuinely oversized trusted file.

#### Nits
None.

WSxDemise (commented):
### Review Summary

LGTM. Recommending approval.

- **Locking & atomic I/O**: Swapping  to  closes the last uncoordinated writer to , properly serializing it with the other writers under .
- **Bounded reads**: Unbounded reads in , , and  have been replaced with , preventing unbounded memory allocation on oversized configs.
- **File mode preservation**:  mask in  correctly extracts standard permission bits when preserving file mode across atomic writes.
- **Tests & CI**: Dedicated test cases added for oversized trusted configs, bounded startup reads, and wizard persistence via the locked store. All CI checks (, , ) are passing.

WSxDemise (commented):
### Review Summary

LGTM. Recommending approval.

- **Locking & atomic I/O**: Swapping `config_wizard.py` to `upsert_mcp_server` closes the last uncoordinated writer to `mcp_servers.json`, properly serializing it with the other writers under `path_lock`.
- **Bounded reads**: Unbounded reads in `trust_command.py`, `config.py`, and `project_config.py` have been replaced with `atomic_io.read_bounded_bytes(...)`, preventing unbounded memory allocation on oversized configs.
- **File mode preservation**: `stat.S_IMODE(...)` mask in `atomic_write_bytes` correctly extracts standard permission bits when preserving file mode across atomic writes.
- **Tests & CI**: Dedicated test cases added for oversized trusted configs, bounded startup reads, and wizard persistence via the locked store. All CI checks (`quality`, `test (macos-latest, 3.13)`, `windows-encoding`) are passing.

WSxDemise (commented):
### Follow-up Review

No new commits since the prior review (HEAD still `1f4f5823d`, same as previously reviewed). No unresolved review threads. Prior findings (locking/atomic-io fix, bounded reads, `stat.S_IMODE` fix, added tests, green CI) stand unchanged and are not repeated here.

**New since last check**: GitHub now reports this PR as `mergeable_state: dirty` (merge conflicts with `main`). This is a rebase/merge-conflict issue, not a code-correctness regression — the reviewed diff content is unaffected. Recommend the author rebase onto current `main` before merge; approval recommendation stands once the branch is conflict-free.

PR #799, opened 2026-08-19
refactor: replace bespoke compaction with pydantic-ai-harness FallbackCompaction

mpfaffenberger:
## Summary

Adds `pydantic-ai-harness==0.23.0` and deletes all bespoke compaction code it supersedes.

```
13 files changed, 703 insertions(+), 1,824 deletions(-)   ->  net -1,121 lines
```

Compaction is now the harness `FallbackCompaction` chain (released last night in harness 0.23):

1. **`SummarizingCompaction`** — first wave, using the configured `summarization_model` resolved through `ModelFactory`
2. **`SlidingWindowCompaction`** — deterministic fallback when the summary call raises

Both waves get `keep_tokens = protected_token_count` and `max_tokens = compaction_threshold * model_context_length`, straight from `config.py`.

## What got deleted

- `summarization_agent.py` (177 lines): the dedicated summarizer sub-agent, its thread pool, and its atexit hook — the harness owns the summary call now
- ~450 lines of delicate ordering machinery in `_compaction.py` (660 -> 368 lines): `_merge_consecutive_same_role`, `_ensure_leading_request`, `_find_safe_split_index`, `split_for_protected_summarization`, `truncate`, `summarize` — the harness strategies preserve tool-call/tool-return pairing and provider ordering themselves
- `tests/agents/test_compaction_summarization_ordering.py` (569 lines): tested internals that no longer exist; those invariants are the harness's contract

## What stayed

- The accumulator-owning history processor (now **async + ctx-taking**, so the live `RunContext` reaches the harness and summary-call usage folds into the run's accounting)
- `/compact` and `/truncate` commands, reimplemented over the harness `compact_now` bridge — `/truncate` can no longer sever tool pairs the way raw list slicing could
- Mid-run `/compact` via the pause controller, dedup-hash bookkeeping, empty-thinking strip, trailing-response trim, tool-id sanitization, `filter_huge_messages`, spinner context updates, and the `pre_compact` / history-processor callbacks
- The live lilac integration tests, updated to spy on `SummarizingCompaction.compact` / `SlidingWindowCompaction.compact` instead of deleted internals

## Behavior changes

- A failed summarization now **slides the window** (drops oldest messages) instead of running the old bespoke truncation — same spirit, harness-shaped
- `FallbackCompaction` advances on `ModelAPIError` / `FallbackExceptionGroup` (the upstream defaults); any other strategy failure leaves history untouched for that cycle so the run always survives

## Testing

[...]

PR #805, opened 2026-08-19
feat: adopt harness ToolOutputLimits for oversized tool returns

mpfaffenberger:
## What

Attaches `pydantic_ai_harness.tool_output_limits.ToolOutputLimits` to every built agent, so oversized tool returns are reduced **once, at production time**, instead of persisting in history and getting re-sent on every later model request.

Continues the harness adoption that #799 (compaction) started — same playbook: harness owns the model-facing mechanics, Code Puppy keeps a thin config-glue module.

## Why

- Today an oversized tool return (huge shell output, giant grep, fat JSON) rides along in history until the history-time clamp (`_history.filter_huge_messages`) shrinks it to a **placeholder** at the 50k-token line — lossy, and the tokens were already paid on every prior request.
- The harness default band is `Spill(then=Truncate())`: **lossless**. The full payload is persisted and the model reads slices back on demand via the capability's bounded `read_tool_result(handle, offset, limit, from_end, pattern)` tool (the Claude Code pattern / pydantic-ai #4352 design). If the store write fails it falls back to bounded truncation — never a silent drop.
- See `docs/HARNESS_COMPAT_MATRIX.md` (included) for the full harness-vs-Code-Puppy redundancy sweep that ranked this as the first adoption target.

## How

- **`agents/_tool_output_limits.py`** (new, mirrors `_compaction.py`): `build_tool_output_limits()` wires config to the capability. Spills land under `~/.code_puppy/tool_output_overflow/` (config-dir convention, not the harness temp-dir default) with a 7-day TTL.
- **`config.py`**: new `/set` knob `tool_output_limit_chars` (default 10,000; `0` disables the capability entirely; garbage/negative fall back to default; no upper clamp).
- **`_builder.py`**: capability appended to the agent's capabilities list, built fresh per agent. Position relative to the `ProcessHistory` entries is inert (it hooks `after_tool_execute`, not history).
- **Kept**: `_history.filter_huge_messages` stays as the history-time backstop — it defends resumed sessions and imported histories that predate production-time reduction.
- **Guard test**: the steer-ordering guard in `test_runtime_pause_leakage.py` tracked the `capabilities=[` literal; updated to track the list built as a local (assertion strength unchanged, plus a fix for the `]`-inside-`List[Any]` slicing footgun).

## Interaction with the `spill` core-plugin

The plugin's `post_tool_call` hook runs inside the tool, before this capability sees the result, and only handles dict-shaped results over 32 KiB. They compose without double-spilling (the plugin's output is under this threshold's preview size). Follow-up in `code_puppy_core_plugins` will deprecate the plugin in favor of this capability, which also covers plain-string returns the plugin never could.

## Tests

- 11 new tests: config parsing, capability wiring shape, and end-to-end reduction through a real `pydantic_ai.Agent` run (oversized return reduced + spilled losslessly; small return untouched).
- Full suite: 7470 passed, 28 skipped, 1 xpassed.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -20,6 +20,7 @@ from pydantic_ai.capabilities import ProcessHistory
 
 from code_puppy.agents._compaction import make_history_processor
 from code_puppy.agents._steer_processor import make_steer_history_processor
+from code_puppy.agents._tool_output_limits import build_tool_output_limits
 from code_puppy.agents.event_stream_handler import event_stream_handler
 from code_puppy.callbacks import (
     on_pre_mcp_autostart,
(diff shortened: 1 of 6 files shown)


mpfaffenberger:
Superseded by #806, which lands the same ToolOutputLimits adoption plus the ClampOversizedMessages request-side guard and actually completes the filter_huge_messages deletion. The two useful deltas from this PR (the end-to-end spill test and the config unset-check hardening) were ported to #806 in 88c48420. The HARNESS_COMPAT_MATRIX doc moves to its own docs-only PR.

PR #806, opened 2026-08-19
feat: finish compaction migration — oversized-payload guards as pure harness capabilities

mpfaffenberger:
## What

Completes the harness compaction migration that #799 started. The last hand-rolled compaction subroutine — `filter_huge_messages` (shrink/drop >50k-token messages inside the history-processor closure) — is retired in favor of two pure `pydantic-ai-harness` capabilities wired straight into `Agent(capabilities=[...])`:

| Capability | Hook | Replaces |
|---|---|---|
| `ToolOutputLimits` | `after_tool_execute` (production time) | the request-side tool-return shrink: an oversized tool return is now reduced **before it ever enters history** — spilled losslessly to `~/.code_puppy/tool_output_spill/` (7-day TTL) and readable back via the harness `read_tool_result` tool, with bounded truncation as fallback |
| `ClampOversizedMessages` | `before_model_request` (request time) | the drop-if-still-huge path for response messages: runaway generations / giant tool-call args are head/tail-clamped under the same 50k-token budget — clamped, not dropped |

## Why this shape

- `filter_huge_messages` shrank a giant tool return to a placeholder *after* it had already been paid for once. `ToolOutputLimits` bounds it at the source and is **lossless** (the model can slice/grep the spilled payload on demand).
- The old filter **dropped** oversized response messages outright; the clamp keeps head+tail with a marker — strictly gentler, and it composes with the summarize/sliding-window strategies already adopted in #799.
- Capability ordering: the clamp sits after both `ProcessHistory` entries so it sees the final steer-injected history; `ToolOutputLimits` hooks a different phase so its position is inert (commented in `_builder.py`).

## Config

- `/set tool_output_limit_chars=<int>` — reduction threshold (default 10,000 chars; `0`/negative disables the capability entirely).
- Clamp budget stays hard-wired at the legacy 50k-token figure.

## Deleted

- `_history.filter_huge_messages` + `_shrink_oversized_tool_returns` + placeholder constant (−61 lines)
- their call site in `compact()`

## Transitional note

A >50k-token tool return already sitting in a **restored legacy session's recent tail** is no longer shrunk (new sessions can't produce one — returns are bounded at production). If that bites in practice, the harness-side fix is a request-side clamp option on `ClampOversizedMessages`, which I'd propose upstream via pydantic-ai-notes rather than resurrect the local filter.

## Follow-up (separate repos)

- `code_puppy_core_plugins`: deprecate the `spill` plugin (subsumed by `ToolOutputLimits`; harmless if both run — the plugin fires first and the reduced result passes under the harness threshold).

## Tests

- New `tests/agents/test_output_limits.py` (10 tests: bands, disable path, store location/TTL, clamp budget, config parsing)
- Removed the `filter_huge_messages` shrink test with the code it covered
- Full suite: **7468 passed, 10 skipped** locally

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -19,6 +19,10 @@ from pydantic_ai import Agent as PydanticAgent
 from pydantic_ai.capabilities import ProcessHistory
 
 from code_puppy.agents._compaction import make_history_processor
+from code_puppy.agents._output_limits import (
+    build_response_clamp,
+    build_tool_output_limits,
+)
 from code_puppy.agents._steer_processor import make_steer_history_processor
 from code_puppy.agents.event_stream_handler import event_stream_handler
 from code_puppy.callbacks import (
@@ -646,9 +650,15 @@ def build_pydantic_agent(
             # compacted away). ProcessHistory capabilities apply in
             # registration order (replaces the deprecated
             # `history_processors=` kwarg, removed in pydantic-ai v2).
+            # ToolOutputLimits reduces oversized tool returns on a different
+            # hook (after_tool_execute), so its position is inert; the
+            # response clamp runs before_model_request and sits LAST so it
+            # sees the final, steer-injected history.
             capabilities=[
+                *build_tool_output_limits(),
                 ProcessHistory(history_processor),
                 ProcessHistory(steer_processor),
+                build_response_clamp(),
             ],
             model_settings=model_settings,
         )
--- a/code_puppy/agents/_compaction.py
+++ b/code_puppy/agents/_compaction.py
@@ -37,7 +37,6 @@ from pydantic_ai_harness.compaction import (
 
 from code_puppy.agents._history import (
     estimate_tokens_for_message,
-    filter_huge_messages,
     hash_message,
     sanitize_tool_call_ids,
 )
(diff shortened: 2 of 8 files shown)


mpfaffenberger (commented):
Agent review (Biscuit, code-puppy-1b2940) focused on config.py key/value wiring, plus adjacent findings.

## Config: solid, three nits

1. **Key registration** — `tool_output_limit_chars` correctly appended to `default_keys` next to `agents_md_max_chars`, so `/set` autocomplete picks it up. Comment present.
2. **Default constant** — `TOOL_OUTPUT_LIMIT_CHARS_DEFAULT = 10_000` genuinely matches the harness `_DEFAULT_THRESHOLD` (verified against `tool_output_limits/_capability.py`). The docstring claim is accurate.
3. **Nit: `if not val:` is fragile.** `get_value` returns `str | None` today so `"0"` is truthy and the opt-out works — but if any future caller round-trips an int through the config cache, `not 0` reads an explicit opt-out as "unset" and silently re-enables the capability at 10k. Suggest `if val is None or not str(val).strip():` — one line, removes the trap.
4. **Nit: negative semantics are untested at the getter level.** `build_tool_output_limits` is tested with a monkeypatched `-5`, but nothing pins `get_tool_output_limit_chars() == -5` for the raw string `"-5"`. Since "negative disables" is a deliberate semantic (and differs from #805, which normalizes negatives back to the default — see below), pin it so a future cleanup can't flip disable back to default-ON. Alternatively: normalize negatives to `0` in the getter so "disabled" has exactly one representation and the `<= 0` convention doesn't have to be replicated by every future consumer.
5. **`CLAMP_MAX_PART_TOKENS` hard-wired at 50k** — agreed, keep it out of config; it's legacy parity and YAGNI applies. The PR body documents it, which is enough.
6. Verified: `clamp_tool_call_args` defaults to `True` in the harness, so the giant-tool-call-args claim in the PR body holds without passing it explicitly.
7. Verified: `ClampOversizedMessages` is a real `AbstractCapability`, and the steer-ordering guard in `test_runtime_pause_leakage.py` still passes against the spliced list literal (the first `]` after `capabilities=[` is still the list's closing bracket).

## Blocking coordination issue: collides with #805

#805 and this PR both add `tool_output_limit_chars`, `TOOL_OUTPUT_LIMIT_CHARS_DEFAULT`, and `get_tool_output_limit_chars()` — with **conflicting semantics** (negative -> default in #805, negative -> disable here), different glue modules (`_tool_output_limits.py` vs `_output_limits.py`), and different spill roots (`tool_output_overflow/` vs `tool_output_spill/`). If both merge mechanically the agent gets two `ToolOutputLimits` instances and two `read_tool_result` registrations.

Recommendation: **this PR supersedes #805** (it is a strict superset — same adoption plus the clamp and the `filter_huge_messages` deletion). Close #805 and port over its two useful deltas:
- the end-to-end test (`Agent.run` with a FunctionModel: asserts the oversized return is reduced AND the full payload is recoverable from the spill dir — this PR currently only shape-tests the builders, nothing exercises a real spill);
- the `val`-parsing hardening from nit 3.

## Transitional risk, quantified

The harness clamp docstring explicitly scopes out tool returns ("oversized tool *returns* are the job of `ClearToolResults`"), so a >50k-token tool return in a restored legacy session's protected tail is now covered by neither capability: TOL is production-time only, the clamp skips tool returns, and SlidingWindow protects the tail where such a monster lives. The transitional note in the PR body owns this honestly. If it ever bites, the cheap fix is a one-shot shrink at session-restore time (in `session_storage` load), not a permanent history-processor pass — worth a line in the note so future-us knows where the fix goes.

PR #813, opened 2026-08-20
feat: opt-in Logfire observability + powered-by-Pydantic banner tagline

mpfaffenberger:
## What

Two small, related things:

### 1. Opt-in Logfire observability
- New `enable_logfire` config key — **default OFF**, strictly opt-in (`/set enable_logfire true`). `CODE_PUPPY_ENABLE_LOGFIRE=1` env override for headless/CI.
- New `code_puppy/observability.py`: lazily imports `logfire`, calls `logfire.configure(send_to_logfire='if-token-present', console=False)` and `logfire.instrument_pydantic_ai()`. Double gate: even opted in, nothing ships without a Logfire token.
- Every failure path (missing package, bad config) warns via i18n'd message and fails soft — observability must never break the CLI.
- Real `logfire` is a new optional extra: `pip install code-puppy[logfire]` (default install stays lean; `[tool.logfire] ignore_no_config` was already in pyproject).

### 2. Banner tagline
Under the giant CODE PUPPY pyfiglet:

```
Powered by Pydantic · https://github.com/pydantic/pydantic-ai-harness
Need observability? Agentic governance? Controls? Check out Logfire / Pydantic Gateway → https://pydantic.dev/logfire
```

Prints even when pyfiglet is missing (tagline sits after the try/except).

## House rules respected
- No edits under `code_puppy/command_line/` — `/set` autocomplete comes free from `get_config_keys()` (same treatment as `enable_dbos`).
- All user-facing strings are `t()`-wrapped (PUP-473); i18n coverage gate passes.
- Could migrate to a `code_puppy_core_plugins` `startup`-hook plugin later if preferred — kept it core since instrumentation must land before any agent runs and it's a first-party Pydantic concern.

## Testing
- 10 new tests in `tests/test_observability.py` (default-off, env override, missing-package soft-fail, configure kwargs, error soft-fail)
- `tests/i18n` + config + cli_runner suites: 434 passed
- ruff check + format clean


mpfaffenberger:
Update: **logfire is now a hard dependency** (extra removed, lockfile updated). Opt-in behavior unchanged — no instrumentation without `enable_logfire`, no egress without a token. Bonus documented-by-test: the SDK honors `OTEL_EXPORTER_OTLP_*`, so opted-in users can ship spans to their own OTLP backend (collector/Tempo/Jaeger/etc.) with `send_to_logfire` untouched — Logfire cloud, BYO backend, or both.

PR #818, opened 2026-08-20
Refresh per-run model routing hook on current main

kvandre12-commits:
## Summary

Narrowly refreshes #760 onto current `main` while preserving Blayne Porterfield (`bporterfielddsc`) as the author of the original feature commit.

The original PR is conflict-stale but remains valuable: it adds a `model_select` callback that can choose an agent model once per run without mutating persisted global, pinned, or JSON-agent configuration.

## Changes in this refresh

- resolve the sole current-main conflict by preserving both runtime system-prompt additions and the new per-run model override state
- pass the current prompt to selectors after `user_prompt_submit` callbacks have rewritten it
- document the callback contract in `AGENTS.md`
- preserve explicit runtime model overrides as the highest-precedence choice

Passing the current prompt is part of the model-selection contract: prior history alone cannot implement the motivating trivial-vs-complex current-turn routing use case.

This refresh deliberately does **not** change shared callback-dispatch semantics, persisted model configuration, or pinned-model precedence.

## Attribution

- `a479dea9` — original #760 feature, authored by Blayne Porterfield
- `8e00c099` — current-prompt contract and documentation
- `976bc92d` — scope reduction restoring unchanged shared callback dispatch

The original author's commit was not squashed or reattributed.

## Validation

- `410 passed` across `tests/agents`, pinned-model regressions, and model-select tests
- `48 passed` across model-select, pinned-model, and callback failure-isolation tests
- Ruff `E4,E7,E9,F,I` checks passed on changed Python files
- Ruff formatting check passed
- `git diff --check` passed

Supersedes the conflict-stale branch in #760; that PR remains the original design discussion.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,7 @@ approval. With `fail_closed=True` its exception is reported as a block instead. 
 | `invoke_agent` | Sub-agent invoked | `(*args, **kwargs) -> None` |
 | `agent_exception` | Unhandled agent error | `(exception, *args, **kwargs) -> None` |
 | `agent_run_start` | Before agent task | `(agent_name, model_name, session_id=None) -> None` |
+| `model_select` | Select a model for one run | `(*, agent_name, current_model, prompt, messages, session_id=None) -> str \| None` — first non-empty result wins |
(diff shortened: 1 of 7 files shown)


mpfaffenberger (approved):
I like it.

PR #226, opened 2026-03-07
Clarify and structure the CodePuppyAgent system prompt

janfeddersen-wq:
## Summary
- reorganize `CodePuppyAgent.get_system_prompt()` into clearly labeled sections
- fix the malformed origin string in the prompt and tighten the wording for the Code Puppy self-description
- keep the existing reasoning/tool-use guidance intact while making the overall prompt easier to scan and maintain

## What changed in `agent_code_puppy`
- added markdown-style section headers: `Identity`, `Personality & Tone`, `Code Standards`, `About You`, `Workflow`, and `Rules`
- switched the prompt literal to `f"""\` so the generated prompt no longer starts with a leading blank line
- restored the missing closing quote/period in the origins response line
- changed the "what is code puppy" prompt text to the more explicit "what Code Puppy is" phrasing
- removed the duplicated/extra `Continue autonomously unless user input is definitively required` rule because the workflow section already covers autonomous execution

## Files in scope
- `code_puppy/agents/agent_code_puppy.py`

## Testing
- Not run (not requested)

## Release Notes

* **Chores**
  * Restructured and enhanced the AI agent's operational guidelines with reorganized sections covering code standards, workflow processes, and behavioral rules.

PR #820, opened 2026-08-20
Refresh approved CodePuppy prompt cleanup

kvandre12-commits:
## Summary

Companion refresh for #226 on current `main`.

This preserves Jan Feddersen's original prompt-structure commit, including its original authorship, and adds only the explicit unresolved inline-review correction: the prompt now names `agent_run_shell_command`, matching `CodePuppyAgent.get_available_tools()`.

## Commits

1. `Refine CodePuppyAgent system prompt structure` — original #226 commit by Jan Feddersen
2. `Align prompt with shell tool name` — narrow review correction and regression assertion

## Validation

- `tests/test_coverage_agents_gaps.py` plus `tests/agents`: **414 passed**
- Ruff check/format passed for both touched files
- `git diff --check` passed

The repository-wide Ruff command with the newly pinned Ruff 0.15 could not be reproduced on Termux because `uvx` had to compile Ruff from source and timed out; hosted CI remains authoritative for that check.

Supersedes #226 after preserving its contributor commit.

### DIFF
--- a/code_puppy/agents/agent_code_puppy.py
+++ b/code_puppy/agents/agent_code_puppy.py
@@ -48,7 +48,7 @@ class CodePuppyAgent(BaseAgent):
             ),
             "loop_rule": (
                 "- You're encouraged to loop between reasoning, file "
-                "tools, and run_shell_command to test output in order "
+                "tools, and agent_run_shell_command to test output in order "
                 "to write programs"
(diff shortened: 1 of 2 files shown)


mpfaffenberger:
I don't think we need these changes.

PR #823, opened 2026-08-21
fix(sessions): leave migrated .pkl files in place, don't archive them away

WSxDemise:
## What

Stop the pickle→JSON session migration from *moving* successfully-migrated
legacy `.pkl` session files into `pre_v2_backup/`. `archive_legacy_pickle()`
is now a no-op: the original pickle simply stays exactly where it always was,
right next to its new `.json` twin.

## Why

The migration never deleted pickles, but it did relocate them into a backup
subfolder — and that's just as bad in practice. Any older code_puppy version
(and even the current version's own `load_session` fallback) only ever looks
for `<name>.pkl` directly next to the session file; none of them know
`pre_v2_backup/` exists. So downgrading after a migration *looked and felt*
exactly like losing every session, even though the bytes were technically
still on disk somewhere you'd never think to check.

## How

- `archive_legacy_pickle()` in `code_puppy/session_format_migration.py` is now
  a no-op with an explanatory docstring — callers still invoke it (from
  `session_storage.load_session`'s lazy-migration path and the startup sweep),
  but it no longer touches the filesystem.
- Updated the module docstring and the startup-sweep summary log line to
  describe the new (correct) behavior.
- New sessions were already JSON-only — `save_session` never wrote a `.pkl`
  in the first place — so this change only affects **pre-existing** legacy
  pickles that get lazily migrated; nothing changes for anything created on
  the current version.
- Left the separate quarantine-for-unmigratable-pickles path
  (`pre_v2_backup/failed/`) untouched. That's a distinct, already-tested
  self-healing retry subsystem for genuinely corrupt pickles — out of scope
  for this fix.

## Testing

- `uv run --frozen pytest tests/test_session_format_migration.py` — 24/24 passed
  (updated 3 assertions that encoded the old move-to-backup behavior)
- `uv run --frozen pytest tests/test_session_migration.py tests/test_session_pickle_loading_safety.py tests/test_session_scoping.py tests/test_session_storage_coverage.py tests/test_agent_tools.py` — all passing
- `uv run --frozen ruff check` / `ruff format --check` on changed files — clean

## Scope

Two files: `code_puppy/session_format_migration.py` (production fix) and
`tests/test_session_format_migration.py` (updated assertions). No dependency,
lockfile, tooling, or CI/config changes.

### DIFF
--- a/code_puppy/session_format_migration.py
+++ b/code_puppy/session_format_migration.py
@@ -12,8 +12,10 @@ caller/validation layer: it round-trips the normalized messages through
 declaring a migration successful.
 
 Startup entry point is :func:`sweep_legacy_pickle_sessions`, idempotent via a
-marker file in the config dir. Originals are never deleted: migrated pickles
-move to ``<dir>/pre_v2_backup/``, failures to ``<dir>/pre_v2_backup/failed/``.
+marker file in the config dir. Originals are never deleted or moved on a
+successful migration -- the ``.pkl`` stays next to its new ``.json`` twin so
+an older code_puppy version can still find it after a downgrade. Only
+unmigratable pickles are relocated, to ``<dir>/pre_v2_backup/failed/``.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


WSxDemise:
## Discussion needed before merging

This PR currently makes `archive_legacy_pickle()` a no-op so migrated legacy
`.pkl` session files stay in the **same folder** as their new `.json` twin,
rather than being moved into `pre_v2_backup/`.

Before merging we want the team's sign-off on which behavior we actually want:

1. **Leave `.pkl` in place next to `.json`** (current state of this PR) — an
   older code_puppy version can find and load it automatically after a
   downgrade, with zero manual steps. Trade-off: permanent double storage for
   every migrated session (small text blobs, but non-zero).
2. **Keep moving `.pkl` into `pre_v2_backup/`** (original behavior) — tidier
   session directories, and nothing is ever deleted, but an older version's
   loader has no idea that folder exists, so a downgrade looks and feels like
   total session loss unless someone manually moves the files back.

Marking this Draft until we've discussed and agreed on the intended trade-off.
No reviewers assigned yet.

PR #825, opened 2026-08-21
feat: add outbound model message transform callback

kevinMEH:
## What

Add a narrow plugin extension point for transforming the final message list before it is sent to a model.

## Why

This enables plugins to contribute request-scoped context after Code Puppy has finished assembling, processing, and compacting history. Example uses include:

- Injecting consistent project or user context into every model request that must survive compaction.
- Applying outbound policy, redaction, or annotation immediately before provider submission.
- Enabling advanced context-management techniques such as cache-aware tool-call trimming, message dropping, and message merging.
- Building model-independent integrations without replacing Code Puppy's agent or model.

This is an established extension boundary in other agent harnesses. OpenCode, for example, exposes `experimental.chat.messages.transform` so plugins can modify final outbound messages without changing persisted session history.

Code Puppy already provides rich plugin callbacks around prompts, agents, and tools, but it does not currently expose this final request boundary. Earlier prompt hooks run before history processing and do not cover every request within an agent run.

I need this for [Operator Memory](https://github.com/aerovato/operator-memory), which provides agents with durable project and user context. Operator renders an immutable context preamble once per conversation and must attach it to every outbound request, including subagent requests and requests between tool calls. The preamble must be added after compaction so it cannot be discarded, while remaining outside persisted conversation history.

Operator currently accomplishes this by temporarily overriding Code Puppy's model. That works through public Pydantic AI APIs, but it is not a supported Code Puppy plugin pattern and is superseded when DBOS installs its own model override. A message-transform callback provides the narrow integration boundary needed without exposing model replacement or arbitrary Pydantic AI capabilities.

## Example

```python
from code_puppy.callbacks import register_callback
from pydantic_ai.messages import ModelMessage, ModelRequest, UserPromptPart


async def transform_model_messages(
    agent_name: str | None,
    messages: list[ModelMessage],
) -> None:
    context = await load_request_context(agent_name)

    messages.insert(
        0,
        ModelRequest(
            parts=[UserPromptPart(content=context)],
        ),
    )


register_callback("transform_model_messages", transform_model_messages)
```

Callbacks mutate the shared message list in place and run sequentially in plugin registration order. Both synchronous and asynchronous callbacks are supported.

## Implementation

- Add `transform_model_messages` to the existing callback registry.
- Dispatch it through one internal Pydantic AI model-request hook.
- Install the hook after built-in history processing for main agents and subagents.
- Copy the request context and message list before transformation so changes affect only the current model request.
- Reuse existing callback ordering, failure isolation, ownership tracking, and disabled-plugin filtering.
- Use the same request path for streaming, non-streaming, tool-loop, and DBOS-wrapped model requests.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ approval. With `fail_closed=True` its exception is reported as a block instead. 
 | `load_model_descriptions` | Inject description overlays | `() -> dict[str, str]` |
 | `get_model_system_prompt` | Per-model prompt | `(model_name, default_prompt, user_prompt) -> dict \| None` |
 | `stream_event` | Response streaming | `(event_type, event_data, agent_session_id=None) -> None` |
+| `transform_model_messages` | Before each model request, after history processing | `(agent_name, messages) -> None` — mutate the final `list[ModelMessage]` in place |
 | `pre_mcp_autostart` | Before bound MCP servers auto-start | `(agent_name, server_names) -> None` (refresh tokens / mint creds here) |
 
 Full list + rarely-used hooks: see `code_puppy/callbacks.py` source.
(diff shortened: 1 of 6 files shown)


kevinMEH:
PS: There are 16 existing failing tests unrelated to this PR

PR #833, opened 2026-08-21
Promote model delivery to a ResolvedModel capability

mpfaffenberger:
## Summary

Sixth in the capability series (#828 SteerInjection, #829 HistoryCompaction, #830 PluginMessageTransform, #831 PerModelSettings, #832 AssembledInstructions): promotes **model delivery** from the `Agent(model=...)` constructor kwarg to a first-class pydantic-ai capability, `ResolvedModel`, on the dedicated `get_model()` seam.

Both construction sites (`agents/_builder.py` and `tools/subagent_invocation.py`) previously handed their fully resolved `Model` instance to the constructor kwarg while everything else about the agent's behavior already composes through `capabilities=[...]`.

## What changed

- New `code_puppy/agents/_resolved_model.py` — `@dataclass ResolvedModel(AbstractCapability[Any])` with a single `model: Model` field, delivering via `get_model()`.
- `_builder.build_pydantic_agent` drops `model=` and hoists **one shared `ResolvedModel` instance** across the probe + final construction passes (mirrors the single `load_model_with_fallback` call; safe because the capability is a static snapshot, default `for_run` returns `self`, and it carries no id).
- `subagent_invocation` drops `model=` and constructs `ResolvedModel(model)` inline.
- **Resolution logic untouched** (same restraint as #831/#832): `load_model_with_fallback` still owns the fallback chain + per-conversation warning dedup; sub-agent pin/explicit-override resolution unchanged; `BaseAgent.cur_model` still set. `build_tool_probe_for_agent` keeps its `model=` kwarg deliberately (stripped probe, per the #832 precedent for `instructions=""`).

## Parity (verified against pydantic-ai 2.31.0 source)

- **Static contribution, resolved once per run.** A non-callable `get_model()` return is used directly as the run's model (`Agent._evaluate_model_contribution`), exactly like the agent-slot model. `_check_dynamic_model_resume` is a no-op for static contributions — no selector semantics introduced.
- **`run(model=...)` and `Agent.override(model=...)` stay authoritative.** Both set `model_is_explicit`, which short-circuits capability contribution evaluation entirely — same precedence as the old kwarg. Pinned by tests.
- **Enter/exit lifecycle preserved.** `Agent.__aenter__` enters a static capability model's context identically (line ~3736: `static_selection = capability_model`), so provider HTTP clients open/close the same.
- **One observable divergence:** the built agent's `.model` property now reads `None` (the model no longer occupies the agent slot). Verified: **zero** code_puppy readers — everything goes through `BaseAgent.cur_model`. One plugin reader exists: wiggum's `_resolve_judges` probes `get_pydantic_agent().model.model_name` for its **default-judge fallback** and now takes its own next fallback, `BaseAgent.get_model_name()` — the configured model name, which is what judge configs resolve against anyway (arguably more correct than the provider model id it used to get). Degrades gracefully by design (`getattr(None, "model_name", None)`); documented in the module docstring and pinned by `test_agent_model_slot_reads_none`.
- **Not spec-constructible by design:** `get_serialization_name() -> None` because the capability holds a live, provider-configured `Model` instance (HTTP clients and all) — same precedent as SteerInjection/HistoryCompaction.
- **List position inert:** `get_model` is a configuration seam, not a request hook; placed first in both blocks for readability.

## Tests

7 contract tests in `tests/agents/test_resolved_model_capability.py`:
1. direct seam identity (`get_model() is model`)
2. spec-construction opt-out
3. kwarg-vs-capability **wire parity** via `FunctionModel` (model-visible content compared; timestamps excluded)
4. `run(model=...)` authoritative over the capability
5. `override(model=...)` authoritative in scope, capability restored outside
6. pinned `.model is None` divergence
7. end-to-end `build_pydantic_agent` drive: built agent runs on the capability-delivered model, `cur_model` tracks it, agent slot empty

Full suite: **7583 passed**; the 3 reds are the documented pre-existing main failures (`test_render_version_check_current`, plus the two full-run-order `FieldInfo has no attribute 'strip'` flakes noted in #832) — both `subagent_invocation`-adjacent ones pass in isolation on this branch.

## Merge-order note

Like its five siblings, this touches the shared `capabilities=[...]` blocks in `_builder.py` + `subagent_invocation.py`. Whichever PRs land last eat trivial rebases.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -24,6 +24,7 @@ from code_puppy.agents._output_limits import (
     build_response_clamp,
     build_tool_output_limits,
 )
+from code_puppy.agents._resolved_model import ResolvedModel
 from code_puppy.agents._steer_processor import make_steer_history_processor
 from code_puppy.agents.event_stream_handler import event_stream_handler
 from code_puppy.callbacks import (
@@ -639,10 +640,14 @@ def build_pydantic_agent(
     history_processor = make_history_processor(agent)
     steer_processor = make_steer_history_processor(agent)
     logical_agent_name = getattr(agent, "name", None) or agent.__class__.__name__
+    # One shared instance across the probe and final construction passes:
+    # the capability is a static snapshot (default ``for_run`` returns
+    # ``self``) and carries no id, so sharing is safe and mirrors the single
+    # model resolution above.
+    resolved_model = ResolvedModel(model)
 
     def _new_pydantic_agent(toolsets: List[Any]) -> PydanticAgent:
         return PydanticAgent(
-            model=model,
             # Explicit name: without it pydantic-ai infers one from the
             # caller's frame variables, so observability spans read
             # "invoke_agent pydantic_agent" instead of the logical agent name.
@@ -660,7 +665,11 @@ def build_pydantic_agent(
             # hook (after_tool_execute), so its position is inert; the
             # response clamp runs before_model_request after both history
             # processors. The plugin transform wraps the final model request.
+            # ResolvedModel is a configuration seam (get_model), not a
+            # request hook, so its position is inert too; it replaces the
+            # former ``model=`` constructor kwarg.
             capabilities=[
+                resolved_model,
                 *build_tool_output_limits(),
                 ProcessHistory(history_processor),
                 ProcessHistory(steer_processor),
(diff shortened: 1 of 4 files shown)


mpfaffenberger:
Review trail: code-puppy clone did two passes. Pass 1: clean APPROVE with three optional test-hardening findings (pin the Agent.__aenter__/__aexit__ lifecycle of the capability model, prove static delivery across a multi-step tool-call run, cover the sub-agent construction site end-to-end). All three applied in c5113779. Pass 2: APPROVE, confirmed the new tests are non-vacuous (lifecycle counters, forced two-step run, subagent test fails if the capability is dropped). 10/10 contract tests green; full suite matches main's failure profile (3 documented pre-existing reds, all reproduce on clean main).

PR #853, opened 2026-08-23
fix: support GPT-5.6 Responses reasoning

### LINKED ISSUE #746: GPT-5.x models added with `/add_model` fail with tools due to   `reasoning_effort`
ran into this while trying to use OpenAI GPT-5.x models added
  through `/add_model`.

  I may be missing some intended setup here, so I wanted to open an
  issue before attempting a PR.

  ## What happened

  I added OpenAI models through `/add_model`, including:

  - `openai-gpt-5.5`
  - `openai-gpt-5.6-terra`

  When I selected one of those models and asked Code Puppy to inspect/
  work on a repo, the request failed with:
 ```text
  Function tools with reasoning_effort are not supported for gpt-5.5
  in /v1/chat/completions.
  To use function tools, use /v1/responses or set reasoning_effort to
  'none'.

  DeepSeek, added as a custom_openai model, worked in the same install,
  so this seems specific to the OpenAI GPT-5.x entries.

  ## Environment

  - OS: Windows
  - Install method: uvx code-puppy
  - Python: 3.12
  - Model that worked: deepseek-deepseek-v4-pro
  - Models that failed:
      - openai-gpt-5.5
      - openai-gpt-5.6-terra

  ## Config shape

  The OpenAI entry looked roughly like this:

  {
    "openai-gpt-5.5": {
      "type": "openai",
      "provider": "openai",
      "name": "gpt-5.5",
      "context_length": 1050000,
      "supported_settings": [
        "temperature",
        "top_p",
         "reasoning_effort",
        "verbosity"
      ]
    }
  }

  ## Possible cause

  From a quick look at the code, it seems like:

  - /add_model can create GPT-5.x OpenAI entries as type: openai
  - model_factory.py adds openai_reasoning_effort for gpt-5 model names
  - type: openai appears to use OpenAIChatModel unless the model key
    contains codex

  So this may be sending GPT-5.x + tools + reasoning_effort through Chat
  Completions, which OpenAI rejects.

  I also noticed the README says to use custom_openai_responses when a
  model/endpoint requires the Responses API, so I’m not sure whether
  these GPT-5.x OpenAI entries should be routed differently or whether /
  add_model should generate a different config.

  ## Question

  Is this expected behavior/configuration user error, or should GPT-5.x
  OpenAI models added through /add_model avoid this Chat Completions +
  reasoning_effort path?

  If it’s a bug, I’m happy to try a small PR, but I wanted to check
  first because there seems to be related GPT-5.6/OpenAI work already in
  progress.

jmdots:
## Summary
- support GPT-5.6 OpenAI aliases through the Responses API
- configure native reasoning mode/context and encrypted reasoning replay
- preserve signed empty reasoning parts needed before tool calls
- preserve complete streamed output items, including encrypted reasoning
- add regression coverage for model settings, compaction, and response reconstruction

Closes #746

## Compatibility note
Existing conversations created before this change may contain incomplete or incompatible Responses API reasoning history. Resuming those conversations may not work reliably; start a fresh conversation if resume fails.

## Testing
- `uv run pytest tests/test_compaction_reasoning.py tests/test_model_factory_reasoning.py tests/test_chatgpt_codex_client.py -q` (61 passed)

### DIFF
--- a/code_puppy/agents/_compaction.py
+++ b/code_puppy/agents/_compaction.py
@@ -263,7 +263,7 @@ async def compact(
 def _strip_empty_thinking_parts(
     messages: List[ModelMessage],
 ) -> Tuple[List[ModelMessage], int]:
-    """Remove empty ThinkingParts; drop messages rendered empty by removal."""
+    """Remove empty ThinkingParts without discarding replay signatures."""
     cleaned: List[ModelMessage] = []
     filtered_count = 0
     for msg in messages:
@@ -272,16 +272,22 @@ def _strip_empty_thinking_parts(
             len(parts) == 1
             and isinstance(parts[0], ThinkingPart)
             and not parts[0].content
+            and not parts[0].signature
         ):
             filtered_count += 1
             continue
-        if any(isinstance(p, ThinkingPart) and not p.content for p in parts):
+        if any(
+            isinstance(p, ThinkingPart) and not p.content and not p.signature
+            for p in parts
+        ):
             msg = dataclasses.replace(
                 msg,
                 parts=[
                     p
                     for p in parts
-                    if not (isinstance(p, ThinkingPart) and not p.content)
+                    if not (
+                        isinstance(p, ThinkingPart) and not p.content and not p.signature
+                    )
                 ],
             )
             if not msg.parts:
@@ -300,7 +306,7 @@ def make_history_processor(agent: Any) -> Callable[..., Any]:
          (preserving the last-message regardless of compacted-hash collisions).
       3. Runs ``compact(...)`` if we're over threshold (or ``/compact`` forced).
       4. Records dropped-message hashes in ``agent._compacted_message_hashes``.
-      5. Strips empty ThinkingParts.
+      5. Strips empty ThinkingParts that carry no replay signature.
       6. Trims trailing ModelResponse messages so history ends with a ModelRequest.
       7. Fires ``on_message_history_processor_end``.
 
--- a/code_puppy/chatgpt_codex_client.py
+++ b/code_puppy/chatgpt_codex_client.py
@@ -25,6 +25,52 @@ import httpx
 logger = logging.getLogger(__name__)
 
 
+def _merge_output_items(
+    envelope_output: list[dict], streamed_items: list[dict]
+) -> list[dict]:
+    """Merge `response.completed` envelope output with streamed item payloads.
+
+    The streamed ``output_item.done`` payloads are richer (they carry
+    reasoning ids and ``encrypted_content`` needed for replay), so they win
+    for any item present in both. Items only the envelope saw (a dropped
+    stream event) keep their envelope position; items only the stream saw
+    (a partial store=false envelope) are inserted before the item that
+    followed them in stream order, preserving reasoning-before-message
+    pairing.
+    """
+    envelope_ids = {item.get("id") for item in envelope_output if item.get("id")}
+    if not envelope_ids:
+        return list(streamed_items)
+
+    streamed_by_id = {item.get("id"): item for item in streamed_items if item.get("id")}
+    if envelope_ids <= streamed_by_id.keys():
+        # Envelope is a subset of the stream: stream order is complete.
+        return list(streamed_items)
+
+    # Rare: the envelope has an item the stream missed. Use the envelope as
+    # the spine, swap in richer streamed twins, and slot stream-only items
+    # ahead of their stream successor (or at the end).
+    pending_before: dict[str, list[dict]] = {}
+    carry: list[dict] = []
+    for item in streamed_items:
+        item_id = item.get("id")
+        if item_id in envelope_ids:
+            if carry:
+                pending_before[item_id] = carry
+                carry = []
+        else:
+            carry.append(item)
+    tail = carry
+
+    merged: list[dict] = []
+    for item in envelope_output:
+        item_id = item.get("id")
+        merged.extend(pending_before.get(item_id, []))
+        merged.append(streamed_by_id.get(item_id, item))
+    merged.extend(tail)
+    return merged
+
+
 def _is_reasoning_model(model_name: str) -> bool:
     """Check if a model supports reasoning parameters."""
     reasoning_models = [
@@ -300,9 +346,13 @@ class ChatGPTCodexAsyncClient(httpx.AsyncClient):
             response_body = dict(final_response_data)
             # The completed envelope may contain a partial output (for example,
             # only the message). Prefer the complete output_item.done payloads,
-            # which preserve reasoning ids and encrypted_content for replay.
+            # which preserve reasoning ids and encrypted_content for replay --
+            # but merge rather than replace, so an item whose done event was
+            # dropped mid-stream is never lost if the envelope still has it.
             if completed_output_items:
-                response_body["output"] = completed_output_items
+                response_body["output"] = _merge_output_items(
+                    response_body.get("output") or [], completed_output_items
+                )
             else:
                 # No items captured either — fall back to text/tool deltas.
                 rebuilt: list[dict] = []
--- a/code_puppy/chatgpt_codex_client.py
+++ b/code_puppy/chatgpt_codex_client.py
@@ -298,10 +298,12 @@ class ChatGPTCodexAsyncClient(httpx.AsyncClient):
         # (empty when store=false) output with the collected output_item.done items.
         if final_response_data:
             response_body = dict(final_response_data)
-            existing_output = response_body.get("output") or []
-            if not existing_output and completed_output_items:
+            # The completed envelope may contain a partial output (for example,
+            # only the message). Prefer the complete output_item.done payloads,
+            # which preserve reasoning ids and encrypted_content for replay.
+            if completed_output_items:
                 response_body["output"] = completed_output_items
-            elif not existing_output:
+            else:
                 # No items captured either — fall back to text/tool deltas.
                 rebuilt: list[dict] = []
                 if collected_text:
--- a/code_puppy/chatgpt_codex_client.py
+++ b/code_puppy/chatgpt_codex_client.py
@@ -153,7 +153,7 @@ class ChatGPTCodexAsyncClient(httpx.AsyncClient):
             forced_stream = True  # Only convert if WE forced streaming
             modified = True
 
-        # Add reasoning settings for reasoning models (gpt-5.2, o-series, etc.)
+        # Add the default reasoning settings for supported reasoning models.
         model = data.get("model", "")
         if "reasoning" not in data and _is_reasoning_model(model):
             data["reasoning"] = {
--- a/code_puppy/model_factory.py
+++ b/code_puppy/model_factory.py
@@ -146,9 +146,18 @@ def _thinking_tags_profile(
     from code_puppy.model_utils import get_thinking_tags
 
     tags = get_thinking_tags(model_name, model_config)
-    if tags is None:
-        return None
-    return OpenAIModelProfile(thinking_tags=tags)
+    profile_kwargs: dict[str, Any] = {}
+    if tags is not None:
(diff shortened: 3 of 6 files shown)


jmdots:
This isn't right, yet. I think it has something to do with "all_turns" reasoning context. 

```
"/REDACTED/REDACTED/src/code_puppy/.venv/lib/python3.13/site-packages/pydantic_ai/models/ope
nai.py", line 221, in _map_api_errors
    raise ModelHTTPError(
        status_code=status_code, model_name=model_name, body=e.body,
headers=dict(e.response.headers)
    ) from e
pydantic_ai.exceptions.ModelHTTPError: status_code: 400, model_name: gpt-5.6-luna,
body: {'message': "Item 'REDACTED' of
type 'message' was provided without its required 'reasoning' item:
'REDACTED'.", 'type':
'invalid_request_error', 'param': 'input', 'code': None}
```

mpfaffenberger:
I was like 99% sure this is already working perfectly. Can you tell me a little more about what you're observing?

`all_turns` - this is our default, and it keeps thinking signatures and thinking blocks from the entire history. 

Without this, GPT 5.6 family models will suffer a 10%ish intelligence loss. Maybe more.

jmdots:
Hi. Please try with mainline cope puppy to use high reasoning level on OpenAI API (platform side, not plus), after adding the gpt-5.6 luna via /add_model. I saw the failure mode by doing that. The linked issue shows someone else faced similar. Open to thoughts. Thanks!

kvandre12-commits:
I treated this as an `all_turns` policy contract rather than only a generic history-retention question and ran the same GPT-5.6/Luna-shaped mutants before and after this PR.

A/B result:

- pre-#853 base `ea0a9b73`: 5/5 policy mutants fail
- #853 head `7e3e2e0a`: isolated `all_turns` provider-shaped suite passes

The mutants cover:
- `gpt-5.6-luna` alias resolving to `openai_reasoning_context="all_turns"`
- syntactically empty `ThinkingPart(content="")` carrying a replay signature
- provider reasoning rows with `summary=[]` or `summary=None` plus encrypted replay state
- partial `response.completed` envelopes that contain the assistant message but omit its required reasoning sibling

The important distinction seems to be:

`all_turns` can be the correct policy, while the retained history still needs provider-shaped reasoning artifacts preserved across normalization / compaction / response reconstruction.

One caveat on the pre-853 control: the failures begin at the policy surface because the Luna alias does not expose `openai_reasoning_context="all_turns"` there, so I would not interpret the 5 failures as 5 independent cleanup defects. What the A/B does show is that pre-853 does not establish the full Luna `all_turns` contract, while this PR does, and the downstream provider-shaped preservation mutants then pass.

Post-853 proof branch:
https://github.com/kvandre12-commits/code_puppy/tree/test/853-isolated-proof

Pre-853 control:
https://github.com/kvandre12-commits/code_puppy/pull/4

mpfaffenberger:
Looking into this

mpfaffenberger:
Confirmed, I reproduced both failure modes against the live API on mainline (`bde6d89f`). Thanks for pushing back on my "99% sure", the missing ingredient was streaming: the TUI always streams, and only the streaming path surfaces GPT-5.6's pre-tool-call reasoning item (`summary=[]`, `encrypted_content` present) as `ThinkingPart(content="", signature=...)`. Non-streaming doesn't even return the reasoning item, which is why my quick checks never saw it.

From there it's exactly what this P/R fixes. Our history processor strips empty ThinkingParts without checking the signature, and since `gpt-5.6-luna`'s profile enables encrypted reasoning replay, pydantic-ai sends item ids back and the API rejects the turn:

```
400: Item 'fc_09cbe...' of type 'function_call' was provided without its
required 'reasoning' item: 'rs_09cbe...'
```

Same validator as your `'message'` variant, it just names whichever item followed the stripped reasoning item. The plain `/add_model` shape (`type: openai`) also reproduces #746's chat-completions rejection on the first request, so the routing half of this P/R checks out too. You were right about it not being an `all_turns` policy problem, the policy is fine, the retained history was losing the signed empty reasoning part.

<details><summary>Repro script (real OpenAI API, isolated HOME, ~$0.02)</summary>

```python
"""Repro for mpfaffenberger/code_puppy#853 (and its parent issue #746).

Runs against the REAL OpenAI platform API (needs OPENAI_API_KEY).

Phase A: gpt-5.6-luna exactly as /add_model creates it (type "openai")
         + reasoning_effort=high + a tool. Expect issue #746's error:
         Chat Completions rejecting function tools with reasoning_effort.

Phase B: same underlying model on the Responses API route
         (custom_openai_responses). Turn 1 forces a tool call, history is
         passed through mainline's _strip_empty_thinking_parts (step 5 of
         the history processor), turn 2 replays it. Expect jmdots' 400:
         "Item 'msg_...' of type 'message' was provided without its
         required 'reasoning' item: 'rs_...'".
"""

import asyncio
import json
import os
import sys
import traceback

# Isolate all code_puppy state BEFORE importing it (config paths use HOME).
os.environ["HOME"] = "/tmp/cp-853-home"
os.makedirs("/tmp/cp-853-home/.code_puppy", exist_ok=True)

EXTRA_MODELS = {
    # Byte-for-byte what add_model_menu.build_model_config emits for
    # provider=openai, model gpt-5.6-luna.
    "openai-gpt-5.6-luna": {
        "type": "openai",
        "name": "gpt-5.6-luna",
        "context_length": 400000,
        "supported_settings": [
            "temperature",
            "top_p",
            "reasoning_effort",
            "verbosity",
        ],
    },
    # Same model, Responses API route (the README-documented workaround
    # for #746, and the route where jmdots hit the reasoning-item 400).
    # NB: the key MUST contain "gpt-5.6" -- make_model_settings matches
    # reasoning config against the config key, not the underlying name.
    "gpt-5.6-luna": {
        "type": "custom_openai_responses",
        "name": "gpt-5.6-luna",
        "custom_endpoint": {
            "url": "https://api.openai.com/v1",
            "api_key": "$OPENAI_API_KEY",
        },
        "context_length": 400000,
        "supported_settings": [
            "temperature",
            "top_p",
            "reasoning_effort",
            "verbosity",
        ],
    },
}

with open("/tmp/cp-853-home/.code_puppy/extra_models.json", "w") as f:
    json.dump(EXTRA_MODELS, f, indent=2)

from code_puppy.config import set_model_setting  # noqa: E402

set_model_setting("openai-gpt-5.6-luna", "reasoning_effort", "high")
set_model_setting("gpt-5.6-luna", "reasoning_effort", "high")

from pydantic_ai import Agent  # noqa: E402
from pydantic_ai.messages import ThinkingPart  # noqa: E402

from code_puppy.agents._compaction import _strip_empty_thinking_parts  # noqa: E402
from code_puppy.model_factory import ModelFactory, make_model_settings  # noqa: E402


def build_agent(model_key: str) -> Agent:
    config = ModelFactory.load_config()
    model = ModelFactory.get_model(model_key, config)
    agent = Agent(
        model=model,
        model_settings=make_model_settings(model_key),
        system_prompt=(
            "You are a test agent. Always use the get_secret_number tool "
            "before answering anything about the secret number."
        ),
    )

    @agent.tool_plain
    def get_secret_number() -> int:
        """Return the secret number."""
        return 42

    return agent


def describe_parts(messages) -> None:
    for i, msg in enumerate(messages):
        kinds = []
        for p in msg.parts:
            label = type(p).__name__
            if isinstance(p, ThinkingPart):
                label += (
                    f"(content={p.content!r:.20}, "
                    f"sig={'yes' if p.signature else 'no'}, id={p.id})"
                )
            kinds.append(label)
        print(f"  [{i}] {type(msg).__name__}: {kinds}")


async def phase_a() -> None:
    print("=" * 72)
    print("PHASE A: /add_model shape (type=openai, Chat Completions route)")
    print("=" * 72)
    agent = build_agent("openai-gpt-5.6-luna")
    try:
        result = await agent.run("What is the secret number?")
        print("UNEXPECTED SUCCESS:", result.output)
    except Exception as exc:
        print(f"FAILED AS REPORTED IN #746 -> {type(exc).__name__}:")
        print(f"  {exc}")


async def phase_b() -> None:
    print("=" * 72)
    print("PHASE B: Responses API route + mainline empty-ThinkingPart strip")
    print("=" * 72)
    agent = build_agent("gpt-5.6-luna")

    # The code_puppy runtime streams responses; only the streaming path
    # surfaces the empty-summary reasoning item (with encrypted_content)
    # as ThinkingPart(content="", signature=...).
    async def noop_handler(ctx, stream):
        async for _event in stream:
            pass

    result = await agent.run(
        "What is the secret number? Use the tool.",
        event_stream_handler=noop_handler,
    )
    print("Turn 1 OK:", result.output[:80])
    history = result.all_messages()
    print("Turn 1 history parts:")
    describe_parts(history)

    stripped, n = _strip_empty_thinking_parts(history)
    print(f"\n_strip_empty_thinking_parts removed parts from {n} message(s)")
    print("History after strip (what mainline replays on turn 2):")
    describe_parts(stripped)

    try:
        result2 = await agent.run(
            "Now double the secret number using the tool result.",
            message_history=stripped,
            event_stream_handler=noop_handler,
        )
        print("\nTurn 2 UNEXPECTED SUCCESS:", result2.output[:80])
    except Exception as exc:
        print(f"\nTurn 2 FAILED AS REPORTED IN #853 -> {type(exc).__name__}:")
        print(f"  {exc}")
        if "reasoning" not in str(exc):
            traceback.print_exc()


async def main() -> None:
    if not os.environ.get("OPENAI_API_KEY"):
        sys.exit("OPENAI_API_KEY is not set")
    await phase_a()
    print()
    await phase_b()


if __name__ == "__main__":
    asyncio.run(main())
```

</details>

I'll leave detailed notes on the diff itself.

mpfaffenberger (commented):
Hey! Reviewed the whole diff with the live repro in hand and the core of this is right. The compaction guard (keep empty ThinkingParts that carry a signature) is exactly the bug, and moving GPT-5.6 onto native `openai_reasoning_context`/`openai_reasoning_mode` with the profile flags set is cleaner than the old `extra_body` delivery. I also checked that pydantic-ai merges a partial `profile=` on top of the provider's inferred one rather than replacing it, so the three flags in `_thinking_tags_profile` are additive and safe.

One real gap I'd like fixed before this lands: the reasoning-settings branch in `make_model_settings` still matches the config key for every type except `openai`. A `custom_openai_responses` entry whose key doesn't contain "gpt-5" (say key `luna-responses` with `name: gpt-5.6-luna`) gets zero reasoning config today, no effort, no context, no summary. I hit this exact shape while reproducing. I think the check should just always consider `model_config.get("name")` alongside the key instead of special-casing `model_type == "openai"`.

Two smaller things. `model_config["name"].lower()` in `get_model` will KeyError on an entry without `name`, `.get("name", "")` keeps it graceful. And the codex client now always prefers `completed_output_items` over the `response.completed` envelope, which is right for the store=false partial envelopes, but if an `output_item.done` event ever gets dropped mid-stream we'd now replace a complete envelope with a partial reconstruction. Merging by item id (or preferring whichever set is a superset) would make that unlosable. Related question, no wrong answer: unsigned empty ThinkingParts that still carry an `rs_` id get stripped, do we know gateways that replay by id without encrypted content won't trip the same validator?

Sort those and I'm happy to take this out of draft and merge. Thanks for sticking with this one and for the compat note about old histories, that saved me a question.

mpfaffenberger:
Merged! I took the liberty of pushing the review follow-ups directly to your branch (b21b4719) so we could land this today: the reasoning-settings branch now matches the underlying model name for every type (alias-keyed custom endpoint entries were still skipping reasoning config), plus the name KeyError nit and merge-by-id in the codex stream client.

Re-verified against the live API on your branch with the fixes: the signed empty ThinkingPart survives the history strip and the follow-up turn goes through cleanly, and the /add_model shape now routes straight to the Responses API on the first request. Full suite green locally (7633 passed).

Thanks for a great P/R and for patiently walking me past my "99% sure". This closes #746 too.

PR #855, opened 2026-08-23
Keep a repeated user prompt when the turn is genuinely new (#781)

### LINKED ISSUE #781: History merge drops a repeated user prompt, and the trailing-response pop then deletes the prior assistant answer

**Where:** `code_puppy/agents/_compaction.py` → `make_history_processor.history_processor`

**What.** The merge loop `continue`s unconditionally when `h in existing_hashes`, before the `i == last_idx` guard whose own comment says it exists precisely to stop short prompts like "yes"/"1" from being silently dropped — that guard only protects against `compacted_hashes`. `hash_message` deliberately ignores timestamps, so a new user turn whose text matches any earlier turn hashes identically and is skipped. The processor then returns a list rebuilt purely from `agent._message_history`, and the `while cleaned and isinstance(cleaned[-1], ModelResponse): cleaned.pop()` that follows removes the now-trailing assistant answer.

**Impact.** A user who answers two different confirmations with "yes" (or repeats "continue", "run the tests", etc.) has the second turn silently never sent. The model receives a history ending at the *earlier* matching turn with the last assistant response popped off, so it re-answers a stale question. The loss is permanent, not per-request.

**Suggested fix.** Move the newest-message exemption ahead of the dedup: `if h in existing_hashes and i != last_idx: continue`. Better, dedup the replayed prefix by position/identity against `history` rather than by content hash, so an intentionally repeated prompt can never be confused with the re-sent prefix. Add a regression test feeding the same `UserPromptPart` text twice at different positions — the existing suite only covers the `compacted_hashes` branch (`test_last_message_preserved_even_on_hash_collision`).

phanisaimunipalli:
Fixes #781.

`hash_message` is deliberately timestamp-independent, so answering two different questions with `yes` produces the same hash both times:

```
hash('yes') turn 1: 6f2245e1533438c3
hash('yes') turn 2: 6f2245e1533438c3
identical: True
```

The merge loop skipped the second one as a duplicate. Because that left history ending on a `ModelResponse`, the trailing-response pop below it then removed the earlier assistant answer too. Running the real merge on a two-turn conversation:

```
history before : [USER 'yes'] [ASSIST 'Deleting the branch now.']
incoming       : [USER 'yes'] [ASSIST 'Deleting the branch now.'] [USER 'yes']
messages merged: 0
history after  : [USER 'yes']

user turns the model can see : 1   (expected 2)
prior assistant answer kept  : False (expected True)
```

Three messages in, one out, and it is written back to `agent._message_history` permanently. The model never sees the second question and loses its own previous answer.

Worth noting the comment already sitting above the collision guard:

```python
# Always keep the newest message even on hash collision — short
# prompts like "yes"/"1" can collide and get silently dropped.
```

That is exactly right, but the unconditional `continue` one line earlier meant the guard was only ever reached for messages that were *not* already in `existing_hashes` — so it never protected the case it describes.

## The fix

A hash collision is not the same thing as a duplicate. pydantic-ai passes the history it already knows about plus this turn's new prompt, so when it hands us more messages than we are holding, the last one is a genuinely new turn regardless of what its hash matches.

```python
has_new_turn = len(messages) > len(history)
...
is_new_turn = i == last_idx and has_new_turn
if h in existing_hashes and not is_new_turn:
    continue
```

## Why not the one-liner in the issue

The issue suggests exempting the last index unconditionally (`if h in existing_hashes and i != last_idx`). That breaks `test_dedupes_by_hash`, because an identical resend with no new turn also ends on a colliding last message, and history would grow every time. The message-count check separates the two cases.

I verified all three states rather than assuming:

| variant | result |
|---|---|
| current `main` | new regression test fails (the bug) |
| unconditional last-index exemption | `test_dedupes_by_hash` fails |
| this PR | 27 passed |

## Testing

Two new tests in `tests/agents/test_compaction.py`: one for the repeated prompt, asserting both that the second turn survives and that the earlier assistant answer is not destroyed; and one guarding the opposite direction, that an identical resend with no new turn still dedupes.

`tests/agents/` is 405 passed. `ruff check` and `ruff format` clean at 0.15.

### DIFF
--- a/code_puppy/agents/_compaction.py
+++ b/code_puppy/agents/_compaction.py
@@ -331,13 +331,22 @@ def make_history_processor(agent: Any) -> Callable[..., Any]:
         existing_hashes = {hash_message(m) for m in history}
         messages_added = 0
         last_idx = len(messages) - 1
+        # pydantic-ai hands us the history it knows about plus this turn's new
+        # prompt, so being handed more messages than we hold means the last one
+        # is genuinely new. Its hash can still collide: hashes are
+        # timestamp-independent, so a repeated short prompt like "yes" or "1"
+        # hashes identically to the earlier one. Treating that as a duplicate
+        # drops the turn, and the trailing-ModelResponse pop below then removes
+        # the preceding assistant answer too.
+        has_new_turn = len(messages) > len(history)
         for i, msg in enumerate(messages):
             h = hash_message(msg)
-            if h in existing_hashes:
+            is_new_turn = i == last_idx and has_new_turn
+            if h in existing_hashes and not is_new_turn:
                 continue
             # Always keep the newest message even on hash collision — short
             # prompts like "yes"/"1" can collide and get silently dropped.
-            if i == last_idx or h not in compacted_hashes:
+            if is_new_turn or h not in compacted_hashes:
                 history.append(msg)
                 messages_added += 1
 
--- a/tests/agents/test_compaction.py
+++ b/tests/agents/test_compaction.py
@@ -517,6 +517,40 @@ class TestMakeHistoryProcessor:
             await make_history_processor(agent)(_ctx(), [_user_msg("yes")])
         assert len(agent._message_history) == 1
 
+    async def test_repeated_user_prompt_is_not_dropped_as_duplicate(self):
+        """A second "yes" answering a different question must survive.
+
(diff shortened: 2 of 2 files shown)


mpfaffenberger:
I've been meaning to fix this forever. Thank you.

PR #867, opened 2026-08-25
Keep MCP registry and user-plugin tree in sync with what actually loads

### LINKED ISSUE #866: MCP registry sync is add-only; file tools can write the user plugin import path
> This issue was posted by Grok Build using grok-4.6 on behalf of David.

`MCPManager.sync_from_config` adds and updates servers from `mcp_servers.json` but never removes names that left that file. `/mcp trust revoke` clears the trust store and tells the user to restart; a new manager still hydrates the old `mcp_registry.json` entries as `enabled=True`.

Separately, the default-agent file tools resolve `~` and can write `~/.code_puppy/plugins/<name>/register_callbacks.py`. That tree is imported at the next process start (`_load_user_plugins`) with no content hash.

Two smaller load-rule mismatches in the same area:

- `safety_permission_level` defaults to `medium`, but `shell_safety` (the plugin that implements that threshold) was only imported at `none`/`low`.
- `SurrogateUnpickler.find_class` still called `super().find_class` for any `dateutil.tz` / `pytz` name. Those packages re-export `os`/`sys`.

Expected:

- Registry names not in the merged json are unregistered on sync; revoke triggers that sync.
- File tools refuse mutations under `~/.code_puppy/plugins`.
- `shell_safety` loads at the default medium (skip only `high`/`critical`).
- tz-library names go through the surrogate path, not live `find_class`.

dsfaccini:
> This pull request was posted by Grok Build using grok-4.6 on behalf of David.

Fixes #866.

`sync_from_config` now unregisters registry entries whose names are no longer in the merged `mcp_servers.json`. `/mcp trust revoke` runs that sync instead of asking the user to restart.

File-tool write/replace/delete refuse paths inside `~/.code_puppy/plugins` (imported at startup, no trust hash).

`shell_safety` loads at the default `safety_permission_level=medium` (still skipped at `high`/`critical`). `SurrogateUnpickler` no longer resolves `dateutil.tz` / `pytz` names through `find_class`.

### DIFF
--- a/code_puppy/mcp_/manager.py
+++ b/code_puppy/mcp_/manager.py
@@ -161,10 +161,7 @@ class MCPManager:
         try:
             from code_puppy.config import load_mcp_server_configs
 
-            configs = load_mcp_server_configs()
-            if not configs:
-                logger.debug("No servers found in mcp_servers.json")
-                return
+            configs = load_mcp_server_configs() or {}
 
             synced_count = 0
             updated_count = 0
@@ -209,9 +206,23 @@ class MCPManager:
                     logger.warning(f"Failed to sync server '{name}' from config: {e}")
                     continue
 
-            if synced_count > 0 or updated_count > 0:
+            configured_names = {
+                name for name, conf in configs.items() if isinstance(conf, dict)
+            }
+            dropped_count = 0
+            for existing in list(self.registry.list_all()):
+                if existing.name not in configured_names:
+                    if self.remove_server(existing.id):
+                        dropped_count += 1
+                        logger.debug(
+                            "Dropped server no longer in mcp_servers.json: %s",
+                            existing.name,
+                        )
+
+            if synced_count > 0 or updated_count > 0 or dropped_count > 0:
                 logger.info(
-                    f"Synced {synced_count} new and updated {updated_count} servers from mcp_servers.json"
+                    f"Synced {synced_count} new, updated {updated_count}, "
+                    f"dropped {dropped_count} servers from mcp_servers.json"
                 )
 
         except Exception as e:
--- a/code_puppy/mcp_/manager.py
+++ b/code_puppy/mcp_/manager.py
@@ -161,7 +161,7 @@ class MCPManager:
         try:
             from code_puppy.config import load_mcp_server_configs
 
-            configs = load_mcp_server_configs() or {}
+            configs = load_mcp_server_configs(raise_on_error=True) or {}
 
             synced_count = 0
             updated_count = 0
--- a/code_puppy/tools/file_modifications.py
+++ b/code_puppy/tools/file_modifications.py
@@ -46,20 +46,59 @@ from code_puppy.tools.file_permission_state import (
 )
 
 
+def _split_existing(path: Path) -> tuple[Path, tuple[str, ...]]:
+    """Deepest existing ancestor of *path*, plus the missing trailing names."""
+    missing: list[str] = []
+    current = path
+    while True:
+        if current.exists():
+            return current.resolve(), tuple(reversed(missing))
+        parent = current.parent
+        if parent == current:
+            return current, tuple(reversed(missing))
+        missing.append(current.name)
+        current = parent
+
+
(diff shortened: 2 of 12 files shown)


code_puppy/tools/file_modifications.py
```diff
@@ -45,6 +46,36 @@
...
+    File tools must not plant ``register_callbacks.py`` there.
+    """
+    from code_puppy.plugins import USER_PLUGINS_DIR
+
+    try:
+        resolved = Path(resolve_path(file_path)).resolve()
+        root = Path(USER_PLUGINS_DIR).expanduser().resolve()
+    except (OSError, RuntimeError, ValueError):
+        return False
+    return resolved == root or root in resolved.parents
```
dsfaccini: **[bug]** `_is_user_plugin_tree_path` decides containment with `Path.resolve()` identity (`resolved == root or root in resolved.parents`). On a case-insensitive filesystem that comparison is case-sensitive, so a mixed-case spelling of the same inode is treated as outside the tree and the write proceeds. Verified on this machine (`os.path.samefile(Path.home()/".code_puppy", Path.home()/".CODE_PUPPY")` is True; `_is_user_plugin_tree_path("/Users/david/.CODE_PUPPY/plugins/evil/register_callbacks.py")` is False). A probe write via a case-variant of a tmp `USER_PLUGINS_DIR` returned `success: True` and the payload appeared at the canonical plugin path. That is the unhashed import tree `_load_user_plugins` execs on next start. `os.path.realpath` / `Path.resolve()` do not casefold on macOS, so following symlinks is not enough. `tests/tools/test_user_plugin_path_guard.py` only uses the exact-case tmp path, so this bypass is untested.

**Suggestion:** Treat "inside the user plugin tree" as an inode/FS check, not a string-equality check. Walk existing ancestors of the resolved target and use `os.path.samefile` against `USER_PLUGINS_DIR` (and against each of its existing parents when the `plugins` dir itself has not been created yet). For components that do not exist yet, compare with `str.casefold()` (macOS `os.path.normcase` is a no-op). Fail closed if canonicalization raises. Add a test that writes through a case-variant of `USER_PLUGINS_DIR` on a case-insensitive volume and expects refusal.

code_puppy/mcp_/manager.py
```diff
@@ -161,10 +161,7 @@ def sync_from_config(self) -> None:
         try:
             from code_puppy.config import load_mcp_server_configs
 
-            configs = load_mcp_server_configs()
-            if not configs:
-                logger.debug("No servers found in mcp_servers.json")
-                return
+            configs = load_mcp_server_configs() or {}
```
dsfaccini: **[bug]** The old `if not configs: return` is gone. `configs = load_mcp_server_configs() or {}` plus the new drop loop unregisters every registry name not in `configured_names`. That is correct when the merged JSON is genuinely empty (revoke, deleted servers). It is not correct when the loader fails: `load_mcp_server_configs` never raises — a missing `mcp_servers` key, bad JSON, or read error is `emit_error` + `return {}` (`code_puppy/config.py` `load_mcp_server_configs`, and `tests/test_config_full_coverage.py::TestMCPServerConfigs.test_bad_json`). Runtime: after registering `keepme` into an isolated `DATA_DIR`, patching `load_mcp_server_configs` to `{}` and calling `sync_from_config()` left `mcp_registry.json` as `{}`. The same `{}` is what a corrupt user-level file produces, so a mid-edit / merge-conflict / truncated `mcp_servers.json` on manager init, `/mcp trust revoke`, or `reload_mcp_servers` persists a wiped registry. `tests/mcp/test_registry_sync_drops_unconfigured.py` only mocks `return_value={}` and so encodes the wipe as the desired empty-config behavior without covering a failed load.

**Suggestion:** Distinguish "loaded, nothing configured" from "load failed". Simplest: make `load_mcp_server_configs` re-raise (or return a sentinel / `(configs, ok)` ) on parse/IO error of an existing user-level file, and skip the drop loop when the load did not succeed. Keep dropping on a successful empty mapping so revoke still works. Add a test that seeds the registry, points `MCP_SERVERS_FILE` at `{not json`, calls `sync_from_config()`, and asserts the existing name is still registered.

PR #868, opened 2026-08-26
feat(editing): Anthropic native text-editor adapter with exact-match-safe replace engine (Phases 1-3)

WSxDemise:
## What

Adds a hardened, three-phase Anthropic editor adapter to Code Puppy's file tools:

1. **Phase 1** — Exact-match safety for `_replace_in_file`: a write only happens
   when `old_str` matches **exactly once**. Zero matches or multiple matches now
   return a typed error (`match_not_found` / `ambiguous_match`) instead of
   guessing.
2. **Phase 2** — Undo-integrity and ambiguity-detection fixes across the
   generic edit tools (line-ending fidelity, capture/commit-based undo so a
   failed write can never poison history).
3. **Phase 3** — A native Anthropic `str_replace_based_edit_tool` adapter
   (`view` / `str_replace` / `create` / `insert`) that dispatches `str_replace`
   and `create` straight into the same hardened engine the portable tools use
   (no duplicated safety logic), and implements `view`/`insert` with their own
   guarded read/write paths.

Includes 13 commits of iterative review-and-fix rounds (Opus + GPT-5.6-Sol
review passes), a merge from `main` (resolving one additive conflict —
`main`'s new `_refuse_user_plugin_tree` guard on `_delete_file`), and two
additional fixes found by a post-merge review pass (see Testing).

## Why

Before this change, `_replace_in_file` used `str.replace(old, new, 1)`: if
`old_str` matched more than once, it silently edited the *first* occurrence
and reported `success: true`, even when that wasn't the location the caller
meant. There was no way to tell, from the tool's own output, that the wrong
spot got edited.

A 50-sample empirical baseline (real `claude-4-5-haiku` completions attempting
a realistic ambiguous-ordinal edit, replayed through both the old and new
`_replace_in_file` engines against identical input) confirmed this is not a
theoretical edge case:

| | Old engine | New engine |
|---|---|---|
| Correctly edited the intended location | 20/50 (40%) | 20/50 (40%) |
| **Silently edited the WRONG location, reported `success: true`** | **30/50 (60%)** | 0/50 |
| Safely refused with `ambiguous_match`, file untouched | 0/50 | **30/50 (60%)** |

The new engine converts every one of those 30 silent-corruption cases into a
safe, actionable failure — by construction (the code never writes unless
`match_count == 1`), not by chance. The Phase 3 native-editor tool inherits
this guarantee for free since it dispatches into the same engine rather than
reimplementing matching.

## How

- `code_puppy/tools/file_modifications.py`: match-count gate in
  `_replace_in_file`; capture/commit undo pattern; `_refuse_user_plugin_tree`
  guard on all write-capable entry points.
- `code_puppy/tools/anthropic_editor_tool.py`: new Anthropic native
  text-editor command dispatch (`view`/`str_replace`/`create`/`insert`),
  reusing the file-modification engine for `str_replace`/`create`.
- `code_puppy/tools/line_endings.py`: line-ending detection/conversion helpers
  shared by the above.

### Fixed during this PR's own merge/review pass

- Merge conflict  `_delete_file` resolved by keeping `main`'s new
  `_refuse_user_plugin_tree` guard (additive, no logic dropped).
- Removed a leftover `UndoManager().record_change(...)` call the merge
  introduced into `_delete_file` — it double-recorded undo history (and fired
  even when permission was later denied), duplicating
  `_delete_file_after_permission`'s own capture/commit pair. The async twin
  never had this call.
- Added the missing `_refuse_user_plugin_tree` guard to `_insert_into_file`
  (Anthropic native `insert` command), which previously called `_write_to_file`
  directly and bypassed the plugin-tree write protection every other
  write-capable path already has.

## Testing

- `ruff check .` / `ruff format --check .`: clean.
- Full suite: `pytest tests/ --ignore tests/integration` → **7592 passed, 14
  skipped, 0 failed**.
- Targeted Phase 1-3 tests (exact-match safety, undo integrity, native editor
  dispatch, plugin-tree guard): all passing.
- 50-sample empirical baseline (see Why) comparing old vs. new
  `_replace_in_file` against identical real-model-generated edit candidates,
  run against isolated temp files in throwaway git worktrees (not committed).
- Two-round `puppy-review` + `code-puppy-1-layer` LOCAL_GATE pass, both clean
  on the final diff.

Not yet done (tracked separately, not blocking this PR): a live end-to-end
smoke test against the real Anthropic API, and Phase 4 production-value
metrics (failure rate, token/latency/cache-hit tracking).

## Scope

Touches `code_puppy/tools/file_modifications.py`,
`code_puppy/tools/anthropic_editor_tool.py`, `code_puppy/tools/line_endings.py`,
and their test files. No unrelated files included — a local `.gitignore` /
`AGENTS.md` / `uv.lock` drift present before this work started was
deliberately excluded (stashed, not committed) as out of scope for this
change.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -757,7 +757,15 @@ def build_tool_probe_for_agent(agent: Any) -> Optional[Any]:
             toolsets=[],
         )
         register_tools_for_agent(
-            probe, agent.get_available_tools(), model_name=resolved_model_name
+            probe,
+            agent.get_available_tools(),
+            model_name=resolved_model_name,
+            # Thread the same fresh `models_config` used to resolve the
+            # model above, so this probe can't briefly disagree with the
+            # real agent build about the Anthropic-native-editor capability
+            # decision (see register_tools_for_agent's own docstring on why
+            # this parameter exists).
+            models_config=models_config,
         )
     except Exception:
         return None
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -688,6 +688,7 @@ def build_pydantic_agent(
         agent_tools,
         model_name=resolved_model_name,
         agent_name=logical_agent_name,
+        models_config=models_config,
     )
 
     existing_tool_names: Set[str] = set(getattr(probe_agent, "_tools", {}) or {})
@@ -703,6 +704,7 @@ def build_pydantic_agent(
         agent_tools,
         model_name=resolved_model_name,
         agent_name=logical_agent_name,
+        models_config=models_config,
     )
 
     agent.cur_model = model
--- a/code_puppy/agents/run_stats.py
+++ b/code_puppy/agents/run_stats.py
@@ -50,6 +50,7 @@ from pydantic_ai.messages import (
     ToolCallPartDelta,
 )
 
+from code_puppy.model_capabilities import NATIVE_EDITOR_TOOL_NAME
 from code_puppy.tools.subagent_context import is_subagent
 
 
@@ -335,6 +336,7 @@ _TOOLS_WITH_RENDERER = frozenset(
         "delete_file",
         "delete_snippet",
         "edit_file",
+        NATIVE_EDITOR_TOOL_NAME,
         "agent_run_shell_command",
         "ask_user_question",
         "activate_skill",
--- a/code_puppy/anthropic_native_editor_model.py
+++ b/code_puppy/anthropic_native_editor_model.py
@@ -1,0 +1,69 @@
+"""The Phase 3 model-layer seam: substitute Anthropic's native, client-executed
+text-editor tool declaration for one specific pydantic-ai tool, leaving every
+other tool's request/response handling completely unmodified.
+
+Why a subclass and not a generic HTTP body rewrite (per the plan's Phase 3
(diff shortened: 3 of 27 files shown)


WSxDemise:
Good morning, @thomwebb, @mpfaffenberger 
I deep dove how Anthropic trains their models and how we might be able to optimize our "make file changes" inside of CodePuppy. This underwent 9 rounds of adversarial review but needs more testing, so kept it draft on my side.

I would appreciate your thoughts and completely understand if "No, this is dumb" is the answer 😂 
I just know with Sonnet|Opus 5, we get a ton of "oops, I broke that file; let me try again" and I'm hoping this fixes that or at a minimum reduces it.

mpfaffenberger:
duplicate of #865

thomwebb (approved):
Reviewed against current upstream/main. No blocking correctness, regression, or security issues found. Focused local editor/capability suite passes (88 tests), and the PR CI checks are green.

Non-blocking follow-ups:
- Add a request-level Anthropic wire-format integration test proving the native text_editor_20250728 declaration is emitted without input_schema.
- The new anthropic_editor_tool.py is 615 lines; it is cohesive today but should be split if it grows further.

Also noting the existing repository comment that this PR may duplicate #865; please confirm before merging.

mpfaffenberger:
Closing since we merged the other one

PR #869, opened 2026-08-26
Add fallback_chain model type for quota-aware auto-degradation

dhruvkam:
## Summary

Adds quota-aware ordered model fallback and makes the requested catalog-provided default chain automatic:

```text
claude-4-8-opus-long (Opus 4.8 Long)
    -> quota/context exhaustion
claude-5-sonnet (Sonnet 5)
    -> quota/context exhaustion
gpt-5.6-luna (Luna)
```

When the runtime catalog contains those three keys, Code Puppy creates `default-fallback-chain` and selects it for a fresh installation with no explicit model selection.

- Permanently switches down the chain when quota/context capacity is exhausted.
- Does not treat ordinary 429 rate limits, 5xx errors, auth failures, or unrelated errors as fallback triggers.
- Emits a warning on every tier transition.
- Checks common provider signatures, structured `ModelHTTPError.body` payloads, and wrapped exceptions.
- Supports provider-specific additions through `budget_exhausted_patterns`.
- Preserves model-specific settings for each child, including Opus/Sonnet thinking settings and Luna reasoning settings.
- Supports `codex` as a configured Responses-compatible model type.
- Validates unavailable child models during factory construction instead of failing later during a request.
- Adds README documentation and focused tests.

This remains distinct from `RoundRobinModel` (load distribution) and the existing same-model transient streaming retry behavior.

## Validation

- `ruff check .` — passed
- `ruff format --check .` — passed
- `git diff --check` — passed
- Focused default/factory/config/fallback suite: **249 passed**
- New fallback module: **100% coverage**
- Exact quota flow: Opus quota exhausted -> Sonnet quota exhausted -> Luna succeeds -> subsequent request remains on Luna
- Factory smoke: sanitized catalog resolves all three children, including Luna `codex` -> `OpenAIResponsesModel`, with no network calls
- Full local suite: **7522 passed, 32 skipped**

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,55 @@ Then just use /model and tab to select your round-robin model!
 
 The `rotate_every` parameter controls how many requests are made to each model before rotating to the next one. In this example, the round-robin model will use each Qwen model for 5 consecutive requests before moving to the next model in the sequence.
 
+## Fallback Chain (Auto-Downgrade on Quota/Context Exhaustion)
+
+Round-robin distributes load across *equivalent* models. A **fallback chain** is different: it's an ordered preference list (e.g. a large/expensive model, then a medium one, then a free/unmetered one), and Code Puppy only moves down the list when the current model's *budget* is actually used up -- a provider quota fully exhausted, or the conversation grown too large for that model's context window. Ordinary transient errors (a 429 rate-limit, a 5xx blip) are already retried against the *same* model with backoff, so they don't trigger a downgrade.
+
+The switch is **one-directional and sticky**: once a model is marked exhausted, Code Puppy never tries it again for the rest of the process -- it just keeps working from wherever it landed. Restart Code Puppy (or reselect a model with `/model`) to reset back to the top of the chain.
+
+### Configuration
+
+```json
+{
+  "gpt-huge": {
+    "type": "openai",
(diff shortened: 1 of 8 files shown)


mpfaffenberger:
Hey there! Thanks for the P/R, there's clearly real thought in here, especially the distinction from round robin and the quota-vs-transient split.

The blocker is that Pydantic AI already ships this: https://ai.pydantic.dev/models/fallback/ supports custom `fallback_on` predicates (sync or async, exception or response based), and it handles streaming and instrumentation correctly. The only genuinely new behavior here is the sticky "never go back up the chain" state, and that's a small wrapper around `FallbackModel` rather than a parallel implementation. If sticky fallback feels generally useful, I'd love to see it proposed upstream instead.

The other thing I can't merge is the hardcoded default chain. Those three model names are your deployment's catalog, and having core silently make that chain the default model for anyone whose catalog matches isn't something the public project can carry. The good news is the plugin system covers this completely: `register_model_type` for the `fallback_chain` type, `load_models_config` to inject the chain, and `model_select` for defaulting. You could ship the whole feature as an internal plugin today with zero changes to core.

Happy to take another look if you rebuild it on that path. The `codex` model type change should be its own P/R either way.

PR #886, opened 2026-08-29
feat(agents): add transform_mcp_toolsets extension seam to BaseAgent

WSxDemise:
## What
Adds a minimal, generic extension seam so a `BaseAgent` subclass can
post-process its resolved MCP toolsets before the pydantic-ai agent is
constructed: `BaseAgent.transform_mcp_toolsets(toolsets)`, a no-op identity
by default.

## Why
An internal Confluence search agent needs to wrap its MCP toolset to compact
oversized search results before they enter the model's context (full page
bodies were causing heavy rate-limit throttling). There was no supported way
for a `BaseAgent` subclass to intercept/replace its own MCP toolsets, only
the module-level plugin callback registry. This seam is intentionally
Confluence-free and Walmart-free -- it's a generic hook any subclass can use.

## How
- `BaseAgent.transform_mcp_toolsets()`: no-op identity default, documented
  extension point. Explicitly documents that fail-open behavior makes this
  seam unsuitable for security-sensitive gating.
- `_builder.build_pydantic_agent`: calls it exactly once, after MCP
  collision filtering and before the final `pydantic_ai.Agent` is built.
  The result drives both agent construction and `agent._mcp_servers`
  lifecycle tracking. A raising override or a non-list return fails open
  (warns, falls back to the original toolsets) rather than crashing.

## Testing
- `pytest tests/agents/test_transform_mcp_toolsets.py` -- 8/8 passing
  (default identity, override-invoked-once via the post-filter list,
  raising-override fail-open, non-list-return fail-open).
- `pytest tests/agents/` -- 443/443 passing (no regression to unrelated
  agents/builder behavior).
- `ruff check` / `ruff format --check` clean on all touched files.

## Scope
Only `code_puppy/agents/base_agent.py`, `code_puppy/agents/_builder.py`, and
one new test file. No new dependencies, no unrelated refactors. This PR has
no in-repo caller yet -- an internal downstream Confluence-agent PR is the
consumer; happy to discuss whether a `callbacks.py` phase would fit better
than a subclass hook if that's the project's preferred extension pattern.

### DIFF
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -20,12 +20,12 @@ from pydantic_ai.capabilities import ProcessHistory
 
 from code_puppy.agents._compaction import make_history_processor
 from code_puppy.agents._model_message_transform import build_model_message_transform
+from code_puppy.agents._subagent_recursion import build_subagent_recursion_guard
 from code_puppy.agents._output_limits import (
     build_response_clamp,
     build_tool_output_limits,
 )
 from code_puppy.agents._steer_processor import make_steer_history_processor
-from code_puppy.agents._subagent_recursion import build_subagent_recursion_guard
 from code_puppy.agents.event_stream_handler import event_stream_handler
 from code_puppy.callbacks import (
     on_pre_mcp_autostart,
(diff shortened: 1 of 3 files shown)


WSxDemise:
Superseded by #896 -- identical commits and diff, recreated from a properly named branch (`feature/transform-mcp-toolsets-seam`). The original branch name referenced an internal downstream consumer, which was misleading for a generic seam in this repo, and GitHub does not allow re-pointing an existing PR's head branch. Continuing review in #896.

PR #900, opened 2026-09-01
fix: reject tool-call JSON repairs that change payload shape

WSxDemise:
## What

Reject `json_repair` output when repairing malformed tool-call arguments changes the top-level JSON shape away from an object.

## Why

Brace-dense content with one escaping error can make `json_repair` silently reinterpret a single tool-call object as a list of unrelated fragments. Accepting that valid-but-corrupted JSON can execute a tool with the wrong arguments instead of surfacing validation failure.

## How

- Strictly parse valid JSON first so it bypasses repair heuristics unchanged.
- Repair only malformed JSON and accept the result only when it parses to an object.
- Return the original malformed input on repair failure or shape mismatch so normal validation can fail loudly.
- Reuse the guarded helper from the current pydantic-ai validation patch.

> [!IMPORTANT]
> **Compatibility with [PR #841](https://github.com/mpfaffenberger/code_puppy/pull/841):** #841 refactors tool-call JSON repair into a capability, but its current implementation still accepts any changed `json_repair` result without verifying that the top-level value remains an object. It therefore does not fix this specific shape-corruption bug. If #841 supersedes this patch, both its capability path and guest-agent fallback must preserve this PR's strict-first, object-shape guard.

## Testing

- `ruff check code_puppy/pydantic_patches.py tests/test_pydantic_patches.py`
- `ruff format --check code_puppy/pydantic_patches.py tests/test_pydantic_patches.py`
- `pytest tests/test_pydantic_patches.py -v` (27 passed)

## Scope

Focused on tool-call JSON repair safety and regression coverage. No dependency or lockfile changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ code_puppy/bundled_skills/
 
 .state/
 
+# Local agent context
+.context/
+
 # Local git worktrees
 .worktrees/
 
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,6 @@ code_puppy/bundled_skills/
 
 .state/
 
-# Local agent context
 .context/
 
 # Local git worktrees
--- a/.gitignore
+++ b/.gitignore
@@ -49,8 +49,6 @@ code_puppy/bundled_skills/
 
 .state/
 
-.context/
-
 # Local git worktrees
 .worktrees/
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,13 @@
 # Contributing to Code Puppy
 
+## Agent context contract
+
+- `~/.code_puppy/AGENTS.md` and its indexed global context apply alongside this project router; these project rules supplement rather than replace them.
+- Search `~/.code_puppy/.context/INDEX.md` narrowly for cross-project expectations, tooling, credentials, access, Walmart workflows, and known global topics.
+- Use `.context/INDEX.md` separately for durable project-local facts and decisions.
+- Context is a discovery aid, never a substitute for checking current code, tests, configuration, or live systems.
+- Never store secrets, tokens, private content, or unnecessary PII in either context tree.
+
 > **Golden rule:** nearly all new functionality should be a **plugin** in the
 > `code_puppy_core_plugins` repository that hooks into core via
 > `code_puppy/callbacks.py`. Don't edit `code_puppy/command_line/`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -165,7 +165,7 @@ quickstart). Rules for new user-facing output (PUP-473):
 - **Model-facing system prompts are OUT of scope** — translating them changes
   LLM behavior. Don't run them through the i18n seam.
 - Add extraction behind the pseudolocale/coverage CI gate
-  (`tests/i18n/test_i18n_coverage.py`): every translated key must exist in the
+  (`tests/i18n/test_i18n_audit.py`): every translated key must exist in the
   `en-US` source, and a pseudolocale run must emit only bracketed text.
 
 ## Rules
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,13 +1,5 @@
 # Contributing to Code Puppy
 
-## Agent context contract
-
-- `~/.code_puppy/AGENTS.md` and its indexed global context apply alongside this project router; these project rules supplement rather than replace them.
-- Search `~/.code_puppy/.context/INDEX.md` narrowly for cross-project expectations, tooling, credentials, access, Walmart workflows, and known global topics.
-- Use `.context/INDEX.md` separately for durable project-local facts and decisions.
-- Context is a discovery aid, never a substitute for checking current code, tests, configuration, or live systems.
-- Never store secrets, tokens, private content, or unnecessary PII in either context tree.
-
 > **Golden rule:** nearly all new functionality should be a **plugin** in the
 > `code_puppy_core_plugins` repository that hooks into core via
 > `code_puppy/callbacks.py`. Don't edit `code_puppy/command_line/`.
--- a/code_puppy/agents/_builder.py
+++ b/code_puppy/agents/_builder.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 
 import uuid
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
 
 from pydantic_ai import Agent as PydanticAgent
 from pydantic_ai.capabilities import ProcessHistory
@@ -42,6 +42,9 @@ from code_puppy.config import (
 from code_puppy.mcp_ import get_mcp_manager
 from code_puppy.messaging import emit_error, emit_info, emit_warning
 from code_puppy.model_factory import ModelFactory, make_model_settings
+
+if TYPE_CHECKING:
+    from code_puppy.model_utils import PreparedPrompt
 
 _AGENT_RULE_FILES = ("AGENTS.md", "AGENT.md", "agents.md", "agent.md")
 _CODE_PUPPY_DIR = ".code_puppy"
@@ -572,21 +575,19 @@ def _agent_exposes_tool(agent: Any, tool_name: str) -> bool:
         return False
 
 
-def _assemble_instructions(agent: Any, resolved_model_name: str) -> str:
-    """Compose full system prompt + puppy rules + extended-thinking note."""
+def _assemble_instructions(agent: Any, resolved_model_name: str) -> PreparedPrompt:
+    """Compose full system prompt + puppy rules.
+
+    Returns the model-prepared prompt: ``instructions`` for the agent plus any
+    standing ``system_prompt`` a plugin wants emitted as its own
+    ``SystemPromptPart`` ahead of them.
+    """
     from code_puppy.model_utils import prepare_prompt_for_model
-    from code_puppy.tools import (
-        EXTENDED_THINKING_PROMPT_NOTE,
-        has_extended_thinking_active,
-    )
 
     instructions = agent.get_full_system_prompt()
     puppy_rules = load_puppy_rules()
     if puppy_rules:
         instructions += f"\n{puppy_rules}"
-
-    if has_extended_thinking_active(resolved_model_name):
-        instructions += EXTENDED_THINKING_PROMPT_NOTE
 
     if _is_gpt_5_6_family(resolved_model_name):
         if _agent_exposes_tool(agent, "invoke_agent"):
@@ -594,10 +595,9 @@ def _assemble_instructions(agent: Any, resolved_model_name: str) -> str:
         if _agent_exposes_tool(agent, "agent_run_shell_command"):
             instructions += _GPT_5_6_RUN_SHELL_COMMAND_GUARD_TEXT
 
-    prepared = prepare_prompt_for_model(
+    return prepare_prompt_for_model(
         agent.get_model_name(), instructions, "", prepend_system_to_user=False
     )
-    return prepared.instructions
 
 
 def build_pydantic_agent(
@@ -637,7 +637,7 @@ def build_pydantic_agent(
         message_group,
         agent_name=getattr(agent, "name", None),
     )
-    instructions = _assemble_instructions(agent, resolved_model_name)
+    prepared = _assemble_instructions(agent, resolved_model_name)
     mcp_servers = load_mcp_servers(agent_name=getattr(agent, "name", None))
     model_settings = make_model_settings(
         resolved_model_name,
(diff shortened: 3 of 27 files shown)


AndrewTilson:
## Code Review — PR #900

Reviewed at `bf97f203` against `main` (`91c99c4e`). Static analysis + targeted non-interactive test runs only; no CLI/TUI session was launched.

**Overall:** the core fix is correct, well-motivated, and well-tested. I reproduced the exact bug this prevents and confirmed the guard works. Findings below are **Medium and Low only — no High/Critical.**

---

### Summary

`_repair_tool_call_json()` replaces the previous unconditional `json_repair.repair_json()` call in `patch_tool_call_json_repair`. New order of operations:

1. Strict `json.loads()` first — valid JSON bypasses repair heuristics entirely.
2. Repair only on `JSONDecodeError`.
3. Accept the repair **only** if it parses back to a `dict`.
4. Otherwise return the original string so normal validation fails loudly.

### Verified

I confirmed the motivating bug is real, not theoretical:

```
'[{"a": 1},]'  → json_repair returns list  → guard REJECTS, original preserved 
'{"a": 1, "b": 2,}' → repairs to dict      → accepted 
"{'a': 1}"          → repairs to dict      → accepted 
'not json at all'   → repair raises        → original preserved 
'[1,2,3]' / '"hello"' / 'null' (already-valid non-objects) → untouched 
```

The strict-parse-first ordering is a genuine improvement beyond the shape guard — previously *valid* JSON was still fed through `json_repair` and could be rewritten (e.g. `NaN`/`Infinity` literals, which Python's `json` accepts, were being silently normalized). Those now pass through unchanged.

Also confirmed locally:
- `pytest tests/test_pydantic_patches.py` → **27 passed** (also passes with `-p no:randomly`, so no ordering dependence)
- `ruff check` → clean; `ruff format --check` → clean
- CI: all 3 checks green; mergeable / CLEAN

Fail-closed handling of non-`JSONDecodeError` exceptions from the strict parse is a nice touch and correctly preserves the module's "patches NEVER raise" contract.

---

### Findings

**1. [Medium] Success-path `logger.warning` writes to stderr on every repair — TUI corruption risk**

`code_puppy` configures no logging handlers anywhere, so Python's `logging.lastResort` handler is active and emits `WARNING`+ straight to `stderr`. I verified this:

```python
buf = io.StringIO()
with contextlib.redirect_stderr(buf):
    _repair_tool_call_json('{"a": 1,}')
# STDERR: 'json_repair modified malformed tool-call JSON arguments\n'
# root handlers: []
```

This is a behavior change: the old code repaired silently. The pre-existing `logger.*` calls in this module all fire once at patch-apply time during startup (inside the splash screen's stdout/stderr capture). These new ones fire **per tool call, mid-session**, after the splash has released the streams — so raw log lines land directly on the terminal and can corrupt TUI rendering.

Suggestion: drop the success-path log to `logger.debug()`. The rejection-path `logger.warning()` is genuinely worth keeping loud — it's rare and indicates real corruption.

**2. [Medium] The dict guard doesn't catch the most common corruption mode — truncated arguments**

The guard checks *top-level shape* only. Truncated tool calls (token-limit cutoffs) repair to a perfectly valid `dict` and sail through:

```python
raw = '{"file_path": "app.py", "content": "import os\n...\ndef cleanup():\n    shutil.rmtree('
# → repairs to dict, guard ACCEPTS
# → content silently ends at 'shutil.rmtree('
```

That's a truncated file write executing as if complete. This isn't a regression — same behavior as before the PR — but the PR title says "change payload shape," and shape is a narrower property than payload integrity. Worth either scoping the claim in the docstring or filing follow-up. Not a blocker for this PR.

**3. [Low] Second `json_repair` call site left unguarded-by-the-same-helper**

`_pydantic_tool_helpers._tool_args_for_pre_tool_call()` independently calls `json.loads(json_repair.repair_json(call_args))`. It happens to have its own `isinstance(parsed, dict)` check, so it isn't currently vulnerable — but it lacks the strict-parse-first step, so valid JSON still gets run through repair heuristics on that path. Consider routing it through the new helper so there's one implementation of this policy.

**4. [Low] Test monkeypatches the global `json` module**

`test_strict_parse_failure_returns_original` does `monkeypatch.setattr(pydantic_patches.json, "loads", parse_failure)`. `pydantic_patches.json` **is** the shared stdlib module object (verified: `pydantic_patches.json is json` → `True`), so this replaces `json.loads` process-wide for the duration. `monkeypatch` reverts it and the suite passes, so it's not currently breaking anything — but it's fragile under parallel/xdist execution. A module-local alias or a genuinely-failing input would be safer.

**5. [Low] Helper placement**

`_repair_tool_call_json` sits at the top of `pydantic_patches.py`, wedged between the imports and the `_LOUD_FAILURES` infrastructure block. The repo already has `code_puppy/_pydantic_tool_helpers.py`, whose docstring is literally *"Small, dependency-light helpers for Code Puppy's pydantic-ai tool patch"* — and which already hosts the sibling `_tool_args_for_pre_tool_call`. That looks like the natural home, and would make fix #3 fall out for free.

**6. [Low] Branch history — please squash-merge**

Intermediate commits add an unrelated `AGENTS.md` + `.gitignore` change and then revert it. The net diff is clean (2 files, exactly the intended change), so this is purely cosmetic — but a squash-merge keeps the reverted churn off `main`.

---

### Repository & Placement

**Repo fit:  Correct.** This patches the pydantic-ai monkeypatch layer in the core CLI runtime (`code_puppy/pydantic_patches.py`). Not plugin code, so it does not belong in `code_puppy_core_plugins`. No proprietary/internal dependencies — correctly public.

**Pattern consistency: ** Naming (`_`-prefixed private helper), the never-raise contract, `logger`-over-`messaging` convention, and type hints all match surrounding code. Only the file-placement nit in #5.

---

### Coordination with #841

#841 (still open) refactors this same repair logic into a `ToolCallJsonRepair` capability, and its current implementation accepts any changed `json_repair` result without a shape check. **Whichever lands second must carry this guard forward**, or the fix silently regresses. Flagging for whoever sequences these.

---

### Verdict

No High or Critical findings. Findings #1 and #2 are worth a look before merge — #1 is a one-line change (`warning` → `debug` on the success path) and would be nice to fold in. Everything else is non-blocking follow-up.

Nice, tightly-scoped fix with real regression coverage.

WSxDemise:

Thanks for the thorough review and verification, @AndrewTilson! Responses to each item below:

| # | Item | Status / Verdict |
|---|---|---|
| 1 | **Success-path `logger.warning` writes to stderr** | Fixed in `8167b345` — lowered success-path log to `logger.debug()`; kept rejection warning loud. |
| 2 | **Truncated tool-call arguments** | Out of scope for this PR. As noted, top-level object shape rejection is an intentional, scoped boundary for this patch. Truncated argument handling / streaming boundary validation is tracked separately. |
| 3 | **Second `json_repair` call site in `_pydantic_tool_helpers`** | Deferred to follow-up / refactor. That helper retains its own `isinstance(parsed, dict)` guard and is refactored alongside capability migration. |
| 4 | **Test monkeypatching global `json`** | Deferred to follow-up test hardening. Isolated to the test process under standard runner cleanup. |
| 5 | **Helper placement in `_pydantic_tool_helpers`** | Out of scope for this minimal patch. Placing `_repair_tool_call_json` alongside the patch function keeps diff blast radius minimal; natural candidate for consolidation during #841. |
| 6 | **Squash-merge recommendation** | Agreed. Will squash-merge when landing to keep `main` history clean. |
| — | **Coordination with #841** | Noted. Guard logic will be carried forward during capability rebase/landing. |

---
C.C. @wblake2 for review

AndrewTilson (approved):
## Review: reject unsafe tool-call JSON repairs

Nice, focused bug fix. I reproduced the failure mode locally rather than taking the description's word for it, and the core claim holds up. **Approving with comments** — nothing here blocks merge, but there are a few follow-ups worth deciding on explicitly.

---

### Verified the bug is real (not theoretical)

Ran the `SHATTERING_MALFORMED_JSON` payload through `json_repair==0.54.2` (the pinned version) on `main`:

```
old repair type: list
old repair value: [{"file_path": "demo.py", "content": "print(f\"wrote {n_rows:,} rows\")\nprint(f", "name": 30}, {"count": 10}]
```

That is genuinely nasty. One escaping error turns a single `edit`-style tool call into a **list of unrelated fragments**, and `{name:<30}` / `{count:>10,}` — ordinary Python f-string format specs inside file content — get reinterpreted as JSON keys `"name": 30` and `"count": 10`. On `main` that valid-but-wrong JSON sails into validation. The guard correctly returns `REJECTED-shape` and preserves the original so validation fails loudly.

Confirmed passing locally: `27 passed`, `ruff check` clean, `ruff format --check` clean. Claims in the description check out.

### Strict-parse-first is a real win, not just a reordering

Bonus nobody claimed: skipping repair on already-valid JSON roughly **halves** hot-path cost for large tool args (2k-line file content, 200 iters):

```
main (always repair): 0.012s
PR   (strict first) : 0.005s
```

Every valid tool call previously paid for a full repair pass. Good.

---

### 1. The shape guard is necessary but not sufficient (main comment)

The guard only rejects `dict -> non-dict`. Repairs that stay a `dict` but **silently mangle content** are still accepted. Same escaping-bug family, verified with the pinned version:

```
truncated_stream   -> ACCEPTED  '{"file_path": "a.py", "content": "def f():\\n    return {"}'
quote_in_value     -> ACCEPTED  '{"cmd": "echo \\"hi\\" > f.txt"}'
dropped_key        -> ACCEPTED  '{"a": 1, "b": "x", "c": 2}'
```

The `truncated_stream` case is the one I'd worry about: a truncated write gets "repaired" into a **syntactically valid tool call that writes a truncated file**. Shape is preserved, so this guard waves it through, and it fails silently rather than loudly — exactly the class of outcome the PR is trying to prevent.

Not asking you to solve content-fidelity heuristics here (that way lies madness, and YAGNI applies until there's a real report). But the docstring and the PR framing read as "repairs are safe now," which is stronger than what's actually enforced. Suggest narrowing the claim:

```python
"""Repair malformed object JSON, rejecting repairs that change the top-level shape.

Shape preservation is a necessary but NOT sufficient safety check: a repair
that stays an object can still alter values (e.g. truncated content). This
guards against catastrophic reinterpretation, not content fidelity.
"""
```

Cheap, and it stops the next reader from over-trusting the guard.

### 2. DRY: three other `repair_json` call sites are still unguarded

The new helper is private to `pydantic_patches.py`, but it isn't the only place doing this dance:

- `code_puppy/_pydantic_tool_helpers.py:40` — `_tool_args_for_pre_tool_call`
- `code_puppy/tools/file_modifications.py:1130`
- `code_puppy/tools/file_modifications.py:1268` — `_try_json_repair`

The first one matters most: it's **the same tool-call path**, feeding args to `pre_tool_call` callbacks. It does check `isinstance(parsed, dict)`, but on mismatch it falls through to `{"raw": call_args}` — so callbacks and validation can now form *different* views of the same malformed call. Worth a follow-up issue to route all four through one shared guarded helper. Three near-identical `json.loads(json_repair.repair_json(x))` snippets with subtly different failure semantics is the classic setup for fixing the bug in one place and leaving it live in three.

### 3. Test hygiene: `test_strict_parse_failure_returns_original` patches stdlib globally

```python
monkeypatch.setattr(pydantic_patches.json, "loads", parse_failure)
```

`pydantic_patches.json` **is** the stdlib `json` module object, so this replaces `json.loads` process-wide for the duration. I confirmed the leak with a probe test:

```
LEAK: unrelated stdlib json.loads now raises: strict parser rejected input
```

`monkeypatch` reverts it afterward, so the suite passes today — but any concurrently-executing or fixture-level code touching `json.loads` during that window breaks in a way that's maddening to debug (and this bites under `pytest-xdist`). Explicit is better than implicit — patch the name in the module's namespace instead:

```python
monkeypatch.setattr(pydantic_patches, "json", SimpleNamespace(loads=parse_failure, JSONDecodeError=json.JSONDecodeError))
```

...or restructure so the helper takes an injected parser. Same coverage, no global blast radius.

### 4. Minor: blanket `except Exception` loses diagnostics

```python
except Exception:
    return raw
```

The non-dict path logs a nice actionable warning; an actual `json_repair` crash logs *nothing at all*. Given this module's stated philosophy — "idempotent and NEVER raises, but failures are not silent" — a `logger.debug("json_repair raised: %r", exc)` would fit the house style. Errors should never pass silently.

### 5. Housekeeping

- Branch is **27 commits behind `main`**. Worth a rebase before merge — `pydantic_patches.py` has moved recently (`8763fbe Fix ruff format violations in pydantic_patches.py`).
- **#841 is still open** and refactors this same code into a capability. Whichever lands second inherits the conflict; since #841 reportedly lacks this shape check, please make sure the guard survives that merge rather than getting refactored away.
- The `# noqa: F401  (optional-dependency gate)` on the outer import is a good catch — keeping the import as an availability probe while the helper does its own lazy import is the right call.

---

### Verdict

**Approve.** Correct fix for a real and well-demonstrated bug, with genuinely good regression coverage (the `SHATTERING_MALFORMED_JSON` fixture is a great artifact — it asserts `json_repair`'s bad behavior *first*, so the test stays honest if upstream ever fixes it). Item 3 is the only one I'd nudge to fix in this PR since it's a two-line change; 1 and 2 are fine as follow-ups.

*Reviewed by Luna — claims verified by execution against pinned `json_repair==0.54.2`, not by eyeballing the diff.*

PR #909, opened 2026-09-03
fix(mcp): one dead MCP server no longer aborts the whole run

breedx:
One MCP server that fails to start kills the entire run — no model call, no answer.

`_run_agent_task_body` enters every `agent_run_context` into one shared `AsyncExitStack`. A dead server raises from `__aenter__`, the stack unwinds, and `_do_run` on the next line never executes.

**Fix:** enter each context in its own `try`. A failure is reported by name; the run continues without that server's tools.

```python
 async with AsyncExitStack() as stack:
     for cm in run_ctxs:
-        await stack.enter_async_context(cm)
+        try:
+            await stack.enter_async_context(cm)
+        except Exception as exc:
+            _note_unusable_run_context(cm, exc, group_id)
     return await _do_run(prompt_payload)
```

Full suite: **7680 passed**. `ruff check` + `format --check` clean.

---

<details>
<summary>Why it looks handled but isn't</summary>

The `except* McpError` handler downstream prints:

> An MCP server failed during this run. Run `/mcp logs <name>` for details, or unbind it via `/agents → B`.

That reads like a warning attached to your answer. There is no answer — the coroutine falls off the end and returns `None`. The message makes a total failure look like a partial one, which is why this took a while to spot.

</details>

<details>
<summary>How it was found</summary>

A connector whose `uvx` dependency resolve failed. The process **existed**, so every status flag said RUNNING and pre-run health checks passed cleanly. It only failed at context-entry time — by which point the all-or-nothing stack had already made it fatal.

No amount of pre-run probing catches this; the failure happens after every check has run. That's why the fix has to be at the entry seam rather than in a health filter.

</details>

<details>
<summary>Design note</summary>

A connector is a **capability, not a precondition**. Losing one should cost its tools, not the answer — the agent still has its built-in file and shell tools, which is most of what it does.

Two details worth noting:

- **Servers that did enter are still owned by the stack**, so a partial failure can't leak a subprocess.
- **A server failing early no longer discards the healthy ones after it.** Previously the loop stopped at the first failure, so connectors later in the list were silently never entered either.

</details>

<details>
<summary>Tests</summary>

`tests/agents/test_run_context_degrades.py`:

- the **old shape** raises, never reaches the run body, and stops entering at the first failure
- the **new shape** reaches it, keeps every healthy connector *including those after the failure*, survives all servers failing, and still unwinds what it entered

The single suite failure (`test_copilot_auth_model::test_ghe_host_is_passed_through`) is **pre-existing** — verified to reproduce identically on an unmodified checkout of `main`.

</details>

### DIFF
--- a/code_puppy/agents/_runtime.py
+++ b/code_puppy/agents/_runtime.py
@@ -893,6 +893,34 @@ async def _run_with_mcp_impl(
         with executing_agent_context(agent):
             return await _run_agent_task_body()
 
+    def _note_unusable_run_context(cm: Any, exc: BaseException, group_id: str) -> None:
+        """Report a run context that could not be entered, and carry on.
+
+        Named rather than inlined so the message stays one thing: what was
+        lost, and where to look. `/mcp logs` is the only place the real cause
+        lives -- this layer knows a context manager raised, never why.
+
+        The name is best-effort: `run_ctxs` holds whatever plugins returned,
+        so a `toolset` attribute, a `name`, or the repr are all plausible.
+        Reporting `<unknown>` beats hiding the failure over a missing label.
+        """
+        name = (
+            getattr(cm, "name", None)
+            or getattr(getattr(cm, "toolset", None), "name", None)
+            or type(cm).__name__
+        )
+        emit_warning(
+            f"MCP server '{name}' could not start for this run - continuing "
+            f"without its tools. Run [cyan]/mcp logs {name}[/cyan] for the "
+            "reason.",
+            group_id=group_id,
+        )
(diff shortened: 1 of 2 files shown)


breedx:
Closing this outdated implementation. Guarding the outer run-context loop does not adequately cover the actual MCP toolset entry and recovery lifecycle. This needs a separately scoped connector-lifecycle fix with end-to-end failure-containment, recovery and cleanup tests, rather than retaining this patch as the proposed solution. The underlying MCP resilience issue remains open; identity/prompt work is separate.

PR #912, opened 2026-09-05
Fix/termflow table row separators

Prathap-P:
termflow's `render_table_complete` draws one horizontal rule (after the header) but nothing between body rows, so wide/dense tables are hard to scan row-by-row.

Adds `patch_termflow_table_row_separators()` to `code_puppy/pydantic_patches.py`, alongside the existing `patch_termflow_clipboard` / `patch_termflow_code_padding`, using the same pattern: swap in an otherwise-identical `render_table_complete` that also draws a separator between consecutive body rows (not redundantly before the bottom border). Registered in `_ALL_PATCHES`.

No termflow files are modified — this only replaces the function reference at runtime, same technique as its two sibling patches.

### DIFF
--- a/code_puppy/pydantic_patches.py
+++ b/code_puppy/pydantic_patches.py
@@ -460,6 +460,76 @@ def patch_termflow_code_padding() -> bool:
         )
 
 
+def _render_table_complete_with_row_rules(header, rows, alignments, width, margin, style):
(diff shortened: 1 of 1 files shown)


Prathap-P:
Before:

before

After:

Screenshot 2026-09-05 at 7 34 37 PM

PR #913, opened 2026-09-06
fix(gemini): drop trailing model turns so /steer works on Gemini 3.x

WSxDemise:
## The problem

Steering a Gemini 3.x model (`/steer`, Ctrl+T) kills the run with a 400 from Google:

```
Gemini API error 400: {"error":{"code":400,
"message":"Requests ending with a model turn are not supported.","status":"INVALID_ARGUMENT"}}
```

Gemini refuses any request whose conversation ends on the model's own turn — the last
message has to come from the user. Two normal situations leave the history in exactly
that state:

- **Steering mid-turn** — the steer is injected, but the history can still end on the model's last reply.
- **An interrupted tool call** — the model asked for a tool, the run was cut short, and no result came back.

`_map_messages()` mapped that trailing reply straight through as a `model` turn with no
check, so the request was rejected before it ever left the client. Every steer failed.

## Why this fix

This is a well-known Gemini 3.x behavior, not something specific to this project. Over a
hundred GitHub issues report the same message, including
[BerriAI/litellm#38537](https://github.com/BerriAI/litellm/issues/38537) (Vertex Gemini 3.7
Flash) and reports against 3.6 and 3.8 Flash in several other agent projects.

Everyone lands on the same answer: **drop the trailing model turn before sending.** Nobody
retries it, because a retry sends the same rejected shape again. Vellum's `history-repair`
does precisely this, logging that it "dropped trailing assistant turns because the provider
requires history to end with a user turn."

Code Puppy already does this for Anthropic — `_compaction.py :: history_processor()` pops
trailing model responses to dodge Anthropic's prefill error. That guard only runs during
compaction, so it never protected the Gemini path.

## The change

Six lines in `gemini_model.py`, placed right after the history is built and before the
request goes out:

```python
while contents and contents[-1].get("role") == "model":
    contents.pop()
```

Deliberately narrow:

- **Only the tail is touched.** Interior model turns are untouched, so conversation history is preserved.
- **Fixes every affected model at once.** The change is in the shared `GeminiModel` base, so subclasses inherit it without their own patch.
- **No new settings, dependencies, or retry logic.**
- **Existing empty-history fallback still applies** — if trimming empties the list, the current "one empty user turn" default takes over.

## Testing

Four tests added to `TestMapMessages`, covering the tails that occur in practice
(text-only, interrupted tool call, mixed), the all-trimmed fallback, and proof that
interior turns survive.

Verified by removing the fix and re-running: **4 tests fail without it, all pass with it** —
so they genuinely catch this bug rather than just passing.

Full suite: **7,773 passed, 32 skipped.** `ruff check` and `ruff format` clean.

One existing test, `test_merge_consecutive_model_responses`, ended on a model turn and so
would have had its subject trimmed away. It now ends with a user turn, keeping it focused on
merging behavior.

### DIFF
--- a/code_puppy/gemini_model.py
+++ b/code_puppy/gemini_model.py
@@ -389,6 +389,12 @@ class GeminiModel(Model):
                     else:
                         contents.append(model_parts)
 
+        # Gemini 3.x 400s on a history ending in a model turn, which /steer
+        # injection and interrupted tool calls both produce. Same trim
+        # _compaction.py :: history_processor() does for Anthropic prefill.
+        while contents and contents[-1].get("role") == "model":
+            contents.pop()
+
         # Ensure at least one content
         if not contents:
             contents = [{"role": "user", "parts": [{"text": ""}]}]
--- a/tests/test_gemini_model_full_coverage.py
+++ b/tests/test_gemini_model_full_coverage.py
@@ -443,15 +443,75 @@ class TestMapMessages:
 
     @pytest.mark.anyio
     async def test_merge_consecutive_model_responses(self, model, default_params):
+        # Trailing user turn keeps the merged model turn out of the tail trim.
         msgs = [
             ModelRequest(parts=[UserPromptPart(content="hi")]),
             ModelResponse(parts=[TextPart(content="a")], model_name="m"),
             ModelResponse(parts=[TextPart(content="b")], model_name="m"),
+            ModelRequest(parts=[UserPromptPart(content="steer")]),
         ]
         _, contents = await model._map_messages(msgs, default_params)
         model_msgs = [c for c in contents if c["role"] == "model"]
         assert len(model_msgs) == 1
         assert len(model_msgs[0]["parts"]) == 2
+
+    @pytest.mark.anyio
+    @pytest.mark.parametrize(
+        "trailing_parts, case",
+        [
+            ([TextPart(content="partial answer")], "text-only tail"),
+            (
+                [ToolCallPart(tool_name="t", args={}, tool_call_id="id1")],
+                "interrupted tool-call tail",
+            ),
+            (
+                [
+                    TextPart(content="a"),
+                    ToolCallPart(tool_name="t", args={}, tool_call_id="id2"),
+                ],
+                "mixed tail",
+            ),
+        ],
+    )
+    async def test_trailing_model_turns_are_dropped(
+        self, model, default_params, trailing_parts, case
+    ):
+        """Gemini 3.x 400s on a history ending in a model turn."""
+        msgs = [
+            ModelRequest(parts=[UserPromptPart(content="hi")]),
+            ModelResponse(parts=trailing_parts, model_name="m"),
+        ]
+        _, contents = await model._map_messages(msgs, default_params)
+        assert [c["role"] for c in contents] == ["user"], case
+        assert contents[0]["parts"] == [{"text": "hi"}], case
+
+    @pytest.mark.anyio
+    async def test_all_model_turns_falls_back_to_empty_user_turn(
+        self, model, default_params
+    ):
+        """Dropping every turn still yields a valid single user content."""
+        msgs = [ModelResponse(parts=[TextPart(content="orphan")], model_name="m")]
+        _, contents = await model._map_messages(msgs, default_params)
+        assert contents == [{"role": "user", "parts": [{"text": ""}]}]
+
+    @pytest.mark.anyio
+    async def test_interior_model_turns_are_preserved(self, model, default_params):
+        """Only the tail is trimmed; real history stays intact."""
+        msgs = [
+            ModelRequest(parts=[UserPromptPart(content="one")]),
+            ModelResponse(parts=[TextPart(content="two")], model_name="m"),
+            ModelRequest(parts=[UserPromptPart(content="three")]),
+            ModelResponse(parts=[TextPart(content="four")], model_name="m"),
+            ModelRequest(parts=[UserPromptPart(content="steer")]),
+        ]
+        _, contents = await model._map_messages(msgs, default_params)
+        assert [c["role"] for c in contents] == [
+            "user",
+            "model",
+            "user",
+            "model",
+            "user",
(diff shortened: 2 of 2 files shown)


thomwebb (approved):

AndrewTilson (commented):
## Review — `fix(gemini): drop trailing model turns before sending to Gemini`

Nice, tight bugfix. Six lines, right layer, real tests. I verified the core claim by reverting the hunk and re-running: **4 tests fail without the fix, 83 pass with it**, and `ruff check` / `ruff format --check` are clean on both files. The tests genuinely bite.

I'm not blocking on any of this, but a few things are worth fixing before merge — one of which is a factual error in the PR description that future readers will trip over.

---

### 1. The stated rationale is wrong — the Anthropic guard is *not* compaction-only 

The description says:

> Code Puppy already does this for Anthropic — `_compaction.py :: history_processor()` pops trailing model responses to dodge Anthropic's prefill error. **That guard only runs during compaction, so it never protected the Gemini path.**

That second sentence isn't accurate. In `_compaction.py`, the pop at lines 383-384 sits in `history_processor()` *outside* any compaction conditional — it runs on **every** model call, not just when `compact()` actually drops anything. The docstring on `make_history_processor()` even lists it as unconditional step 6: "Trims trailing ModelResponse messages so history ends with a ModelRequest."

And `_builder.py:649-650` wires that processor into every agent, with `make_steer_history_processor` appending steers as `ModelRequest` (user) turns *after* it.

So the justification as written doesn't hold, which means **the actual root cause is still unexplained**. Either:
- there's a path that bypasses `history_processor` (worth naming explicitly), or
- this is defense-in-depth at the transport boundary.

I actually think it *is* the second one, and that's a perfectly good reason on its own — the mapping layer is the correct place to enforce a wire-format invariant, since it's the last thing before the bytes go out. But please say that, rather than an incorrect claim about the existing guard. Whoever debugs the next Gemini 400 will start from this description.

**Ask:** correct the description, and if you know the actual bypassing path, name it.

### 2. Silent data loss with a logger sitting right there 

`gemini_model.py` already has `logger = logging.getLogger(__name__)` (line 41) and the loop uses it zero times. This discards real model output — a completed answer or a pending tool call — and leaves no trace. When someone reports "Gemini dropped my response," there is nothing to grep for.

Your own PR body quotes Vellum doing exactly the right thing here: they *log* that they "dropped trailing assistant turns." Let's match that.

```python
dropped = 0
while contents and contents[-1].get("role") == "model":
    contents.pop()
    dropped += 1
if dropped:
    logger.debug("Dropped %d trailing model turn(s); Gemini requires history to end on a user turn.", dropped)
```

Explicit is better than implicit. One `debug` line, no behavior change.

### 3. The all-trimmed fallback probably trades one 400 for another 

`test_all_model_turns_falls_back_to_empty_user_turn` asserts this:

```python
contents == [{"role": "user", "parts": [{"text": ""}]}]
```

The PR frames the existing fallback as a feature ("Existing empty-history fallback still applies"). The catch: **this fix is what makes that branch newly reachable.** Before, `contents` was only empty when there were no messages at all. Now an all-model-turn history lands there too.

And Gemini is well known for rejecting empty `text` parts with its own `400 INVALID_ARGUMENT`. So in that path we may be swapping "ends with a model turn" for "empty text part" — same failure, different message, now with the model's actual output thrown away first.

The test locks in the shape, not the outcome. Worth confirming against the live API that `{"text": ""}` is actually accepted; if it isn't, that branch deserves a real error rather than a doomed request.

### 4. This is now the third copy of the same trim — DRY 

Same invariant, three implementations:
- `_compaction.py:383` — `while cleaned and isinstance(cleaned[-1], ModelResponse): cleaned.pop()`
- `gemini_model.py:392` (this PR) — same thing on mapped dicts
- `gemini_code_assist.py:_build_request()` — builds `{"role": "model", ...}` contents (line 206) with **no trim at all**

Two operate on `ModelMessage` objects and one on wire dicts, so they can't share code trivially — I'm not going to demand an abstraction that doesn't fit. But the third bullet is the interesting one: `GeminiCodeAssistModel` has the identical latent bug.

Mitigating factor, and why this is a nit and not a blocker: I grepped, and `GeminiCodeAssistModel` is never instantiated anywhere outside its own tests — `model_factory.py` only constructs `GeminiModel` (lines 775, 1044). It's effectively dormant. Still worth a `TODO` referencing this fix so it doesn't ship broken the day someone wires it up.

### 5. Comment says "3.x", code applies to everything 

The comment scopes the behavior to Gemini 3.x, but the trim runs for every model `GeminiModel` serves, including 2.5, which doesn't have this restriction. Harmless in practice — a trailing model turn is useless to send regardless — but the comment and the code disagree. Either widen the comment or note deliberately that it's applied universally for simplicity.

### 6. Nice touches worth calling out 

- Fixing `test_merge_consecutive_model_responses` by adding a trailing user turn — rather than weakening the assertion — keeps it testing merge behavior. Correct call.
- Parametrizing the three tail shapes with a `case` label that surfaces in assertion output is good test hygiene.
- Placing the trim after mapping and before the empty-check is the right ordering.
- Verifying the tests fail without the fix, and saying so in the description, is exactly the diligence I want to see. I reproduced it and it checks out.

---

### Verification performed

| Check | Result |
|---|---|
| `pytest tests/test_gemini_model_full_coverage.py` | 83 passed |
| Same suite, fix reverted | 4 failed, 79 passed  tests are load-bearing |
| `ruff check` (both files) | All checks passed |
| `ruff format --check` | 2 files already formatted |
| 3 consecutive trailing model turns | correctly trimmed to `['user']` |
| Interior turns w/ orphaned tool call | preserved, `thoughtSignature` intact |

### Risk: **Low**

Narrow, well-tested, correct layer, no new deps or config. The code is fine. Items 1-3 are about the description being wrong, the drop being invisible, and a fallback that may not actually work — worth a follow-up commit, not a rewrite.

**Verdict:** Approve once the description is corrected and the `logger.debug` is added. Items 3-5 are fine as follow-ups.

 *Reviewed by Luna*

Issue #2, opened 2025-05-29
Request support for URI resource access


JuliusWeaver:
One of the current standards of documentation for live APIs is OpenAPI or Swagger and it provides that information using URLs or URIs the user can navigate to. Being able to use that standardized structure of documentation in instructions and configuration would be great and would improve the flexibility of this tool beyond most alternatives.

PR #914, opened 2026-09-06
fix(gemini): keep steering in current task

WSxDemise:
## Summary

- Treat `/steer` as additional guidance for the task that is already running.
- Keep Gemini tool results separate from steer text, avoiding invalid requests that mix `function_response` and text in one message.
- Keep steer guidance active across intermediate tool calls, then retire it when the next normal user prompt starts.
- Preserve steer attachments as part of the same current-task guidance.

## Interaction behavior

- `/steer` changes the direction of the current running task; it is not a new prompt.
- `Ctrl+C` stops the current task. A later prompt then starts a new task.
- A normal prompt entered while work is running is queued until the current final answer returns.
- `/queue` opens the queue manager.

## Why this is not Vertex-only

Public reports show the same Gemini tool-result/message-ordering behavior through both the regular Gemini Developer API / AI Studio and Vertex AI. The fix therefore belongs in the shared Gemini message mapper rather than a Vertex-specific integration.

Examples: [Pydantic AI #4210](https://github.com/pydantic/pydantic-ai/issues/4210), [OpenCode #45359](https://github.com/anomalyco/opencode/issues/45359), [OpenCode #47034](https://github.com/anomalyco/opencode/issues/47034), and [Pydantic AI #3763](https://github.com/pydantic/pydantic-ai/issues/3763).

## Validation

- 91 focused Gemini and steer-runtime tests passed.
- Ruff lint and formatting checks passed.
- A live OSS-to-Element test completed the original tool-based task and followed the steer in the same run.
- Element accepted separate tool-result and steer messages, while the mixed `function_response` + text shape returned HTTP 400.
- SAST found no blocking issues. SCA was unavailable because `uv.lock` is already out of sync with `pyproject.toml` on `main`; this PR does not change either file.


AndrewTilson (commented):
## Review: `fix(gemini): keep steering in current task`

Nice, tightly-scoped fix — the core insight (a steer is *guidance for the running task*, not a new turn, and must not be glued onto a `function_response` block) is correct, and routing it through the shared Gemini mapper rather than a Vertex-only path is the right call. I checked out the branch and verified the claims rather than eyeballing the diff.

**Verified locally:** 91 tests pass in `tests/test_gemini_model_full_coverage.py` + `tests/agents/test_runtime_pause_leakage.py`. Ruff reports 7 findings in the touched files, but all 7 are **pre-existing on `main`** (`UP035`/`UP006` typing nits) — not introduced here. `ruff format --check` is clean. `metadata` is a real `ModelRequest` field on the pinned pydantic-ai, and it survives a `ModelMessagesTypeAdapter` JSON round-trip, so the marker persists across session save/load.

The happy path genuinely works: steer → model turn → tool result maps to clean, unmixed blocks, and the steer correctly retires on the next normal prompt.

---

### 1. The mixed `function_response` + text shape can still be produced

This is the exact HTTP 400 shape the PR sets out to eliminate. The steer block is appended **unconditionally**, but the generic path below it still merges into the last block when `contents[-1]["role"] == "user"`. So any `ToolReturnPart` that lands after a steer *without* an intervening model turn that maps to at least one part merges straight back into the steer block.

Reproduced on this branch:

```
D) parallel tool calls: c1 returns, steer arrives, c2 returns
   [3] user: [['text'], ['text'], ['function_response']]   <-- mixed, 400 shape

C) model turn that maps to zero parts (empty ModelResponse)
   [3] user: [['text'], ['text'], ['function_response']]   <-- mixed, 400 shape
```

To be fair on likelihood: the ordinary sequence is clean, and a thinking-only model turn is also clean. It needs an outstanding tool call spanning the steer (parallel tool calls) or a model turn that maps to nothing. But that's precisely the window `/steer` is designed for — the user steers *while tools are in flight* — so it's reachable in the wild, and the failure mode is a hard 400 that kills the run.

Suggested fix: mark the steer block so the merge branch refuses to absorb into it, e.g. tag it and check before merging:

```python
if contents and contents[-1].get("role") == "user" and not contents[-1].pop("_steer", None):
    contents[-1]["parts"].extend(message_parts)
else:
    contents.append({"role": "user", "parts": message_parts})
```

(Your own validation notes Element accepts *separate* tool-result and steer messages, so emitting an adjacent user block is the safe branch.) Worth a regression test for the parallel-tool-call ordering.

### 2. Steers are permanently amnesiac — on Gemini only

Once the next normal prompt arrives, the steer isn't just deprioritised, it's **erased from the history sent to Gemini**:

```
history: "build a web app" → tool → steer "use PostgreSQL, NOT MySQL"
         → "done" → "now add a users table"
sent to Gemini: "PostgreSQL" present? False
```

Retiring the *steer framing* is right and clearly intended. Dropping the *user's actual words* forever is a stronger claim, and it's provider-inconsistent: every other model keeps that message (the processor mirrors it into `agent._message_history`, and only `gemini_model.py` filters on the marker). Combined with the confirmed metadata JSON round-trip, a reloaded session silently loses those instructions on Gemini for good. A user who steers "use Postgres, not MySQL" reasonably expects that to stick.

Consider demoting retired steers to a plain user message (drop the preamble, keep the content) instead of skipping them entirely.

### 3. `"code_puppy_steer"` is a magic string in 5 places

Repeated across `_steer_processor.py`, `gemini_model.py` (×2) and both test files. One typo in the producer and the consumer silently stops matching — a failure that is invisible until someone gets a 400. Hoist it to a shared constant and import it on both sides. Same DRY point for `(m.metadata or {}).get(...)`, which appears twice in `_map_messages`.

### 4. Two steers → duplicated preamble and adjacent user blocks

Three consecutive `user` blocks with the preamble repeated verbatim:

```
roles: ['user', 'user', 'user'];  preamble count: 2
```

The rest of `_map_messages` works hard to coalesce same-role blocks; this path deliberately opts out. Merging consecutive active steers into one block with a single preamble would be tidier and more consistent with the file's own invariant.

### 5. Minor

- A `SystemPromptPart` inside a steer request is silently dropped (the steer branch only handles `UserPromptPart`, and `continue` skips the `system_parts` collection). Not reachable today since the processor only emits `UserPromptPart`, but it's a quiet trapdoor for the next person — worth a comment or a defensive `extend`.
- `assert "steer" in str(continued_contents)` is a stringly-typed assertion over the whole structure; it would pass even if the steer landed in a wrong-shaped block. Asserting on the specific block's parts would actually pin the behaviour this PR cares about.
- Unrelated blank-line removal before `# Build system instruction` — trivial diff noise.

### Risk

**Medium.** Small, well-tested, correctly-placed change on the happy path, but it touches core message mapping for every Gemini request, and finding #1 means the 400 it fixes is still reachable under parallel tool calls. Worth resolving #1 (and ideally #2) before merge; the rest are cleanups.

Good fix overall — the diagnosis and the placement are both right, it just needs the merge-back hole closed. 

<sub>Reviewed by Luna — branch checked out, tests executed, findings reproduced locally.</sub>

WSxDemise:

Thanks — finding #1 was a real hole, and the parallel-tool-call repro made it trivial to confirm. All five items are addressed in d94a718e.

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Mixed `function_response` + text still reachable | **Fixed** | `gemini_model.py:366` tracks `steer_block_open`; the merge branch at `:423` refuses to absorb into an open steer block, so a tool result outstanding across the steer starts its own block |
| 2 | Steers permanently amnesiac on Gemini | **Fixed** | Retired steers now map through the same branch with the preamble omitted (`gemini_model.py:384`) instead of being skipped — the user's words stay in history, matching every other provider |
| 3 | `"code_puppy_steer"` magic string ×5 | **Fixed** | Hoisted to `code_puppy/steer_metadata.py` (`STEER_METADATA`, `is_steer_request()`); all five sites import it |
| 4 | Two steers → duplicated preamble | **Fixed** | Consecutive active steers coalesce into one block with one preamble (`gemini_model.py:377`) |
| 5a | `SystemPromptPart` in a steer silently dropped | **Fixed** | The steer branch now collects it into `system_parts` (`gemini_model.py:373`) |
| 5b | Stringly-typed `assert "steer" in str(...)` | **Fixed** | Replaced with exact block-shape/parts assertions pinning preamble position, adjacency, and part kinds |
| 5c | Unrelated blank-line removal | **Fixed** | Restored |

I took the state-flag approach rather than the suggested `contents[-1].pop("_steer")` because popping mutates the emitted payload as a side effect of a read, and the flag also gives #4 its "are we still in the same steer block?" answer for free.

**Verification of your two repros, before vs. after on this branch:**

```
              before (b17cff2b)                     after (d94a718e)
C)  user: [text, text, function_response]  →  user: [text, text] + user: [function_response]
D)  user: [text, text, function_response]  →  user: [text, text] + user: [function_response]
memory)  "PostgreSQL" retained: False      →  "PostgreSQL" retained: True (preamble still retired)
```

Both mixed 400-shapes are gone; no block in either scenario now carries `function_response` alongside `text`.

**Tests:** 94 pass in the two touched files (was 91) — added regressions for parallel-tool-call ordering, consecutive-steer coalescing, and the steer `SystemPromptPart`. Full suite: 7776 passed, 1 failed — `tests/command_line/test_autosave_menu.py::test_open_project_and_select_session`, which I confirmed fails identically on the base commit with this branch's changes stashed, so it is pre-existing and unrelated. Ruff check + format clean on all five files.

One note for anyone reproducing locally: a local `gemini_model_turn` plugin detects "is core already fixed?" by grepping `_map_messages` source for these literals. Hoisting them into constants defeats that check, so a stale copy of that plugin will re-apply its monkeypatch and fail this file's tests. That is a local-environment artifact only — CI and a clean `HOME` are green.

---
C.C. @wblake2 for review

AndrewTilson (commented):
## Follow-up review: `d94a718e`

Re-reviewed the new commit against my earlier findings. I re-ran both original repros and added fresh edge cases rather than taking the table at face value. **All five findings confirmed fixed** — and the `steer_block_open` state flag is the better call over my suggested `contents[-1].pop("_steer")`. Popping mutates the emitted payload as a side effect of a read, which is precisely the kind of spooky-action-at-a-distance the Zen warns about; the flag is explicit and does double duty for #4. Good pushback.

### Independently reproduced on `d94a718e`

```
                                   before (b17cff2b)              after (d94a718e)
C) empty model turn        user:[text,text,function_response] -> user:[text,text] + user:[function_response]  clean
D) parallel tool calls     user:[text,text,function_response] -> user:[text,text] + user:[function_response]  clean
memory) "PostgreSQL"       retained: False                    -> retained: True, preamble retired: True
#4) two steers             preamble x2, 3 blocks              -> preamble x1, coalesced
#5a) SystemPromptPart      dropped                            -> preserved in system_parts
```

New edge cases I threw at it that the tests don't cover — **all clean, no mixed blocks**: thinking-only model turn between steer and tool result; retired steer directly adjacent to the normal prompt; steer carrying *only* a `SystemPromptPart`; steer as the very first message followed by a tool result.

**Claims audited:** 94 tests pass in the two touched files (was 91). `ruff check` clean on the new `steer_metadata.py`, all three files formatted. The `test_autosave_menu.py::test_open_project_and_select_session` failure is **genuinely pre-existing** — I ran it on `1e9ba09` (main) and it fails identically there. Not this PR's problem.

The rewritten assertions are a real upgrade: exact block-shape comparisons pinning part kinds and adjacency, replacing the stringly-typed `assert "steer" in str(...)`. `test_steer_never_mixes_with_a_parallel_tool_result` is exactly the regression that was missing. `steer_metadata.py` is nicely scoped and its docstring explains *why* it's dependency-free.

---

### One residual edge (non-blocking)

A normal prompt whose parts map to **zero** parts doesn't close the steer block, so a retired steer and an active steer can merge and the active steer loses its preamble:

```
retired steer -> UserPromptPart(content=[]) -> active steer
result: [0] user: [{'text': 'retired steer'}, {'text': 'active steer'}]   preamble count: 0
```

`steer_block_open = False` lives inside `if message_parts:`, so an empty-mapping prompt skips the reset. Note the comment at `gemini_model.py:377` asserts a kind change "can only happen across the normal prompt that closes the block" — this is the one case where that invariant doesn't hold.

Reachability is low: `content=""` behaves correctly (3 blocks, preamble intact); it needs a literally empty list/parts. So this is a latent-invariant nit, not a live 400. Moving the reset out of the `if` (or resetting on any non-steer `ModelRequest`) would make the comment unconditionally true.

### Nits

- `STEER_METADATA` is a mutable module-level dict; mutating it through one import is visible to all others. The producer correctly does `dict(STEER_METADATA)`, but `test_runtime_pause_leakage.py` now asserts `metadata == STEER_METADATA` against the shared object. A `MappingProxyType` (or comparing against a literal in the test) would remove the footgun.
- Re your local `gemini_model_turn` plugin grepping `_map_messages` for the literals: agreed that's the plugin's problem, not core's. Worth having it key off `is_steer_request` / `STEER_PREAMBLE` from `code_puppy.steer_metadata` so it degrades predictably.

### Verdict

**Risk: low** — down from medium. The 400-shape hole is closed and regression-tested, Gemini now has provider parity on steer memory, and the magic string is gone. The residual edge is a latent invariant gap, not a live failure. Ship it; the two nits are follow-up cleanups.

Nice turnaround.

<sub>Reviewed by Luna — commit d94a718e checked out, repros re-run, pre-existing failure verified against main.</sub>

WSxDemise:

Followed up on Luna's review feedback from commit `d94a718e`:

1. **Steer block closed on non-steer request regardless of mapped parts count**:
   - `steer_block_open = False` has been moved out of `if message_parts:` so that a non-steer `ModelRequest` with zero mapped parts (e.g. `UserPromptPart(content=[])`) unconditionally closes the steer block. This ensures that any subsequent steer starts a new block with its preamble intact. Added regression test `test_empty_mapping_prompt_closes_steer_block`.

2. **Immutable `STEER_METADATA`**:
   - Wrapped `STEER_METADATA` in `MappingProxyType` in `code_puppy/steer_metadata.py` to prevent accidental mutation of the shared module constant.

All 95 unit tests pass and formatting/lint checks are clean. Pushed in commit `636203cf`.

AndrewTilson (commented):
## Follow-up review: `636203cf`

Both residual nits from my last pass are fixed and independently verified. No new findings of substance — this is ready.

### Verified on `636203cf`

**Empty-mapping prompt now closes the steer block.** `steer_block_open = False` moved out of `if message_parts:`, so any non-steer `ModelRequest` closes the block regardless of how many parts it maps to. My exact repro flips:

```
before (d94a718e):  [0] user: [retired steer, active steer]     preamble count: 0
after  (636203cf):  [0] user: [retired steer]
                    [1] user: [PREAMBLE, active steer]          preamble count: 1
```

The invariant in the `:377` comment now actually holds. `test_empty_mapping_prompt_closes_steer_block` pins it.

**`STEER_METADATA` frozen.** `MappingProxyType` confirmed — `'mappingproxy' object does not support item assignment`. Producer still round-trips correctly via `dict(STEER_METADATA)`.

**No regressions.** Re-ran the full battery from both prior rounds — parallel tool calls, empty model turn, retired-steer memory, consecutive steers, system-part-in-steer, thinking-only turn, steer-first, steer-adjacent-to-prompt. **Every scenario clean, zero mixed `function_response`+text blocks.** 95 tests pass (was 94). `steer_metadata.py` is ruff-clean; the 7 findings in the other two files are byte-identical to `main` (1 in `gemini_model.py`, 6 in `_steer_processor.py`) — pre-existing, not this PR's.

### Two cosmetic notes for a future cleanup (not blocking)

**1. Tests pass the proxy where production passes a dict.** Production correctly does `metadata=dict(STEER_METADATA)`; the tests do `metadata=STEER_METADATA` in 7 places. pydantic doesn't love it:

```
dict(STEER_METADATA)  [production] : dump_python OK, dump_json OK
STEER_METADATA proxy  [tests]      : dump_python OK (+ serializer warning), dump_json FAILS
                                     PydanticSerializationUnexpectedValue: expected dict[str, any],
                                     got mappingproxy
```

Harmless today since `_map_messages` only reads the mapping and nothing serializes these fixtures — but the tests now exercise a shape production never produces, which is the wrong direction for a fixture. `dict(STEER_METADATA)` in the fixtures would match reality and silence the warning. (`assert msgs[0].metadata == STEER_METADATA` is fine — proxies compare equal to dicts.)

**2. `ModelResponse` keeps the old asymmetry.** Its `steer_block_open = False` still sits inside `if model_parts:`, so a zero-part `ModelResponse` doesn't close the block. I probed it — **not exploitable**: the steer branch's own `steer_block_open` guard means a following steer coalesces (correct) and a following tool result opens its own block (correct, verified clean). Purely a consistency wart now that the `ModelRequest` side is unconditional.

### Verdict

**Risk: low. Approve.** Three rounds in, every finding is closed with a regression test behind it: the 400-shape hole, Gemini steer-memory parity, the magic string, preamble duplication, the dropped system part, and now the block-lifecycle edge. The two notes above are tidy-ups, not merge blockers.

Nice work — the state-flag design held up under everything I threw at it.

<sub>Reviewed by Luna — commit 636203cf checked out, all prior repros re-run, lint delta diffed against main.</sub>

WSxDemise:

Addressed both cosmetic cleanups noted in Luna's approval review on `636203cf`:

1. **Test fixture metadata types**:
   - Passed `dict(STEER_METADATA)` instead of the raw `MappingProxyType` proxy in test message fixtures so test fixtures mirror exact production data structures and avoid Pydantic serialization warnings on MappingProxy.
2. **ModelResponse steer block reset symmetry**:
   - Moved `steer_block_open = False` outside `if model_parts:` in the `ModelResponse` handling branch of `_map_messages`, making steer block closure unconditional across all response types and maintaining symmetric lifecycle handling with `ModelRequest`.

All 95 unit tests pass, ruff check/format clean. Pushed in commit `2b31e3fc`.

AndrewTilson (changes requested):
## Follow-up review: `2b31e3fc` — please hold merge

The test-fixture cleanup is correct. **But the `ModelResponse` symmetry change reintroduces the exact HTTP 400 shape this PR exists to eliminate.** Requesting a revert of that one line.

I flagged the `ModelResponse` asymmetry last round as a "consistency wart" and explicitly noted I'd probed it as *not exploitable*. I was wrong to imply it was safe to "fix" — making it symmetric is what breaks it. My apologies for the misleading nudge.

### The regression

`steer_block_open = False` moved outside `if model_parts:` in the `ModelResponse` branch. A model turn that maps to **zero** parts now closes the steer block without having appended anything, so `contents[-1]` is *still the steer block* — and the next tool result merges straight into it.

Bisected on the two commits, same input both times:

```
steer -> ModelResponse(parts=[]) -> tool result

636203cf:  [1] user: [text, text]
           [2] user: [function_response]                 clean

2b31e3fc:  [1] user: [text, text, function_response]     <-- mixed, HTTP 400
```

Two shapes reach it, both mapping to zero parts:

```
ModelResponse(parts=[])                     -> 400-SHAPE
ModelResponse(parts=[ThinkingPart('')])     -> 400-SHAPE   <-- empty thought
ModelResponse(parts=[ThinkingPart('hmm')])  -> clean
```

The second is the one that worries me: an empty-content `ThinkingPart` is a plausible artifact from a streaming thinking model, which is exactly the Gemini 3.x configuration this whole PR targets. Combined with a steer, that's a hard 400 that kills the run.

**The 95 tests pass identically with and without the change** — I ran them both ways. The suite cannot currently tell these two behaviours apart, which is why this slipped through green.

### Why the two branches are legitimately asymmetric

The flag is doing two different jobs, and they want opposite handling:

1. **Block safety** — "don't merge foreign parts into the steer block." True for as long as `contents[-1]` *is* the steer block. A zero-part `ModelResponse` doesn't change `contents[-1]`, so it must **not** reset.
2. **Semantic window** — "the current-task steer is over." A normal user prompt closes it even when it renders nothing, which is why the unconditional reset in the `ModelRequest` branch (`636203cf`) was right.

Same variable, two meanings. The asymmetry wasn't a wart; it was concern #1 and concern #2 landing in the same boolean.

### Suggested fix

Minimal — revert that one line:

```python
                    else:
                        contents.append(model_parts)
                    steer_block_open = False   # back inside `if model_parts:`
```

Verified: restores clean output on every zero-part shape above, keeps all 95 tests green, and R1 (two steers around an empty model turn) actually *improves* — one coalesced block with one preamble instead of two split blocks.

More principled, if you'd rather kill the ambiguity: track the block itself rather than a boolean, so safety can't desync from bookkeeping.

```python
steer_block = contents[-1] if <just appended a steer> else None
# merge guard: `contents[-1] is not steer_block`
```

Then "is `contents[-1]` still the steer block?" is answered by identity, and the `ModelRequest` branch keeps its separate semantic reset.

Either way, please add a regression test for `steer -> zero-part ModelResponse -> tool result`. That's the gap that let a green suite hide this.

### Everything else on `2b31e3fc` still good

Test fixtures now use `dict(STEER_METADATA)` in all 7 places — matches production, serializer warning gone. Re-ran the full battery: parallel tools, retired-steer memory (`PostgreSQL` retained), consecutive-steer coalescing, empty-list prompt closing the block, thinking-only turn, system-part-in-steer, steer-first. **All clean** — R2 is the only regression. Lint unchanged from main (1 pre-existing in `gemini_model.py`), formatting clean.

### Verdict

**Request changes.** One-line revert plus a regression test and this is back to ship-shape — everything from rounds 1-3 is still solid. Sorry again for pointing you at this one; good reminder that "make it symmetric" and "make it correct" aren't always the same commit.

<sub>Reviewed by Luna — 2b31e3fc checked out, regression bisected against 636203cf, fix verified locally.</sub>

AndrewTilson (commented):
**[puppy-review]** Reviewed at `2b31e3fc`.

The core design here is sound and genuinely well-reasoned. Computing the preamble at map time rather than storing it in history is the right call (retirement stays a pure function of position, so re-mapping the same history is deterministic and no preamble text ever pollutes a persisted session), and retiring the *framing* while keeping the user's words is the correct resolution of the earlier review round.

That said, I found **two defects that survive at current HEAD**, both confirmed by executing against this branch rather than by reading the diff. Both produce malformed Gemini payloads, and one of them was *introduced* by a fix applied in response to earlier review feedback.

| # | Finding | Severity | Location |
|---|---|---|---|
| 1 | Steer branch silently discards `ToolReturnPart` / `RetryPromptPart`, producing a dangling `function_call` | **High** | `gemini_model.py:369-387` |
| 2 | A zero-part request re-opens the merge path, resurrecting the mixed `text` + `function_response` 400 | **Medium-High** | `gemini_model.py:431` |
| 3 | `is_steer_request()` raises `AttributeError` on non-dict metadata | Low-Medium | `steer_metadata.py:28` |

Details inline. Finding #1 is the one I'd prioritise: it is reachable on every run via `prune_interrupted_tool_calls()`, not just in a contrived history.

### What I checked and found correct

- The central mechanism works. With a steer-tagged request, `function_response` genuinely does stay out of the steer text block, and preamble retirement behaves as advertised.
- Round-tripping `metadata` through `ModelMessagesTypeAdapter` (pydantic-ai 2.35.0) preserves the tag, so persisted sessions restore steer framing correctly.
- Placing `steer_metadata.py` at the package root rather than under `code_puppy/agents/` is justified, and I verified the stated constraint is real: importing `code_puppy.agents` transitively pulls `agent_manager` -> `base_agent` -> `_builder` -> `_compaction` -> `pydantic_ai_harness`. A provider module importing the tag from `agents/` would drag that whole graph in.
- Repository fit is correct. This is core runtime work (a built-in provider mapper plus the agent history-processor pipeline), not plugin code, so `code_puppy_core_plugins` does not apply. No Walmart signal anywhere.

### Scope observation (not a defect)

The steer tag is consumed only by the Gemini mapper; OpenAI/Anthropic paths ignore it, so a steer reads as a plain user turn there. Consistent with the `fix(gemini)` title, but it does mean cross-provider steer semantics now differ. Worth a line in the `steer_metadata.py` docstring so the next provider author knows the hook exists.

### On the plugin note in the PR description

You mention a local `gemini_model_turn` plugin that detects "is core already fixed?" by grepping `_map_messages` source for literals. If that plugin lives in `code_puppy_core_plugins`, this PR deliberately defeats its heuristic and a coordinated follow-up PR over there is needed.

**Recommendation: changes requested** (posting as a comment rather than a blocking review). The two defects are narrow gaps in the new steer branch's part handling, not architectural problems. Happy to re-review once they're pushed.

code_puppy/gemini_model.py
```diff
@@ -338,8 +346,45 @@ async def _map_messages(
...
+                        # Consecutive steers share one block and one preamble.
+                        # A kind change (retired -> active) can only happen
+                        # across the normal prompt that closes the block.
+                        contents[-1]["parts"].extend(steer_parts)
+                    else:
+                        if i > last_prompt_index:
+                            steer_parts.insert(0, {"text": STEER_PREAMBLE})
+                        contents.append({"role": "user", "parts": steer_parts})
+                        steer_block_open = True
+                continue
```
AndrewTilson, line 455: ## :red_circle: High - steer branch drops `ToolReturnPart` and `RetryPromptPart`

This branch handles only `SystemPromptPart` and `UserPromptPart`. Every other part type falls through the loop silently, and then `continue` (line 387) skips the general `ModelRequest` handler below, so those parts are never mapped at all.

This is **reachable in normal operation**, not theoretical. `_history.prune_interrupted_tool_calls()` repairs a dangling tool call by splicing a synthesized `ToolReturnPart` into whichever request follows it, via `_insert_tool_returns()` -> `dataclasses.replace(request, parts=parts)`. Critically, **`dataclasses.replace` preserves `metadata`** - so a steer injected while a tool call is outstanding becomes exactly that request, and stays steer-tagged.

Reproduced against this branch:

```python
msgs = [
    ModelRequest(parts=[UserPromptPart(content='start')]),
    ModelResponse(parts=[ToolCallPart(tool_name='shell', args={}, tool_call_id='call-1')], model_name='m'),
    ModelRequest(parts=[UserPromptPart(content='steer me')], metadata=dict(STEER_METADATA)),
]
pruned = prune_interrupted_tool_calls(msgs)
# -> ModelRequest(parts=[ToolReturnPart, UserPromptPart], metadata={'code_puppy_steer': True})
```

Mapping that repaired history through `_map_messages`, toggling only the steer tag:

```
steer-tagged=False  ->  function_response present = True    # correct, main's behavior
steer-tagged=True   ->  function_response present = False   # dropped
```

The emitted `contents` then carries a `function_call` in the model turn with **no matching `function_response`** - a dangling call, which is itself a Gemini 400 shape and defeats the entire point of the repair pass. `RetryPromptPart` is lost through the same gap.

This is a regression introduced by this PR: the byte-identical message *without* the steer tag maps correctly today.

Worth noting the blast radius - `prune_interrupted_tool_calls()` runs on every run at `_runtime.py:903` and again in the `finally` at `_runtime.py:995`. So a `Ctrl+C` or crash during a steered tool call reliably produces the bad shape, and it is then persisted into `agent._message_history`.

**Suggested direction:** handle `ToolReturnPart` / `RetryPromptPart` in this branch by emitting them as their **own preceding `contents` block** - never merged into the steer text block. That preserves the no-mixing invariant while keeping the tool result on the wire.

code_puppy/gemini_model.py
```diff
@@ -375,10 +420,15 @@ async def _map_messages(
...
-                    if contents and contents[-1].get("role") == "user":
+                    if (
+                        contents
+                        and contents[-1].get("role") == "user"
+                        and not steer_block_open
+                    ):
                         contents[-1]["parts"].extend(message_parts)
                     else:
                         contents.append({"role": "user", "parts": message_parts})
+                steer_block_open = False
```
AndrewTilson, line 499: ## :orange_circle: Medium-High - a zero-part request re-opens the merge path

Moving `steer_block_open = False` outside `if message_parts:` (commit `636203cf`, in response to review feedback) fixed one hole but opened another.

A request that maps to **zero** parts emits nothing, yet still clears the flag. The steer block is therefore still `contents[-1]`, still `role == "user"`, and now merge-eligible again at lines 423-427. The next tool return merges straight into the steer's text block.

The earlier round used `UserPromptPart(content=[])` as the example, which reads as synthetic. Here is a realistic trigger - a bare `SystemPromptPart` routes to `system_parts` and contributes zero `message_parts`:

```python
msgs = [
    ModelResponse(parts=[ToolCallPart(tool_name='shell', args={}, tool_call_id='c1')], model_name='m'),
    ModelRequest(parts=[UserPromptPart(content='steer')], metadata=dict(STEER_METADATA)),
    ModelRequest(parts=[SystemPromptPart(content='sys')]),      # zero mapped parts
    ModelRequest(parts=[ToolReturnPart(tool_name='shell', content='done', tool_call_id='c1')]),
]
_, contents = await model._map_messages(msgs, params)
# MIXED BLOCK: True   -> {'text', 'function_response'} together in one user block
```

That is precisely the HTTP 400 shape this feature exists to prevent.

**Suggested direction:** the deeper issue is that `steer_block_open` is loop state that can desync from the emitted `contents` list - which is exactly what happens here. Rather than only clearing the flag when a block was actually emitted (the minimal fix), consider deriving "is `contents[-1]` a steer block?" as a property of the emitted block itself. Then the two can't drift apart by construction.

tests/test_gemini_model_full_coverage.py
```diff
@@ -513,6 +515,226 @@ async def test_interior_model_turns_are_preserved(self, model, default_params):
...
+        system_instruction, contents = await model._map_messages(msgs, default_params)
+
+        assert system_instruction["parts"] == [{"text": "be terse"}]
+        assert contents[-1]["parts"] == [
+            {"text": STEER_PREAMBLE},
+            {"text": "steer"},
+        ]
+
+    @pytest.mark.anyio
+    async def test_empty_mapping_prompt_closes_steer_block(self, model, default_params):
```
AndrewTilson, line 783: :link: Related to the finding on `gemini_model.py:431`.

This test pins the unconditional flag-clearing behaviour that *enables* the mixed-block regression, so the suite currently locks the hole in rather than catching it. It passes while the payload it implies is a Gemini 400.

When you fix #2, this test will need to change shape. Suggest replacing it with an assertion on the invariant that actually matters, rather than on the flag mechanics:

```python
assert not any(
    {k for part in content["parts"] for k in part} >= {"text", "function_response"}
    for content in contents
)
```

Asserting the no-mixing invariant directly means the test keeps its value regardless of how the block-tracking is implemented internally - and it would have caught this regression.

code_puppy/steer_metadata.py
```diff
@@ -0,0 +1,29 @@
...
+
+STEER_METADATA: MappingProxyType[str, bool] = MappingProxyType(
+    {STEER_METADATA_KEY: True}
+)
+
+
+def is_steer_request(message: ModelMessage) -> bool:
+    """True if ``message`` is a ``/steer`` request injected by the processor."""
+    return isinstance(message, ModelRequest) and bool(
+        (message.metadata or {}).get(STEER_METADATA_KEY)
```
AndrewTilson: :yellow_circle: **Low-Medium** - raises on non-dict metadata.

`metadata` is typed `Any` in pydantic-ai 2.35.0. The `(message.metadata or {})` guard covers `None`, but not a truthy non-mapping:

```
is_steer_request(ModelRequest(parts=[...], metadata='oops'))
-> AttributeError: 'str' object has no attribute 'get'
```

Since this runs inside the wire mapper on every single request, a foreign `metadata` value - set by another capability, a plugin, a hand-built history, or a future pydantic-ai use of the field - would hard-fail the turn rather than degrade gracefully.

Low probability today: I verified the round-trip through `ModelMessagesTypeAdapter` preserves the dict correctly, so nothing in-tree currently triggers it. But the blast radius is a crashed run and the guard is one `isinstance`:

```python
metadata = message.metadata
return (
    isinstance(message, ModelRequest)
    and isinstance(metadata, Mapping)
    and bool(metadata.get(STEER_METADATA_KEY))
)
```

Using `Mapping` rather than `dict` also keeps the `MappingProxyType` constant itself accepted, should it ever be passed through this predicate.

code_puppy/gemini_model.py
WSxDemise: Fixed in commit `60cc1cbe`.

The steer branch now handles `ToolReturnPart` and `RetryPromptPart` (with tool names), mapping them into `tool_parts` and emitting them as their own preceding block before the steer guidance. This prevents dangling tool calls when `prune_interrupted_tool_calls` splices tool returns into steer-tagged requests, and guarantees tool returns are never mixed into steer text blocks. Added regression test `test_steer_with_tool_return_part_does_not_drop_or_mix`.

code_puppy/gemini_model.py
WSxDemise: Fixed in commit `60cc1cbe`.

Separated block-safety from coalescing bookkeeping by tracking steer blocks directly in a set (`_is_steer_block`). The merge guard now checks `and not _is_steer_block(contents[-1])`. Even if zero-part requests (like `SystemPromptPart`) or empty `ModelResponse` turns intervene, `contents[-1]` remains identified as a steer block and will never absorb subsequent tool returns or other parts. Added regression test `test_zero_part_request_does_not_reopen_merge_into_steer_block`.

tests/test_gemini_model_full_coverage.py
WSxDemise: Addressed in commit `60cc1cbe`.

Updated the test assertions to check the direct invariant:
```python
assert not any(
    {k for part in content["parts"] for k in part} >= {"text", "function_response"}
    for content in contents
)
```
Applied across all steer tests so invariant safety is directly validated rather than internal flag mechanics.

code_puppy/steer_metadata.py
WSxDemise: Fixed in commit `60cc1cbe`.

Updated `is_steer_request()` to guard against non-mapping metadata using `isinstance(metadata, Mapping)`:
```python
metadata = getattr(message, "metadata", None)
return (
    isinstance(message, ModelRequest)
    and isinstance(metadata, Mapping)
    and bool(metadata.get(STEER_METADATA_KEY))
)
```
Gracefully returns `False` on foreign metadata values (such as strings) and handles mappings/proxies correctly. Covered by unit test `test_is_steer_request_handles_non_mapping_metadata`.

WSxDemise:

All feedback from Andrew Tilson's (Luna's) review passes has been resolved and verified in commit `60cc1cbe`:

| Finding | Resolution |
|---|---|
| **1. Dropped `ToolReturnPart` / `RetryPromptPart` in steer requests** | The steer branch now maps `ToolReturnPart` and `RetryPromptPart` into `tool_parts` and emits them in their own preceding block before the steer text. Fixes dangling tool calls when `prune_interrupted_tool_calls()` splices tool returns into steer-tagged requests. |
| **2. Zero-part request reopening merge path (mixed 400 shape)** | Decoupled block safety from coalescing bookkeeping by directly tracking steer block identities in `steer_blocks: set[int]`. Non-steer parts will never merge into a steer block even if zero-part requests or empty `ModelResponse` turns intervene. |
| **3. Non-dict metadata AttributeError** | Guarded with `isinstance(metadata, Mapping)` in `is_steer_request()` in `code_puppy/steer_metadata.py`, safely handling non-mapping metadata, `None`, and `MappingProxyType`. |
| **4. Test coverage & invariant pinning** | Added dedicated regression tests for spliced tool returns, zero-part / empty model turn merge protection, and non-mapping metadata. Updated steer assertions to directly assert the no-mixing invariant (`not any({"text", "function_response"} <= parts)`). |

All 98 tests pass cleanly, ruff check/format clean. Ready for final approval and merge.

AndrewTilson (approved):
## Follow-up review: `60cc1cbe` — blocking regression resolved

My `CHANGES_REQUESTED` from `2b31e3fc` is cleared. The identity-tracking approach is the right fix and it's verified working. **One merge-integration item to sort before this lands** — `main` moved underneath you and now solves part of the same problem.

### The regression I blocked on is fixed

`steer_blocks: set[int]` + `_is_steer_block()` replaces the overloaded boolean in the merge guard, so safety is now answered by *"is `contents[-1]` literally the steer block?"* rather than by stale bookkeeping. Every zero-part shape that produced a 400 on `2b31e3fc` is clean:

```
steer -> ModelResponse(parts=[])                -> tool result :  clean
steer -> ModelResponse([ThinkingPart('')])      -> tool result :  clean
steer -> ModelResponse([ThinkingPart('hmm')])   -> tool result :  clean
```

Full battery re-run — parallel tools, retired-steer memory (`PostgreSQL` retained), consecutive-steer coalescing, empty-list prompt, thinking-only turn, system-part-in-steer, steer-first: **all clean, zero mixed blocks.**

**On keeping both `steer_block_open` and `steer_blocks`:** I tried deleting the boolean from the coalescing guard to see if it was now vestigial. It isn't — `test_empty_mapping_prompt_closes_steer_block` fails and B4 loses its preamble. The two variables genuinely encode the two different concerns (block *identity* vs. semantic *window*). Correct as written; disregard any temptation to collapse them.

The other reviewer's findings also check out: spliced `ToolReturnPart`/`RetryPromptPart` in a steer request now emit their own preceding block, ordering is `function_response` before steer text, and I confirmed **no dangling `function_call`** — every call id has a matching response. `is_steer_request()` is `Mapping`-guarded.

---

### Blocking-ish: `main` shipped an overlapping fix; PR is CONFLICTING

`main` landed `87f588cd` — *"fix(gemini): never mix function responses and text in one user turn"* — adding `_split_mixed_user_contents()`, a general post-pass over `contents` that splits any user turn mixing `function_response` with text. That is a broader solution to the same class of bug this PR fixes structurally.

I resolved the conflict locally to see how they interact. **Good news: they compose cleanly.**

- Source auto-merges; the conflict is **only** in `tests/test_gemini_model_full_coverage.py`, where both sides appended tests at the same spot. Resolves as a straight union.
- Merged tree: **100 tests** in the two touched files, **558** across `tests/agents/` + the Gemini file. All green.
- My full battery on the merged tree: all clean.

I instrumented `_split_mixed_user_contents` to see whether it ever actually fires on steer paths:

```
steer + parallel tools                 split_fired = no
steer w/ spliced ToolReturn            split_fired = no
NON-steer tool+prompt (main's case)    split_fired = YES
```

So the PR's structural fix keeps steer payloads clean *before* the post-pass ever sees them; main's net catches the non-steer case this PR never touched. Belt and braces, no double-splitting, no conflict in behaviour.

**Action needed:** merge `main` into the branch and resolve that one test conflict as a union (keep both `test_active_steer_is_transient_current_task_guidance` and main's `test_tool_return_and_user_text_never_share_a_content`). Worth a sentence in the PR description noting the two mechanisms are complementary, so a future reader doesn't "simplify" one away.

Lint on the merged tree: 2 findings, both pre-existing on `main` (`UP017` in `gemini_model.py`, `RUF059` in the test file). Formatting clean.

---

### Verdict

**Regression cleared.** Approving on behaviour — the steer mechanism is correct across every scenario I can construct, and it composes with main's new post-pass rather than fighting it.

Only ask: land the `main` merge with the union test resolution before pushing the button, since GitHub currently reports `CONFLICTING`. No further code changes needed from my side.

Thanks for riding out five rounds on this one — the identity-tracking design is meaningfully sturdier than where we started.

<sub>Reviewed by Luna — 60cc1cbe checked out, conflict resolved in a scratch worktree, 558 tests + behavioural battery run against the merged result.</sub>

PR #916, opened 2026-09-07
fix(completion): apostrophes in prose no longer hide later @file refs

### LINKED ISSUE #915: Regression: apostrophes in prompt prose disable subsequent @file completion
## Summary

After 029ca240d8dbd3b89a154f1eedcaf486ee393b72, an apostrophe in ordinary prompt prose can prevent a later `@file` token from offering completions.

For example, with `target.py` available in the current directory/index:

- `read @target` offers `target.py`.
- `don't change @target` offers no candidates.
- `it's in @target` also offers no candidates.

The cursor is at the end of each prompt.

## Expected behavior

A contraction in preceding prose should not prevent completion of a later attachment token. Quoted and escaped filenames should continue to work, and email addresses or prose following a completed attachment should not be treated as active references.

## Reproduction and verification

Reproduced using the real `FilePathCompleter.get_completions` with an injected file-index snapshot containing `target.py`, with background indexing disabled to make the test deterministic:

1. Set the current directory and the snapshot root to the same temporary directory.
2. Set the indexed paths to `["target.py"]`.
3. Request completions for `Document("don't change @target")` with the cursor at the end.
4. Observe an empty list instead of a completion for `target.py`.

The same completer probe returns `target.py` at pre-change commit `370c8b4fb195ff1215f63d706b190455d2147ff2`; it returns no candidates at `1d25d696` containing the change.

The parser itself also demonstrates the failure:

- `active_reference("read @target")` returns `("target", 6)`.
- `active_reference("don't change @target")` returns `None`.

This was verified through parser/completer calls on Linux, not a full interactive TUI test.

## Likely cause

`code_puppy/file_completion_tokens.py:25-26` treats every apostrophe or double quote in the entire preceding prompt as a quote delimiter. The apostrophe in `don't` opens quote state; subsequent spaces no longer advance the active token boundary, so the later `@target` is not recognized.

Please scope filename quote/escape handling to the active attachment token, or otherwise distinguish contractions in surrounding prose from quoted paths. Add regression coverage for contractions before `@file`, alongside the existing quoted/escaped-path and token-boundary tests.

12britz:
Fixes #915

## Summary

Since 029ca24, an apostrophe anywhere in the prompt could disable later `@file` completion: `don't change @target` offered no candidates while `read @target` worked. `active_reference()` treated *every* `'`/`"` in the prompt as a quote delimiter, so the contraction opened quote state, subsequent whitespace stopped advancing the active-token boundary, and the later `@token` was never recognized.

## Fix

A quote character now opens a quoted span only where a filename is legitimately quoted:

- at a word boundary (start of input or right after whitespace), or
- immediately following the attachment symbol (`@"my file`)

A quote embedded inside a word — the apostrophe in `don't`, `it's` — is ordinary prose and never opens quote state.

Quoted/escaped path semantics from 029ca24 are preserved: `@"my file.txt`, `@'my file`, `@my\ file.txt` still complete, and `@"my file.txt" done` still returns no active reference so trailing prose is never replaced.

## Also fixes a failing test on main

`tests/test_completions_and_small_modules.py::TestFilePathCompleterMissedLines::test_tilde_prefix_display` currently fails on `main`: it asserts insertion text starts with `~`, but 029ca24 quotes inserted paths via `shlex.quote` (its own regression test asserts `shlex.split(inserted)` parses back). Updated it to decode via `shlex.split` before asserting the tilde-prefixed path.

## Test plan

- [x] New regression cases in `tests/command_line/test_file_completion_regressions.py`: contractions before `@file` (`don't`, `it's`, `can't`, `they're`, `isn't`), quote grouping, closed-quote suppression
- [x] `uv run pytest tests/command_line/ tests/messaging/test_completion_stable_popup.py tests/messaging/test_editor_completion.py tests/plugins/test_completion_provider.py tests/test_completions_and_small_modules.py` — 1062 passed
- [x] `ruff check` + `ruff format` clean

### DIFF
--- a/code_puppy/file_completion_tokens.py
+++ b/code_puppy/file_completion_tokens.py
@@ -3,11 +3,28 @@
 import shlex
 
 
+def _opens_quote(text: str, i: int, symbol: str) -> bool:
+    """Decide whether the quote character at ``text[i]`` starts a quoted span.
+
+    Quotes group characters only when they begin a word (start of input or
+    right after whitespace) or immediately follow the attachment symbol —
+    the two positions a user quotes a filename from. A quote embedded in a
+    word (the apostrophe in ``don't``) is ordinary prose and must never open
+    quote state, or it would swallow later whitespace and hide the active
+    ``@token`` boundary.
+    """
+    if i == 0 or text[i - 1].isspace():
+        return True
+    return i >= len(symbol) and text[i - len(symbol) : i] == symbol
+
+
 def active_reference(text: str, symbol: str = "@"):
     """Return decoded path and raw replacement length, or no active reference.
 
-    Only a token beginning with @ qualifies. Spaces require quotes or escaping;
-    a closed quote ends completion so subsequent prose is never replaced.
+    Only a token beginning with @ qualifies. Within the active token, spaces
+    require quotes or escaping; a closed quote ends completion so subsequent
+    prose is never replaced. Apostrophes in surrounding prose (contractions)
+    are literal characters and never open a quote.
     """
     start = 0
     quote = None
@@ -23,7 +40,8 @@ def active_reference(text: str, symbol: str = "@"):
                 quote = None
                 closed = True
         elif char in "\"'":
-            quote = char
+            if _opens_quote(text, i, symbol):
+                quote = char
         elif char.isspace():
             start = i + 1
             closed = False
--- a/tests/command_line/test_file_completion_regressions.py
+++ b/tests/command_line/test_file_completion_regressions.py
@@ -21,6 +21,45 @@ def test_not_an_active_reference(text):
     assert active_reference(text) is None
 
 
+@pytest.mark.parametrize(
+    "text",
+    [
+        "don't change @target",
+        "it's in @target",
+        "can't @",
+        "they're @a.py",
+        "isn't @~/Documents",
+    ],
+)
+def test_contraction_prose_does_not_hide_active_reference(text):
+    # Issue #915: an apostrophe in preceding prose must not open quote state;
+    # the later @token still completes.
+    token = text[text.rfind("@") :]
+    decoded, raw_len = active_reference(text)
+    assert decoded == token[1:]
+    assert raw_len == len(token) - 1
+
+
+@pytest.mark.parametrize(
+    "text,expected",
+    [
+        ('he said "hi" @target', "target"),
+        ('read @"my file.txt" done', None),
+        ("@'unfinished", "unfinished"),
+        ('@"my file.txt', "my file.txt"),
+    ],
+)
+def test_quotes_still_group_attachment_tokens(text, expected):
+    # Word-boundary quotes keep grouping spaces (filenames with spaces), and
+    # a closed quote still ends completion so later prose is not replaced.
+    result = active_reference(text)
+    if expected is None:
+        assert result is None
+    else:
+        assert result is not None
+        assert result[0] == expected
+
+
 @pytest.mark.parametrize("raw", ['@"space fi', "@'space fi", r"@space\ fi"])
 def test_quoted_completion_replaces_entire_raw_path(raw, monkeypatch, tmp_path):
     monkeypatch.chdir(tmp_path)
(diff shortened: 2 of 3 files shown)


breedx:
Thanks for picking up #915! I confirmed that 38d3ba76 fixes the contractions from our report. The two changed test files pass locally (57 tests), and Ruff lint/format checks pass for all three changed files.

I found one compatibility regression worth addressing before merge: the new quote-opening restriction also applies inside an active attachment path. With `dir/my file.txt` present and the cursor at the end:

- `@dir/"my fi` offered a completion before this patch but now returns no candidates.
- `@dir/'my fi` likewise worked before and now returns no candidates.
- `@"dir/my fi` and `@dir/my\ fi` still work.

I reproduced this through the real `FilePathCompleter.get_completions` with temporary files and background indexing disabled, swapping only `active_reference` between base `1d25d696` and head `38d3ba76`. The base parser returns `("dir/my fi", 10)` for both partially quoted inputs; the PR parser returns `None`. The `_opens_quote` guard ignores the quote after `/`, so the following space advances the token boundary past the attachment.

Could we preserve shell-style quote handling within an active attachment token while ignoring apostrophes in preceding prose, and add both partially quoted path cases to the regression tests? The contraction fix itself is working; this is a narrow compatibility edge to retain alongside it.

Validation was on Linux through parser/completer calls and the focused tests, not a full interactive TUI run.

thomwebb:
Rebased this onto current `main` for you (had your blessing via "Allow edits from maintainers") since it had a real conflict against `main` — not from your change, but because `test_tilde_prefix_display` in `tests/test_completions_and_small_modules.py` was independently fixed on `main` by `612de2b1` (a maintainer commit that landed the same day this PR was opened, for the same shlex-quoting reason your version's `test_tilde_prefix_display` change addressed). Since `main`'s version is already the one everyone else is building against, I kept it and dropped the now-redundant duplicate hunk from this branch — your actual `file_completion_tokens.py` fix and the two new regression tests in `test_file_completion_regressions.py` are untouched.

Verified after rebase:
- `python -m pytest tests/command_line/ tests/messaging/test_completion_stable_popup.py tests/messaging/test_editor_completion.py tests/plugins/test_completion_provider.py tests/test_completions_and_small_modules.py` → **1134 passed**
- Full `tests/` run → 6 failures, all in files this PR never touches (`test_tool_failure_not_fatal.py`, `test_herdr_plugin.py`) and reproducible identically on unpatched `main`, so pre-existing/environmental, not caused by this rebase
- `ruff format --check` clean on both changed files; `ruff check` shows only pre-existing debt on lines this PR didn't touch

Full review incoming once CI settles on the rebased commit.

thomwebb (approved):
Reviewed the actual fix on top of the rebase from my earlier comment. This is a solid, correctly-scoped bug fix.

## What it does
`active_reference()` used to treat *any* `'`/`"` character as opening a quoted span — so the apostrophe in `don't change @target` opened quote state, swallowed the space after it, and hid the `@target` reference from completion entirely (#915). The fix adds `_opens_quote()`, which only lets a quote character open a span at the two positions a user actually quotes a filename from: the start of a word (start of input or right after whitespace), or immediately after the attachment symbol (`@"my file`). A quote embedded mid-word — the apostrophe in a contraction — no longer opens quote state.

## Verified
- Traced through the state machine by hand for the tricky cases: `@"it's a file"` (apostrophe inside an already-open quote — correctly ignored since the `elif quote:` branch is checked before the open-check), `@'unfinished` (quote immediately after symbol — correctly opens), and `don't @target` (mid-word apostrophe — correctly doesn't open). All match the new tests.
- `test_contraction_prose_does_not_hide_active_reference` covers 5 contraction cases (`don't`, `it's`, `can't`, `they're`, `isn't`) — all real regressions of the bug being fixed.
- `test_quotes_still_group_attachment_tokens` confirms legitimate quoting (`@"my file.txt"`) and closed-quote suppression (`read @"my file.txt" done` → no active reference) still work, so the fix doesn't regress the quoted-path behavior from `029ca24`.
- Ran the targeted suite post-rebase: `tests/command_line/`, completion-related messaging/plugin tests, and `test_completions_and_small_modules.py` — **1134 passed**.
- Full `tests/` run: 6 failures, all in files this PR never touches, reproducible identically against unpatched `main` — pre-existing/environmental, not this PR's doing.
- `ruff format --check` clean on both changed files.

No blocking issues. Approving.

PR #919, opened 2026-09-07
fix(mcp): accept environment-based endpoints without exposing resolved URLs

breedx:
MCP endpoint templates can fail validation even when their environment variables resolve to valid HTTP URLs. Validate the resolved value so these configurations work, without including the resolved endpoint in validation errors or replacing the stored template.

## Problem

Startup expands environment variables, but registry validation checks the unexpanded URL. A configuration such as `${MCP_BASE}/mcp` is rejected before startup even when `MCP_BASE` contains a valid HTTP(S) base URL.

## Change

- Reuse startup's existing environment expansion before the HTTP/SSE scheme check.
- Preserve the original template in memory and persisted configuration.
- Keep validation errors generic: no resolved endpoints or environment values.
- Cover actual register/persist/reload and invalid-registration error paths, not only the validation helper.

No connector lifecycle or schema changes; this patch is independent of other pending MCP fixes. It does not change configuration file reading/writing policy (#802 addresses separate configuration I/O).

## Validation

Base: `1d25d696`; Linux / Python 3.13.13 with the unchanged upstream lock. Tests use disposable HOME/XDG, no inherited credentials and blocked socket connect/DNS/bind.

- Regression file on unpatched base: **6 failed, 8 passed**. HTTP/SSE template acceptance and actual registration fail before this patch.
- `python -m pytest -q -o addopts= tests/mcp/test_expanded_url_validation.py tests/mcp/test_registry_comprehensive.py tests/mcp/test_registry_coverage.py tests/mcp/test_registry_sync_drops_unconfigured.py`: **82 passed**.
- `python -m pytest -q -o addopts= tests/mcp`: **487 passed, 6 skipped**, no warnings. The skips cover explicitly unimplemented MCP search.
- Ruff lint, formatting and diff checks pass on changed files.

## Compatibility and limits

No dependency or package-version changes. This retains the existing HTTP(S) scheme-prefix policy; it is not full URL validation, SSRF protection, authorization, or a guarantee that unresolved variables elsewhere in an otherwise prefixed URL will be rejected. Environment changes between validation and startup remain possible. No live-connector qualification is claimed.

The base currently fails the unrelated tilde-completion assertion corrected separately in #916. This PR neither incorporates that change nor claims the full upstream suite is green.

### DIFF
--- a/code_puppy/mcp_/registry.py
+++ b/code_puppy/mcp_/registry.py
@@ -14,7 +14,7 @@ from typing import Dict, List, Optional
 
 from code_puppy import config
 
-from .managed_server import ServerConfig
+from .managed_server import ServerConfig, _expand_env_vars
 
 # Configure logging
 logger = logging.getLogger(__name__)
@@ -262,13 +262,14 @@ class ServerRegistry:
                 errors.append(
                     f"{server_type.upper()} server URL must be a non-empty string"
                 )
-            elif not (
-                server_config["url"].startswith("http://")
-                or server_config["url"].startswith("https://")
-            ):
-                errors.append(
-                    f"{server_type.upper()} server URL must start with http:// or https://"
-                )
+            else:
+                # Match startup's expansion, but retain the template in config.
+                # Expanded values can contain secrets: never echo them in errors.
+                expanded_url = _expand_env_vars(server_config["url"])
+                if not expanded_url.startswith(("http://", "https://")):
+                    errors.append(
+                        f"{server_type.upper()} server URL must start with http:// or https://"
+                    )
 
             # Optional parameter validation
             if "timeout" in server_config:
--- a/tests/mcp/test_expanded_url_validation.py
+++ b/tests/mcp/test_expanded_url_validation.py
@@ -33,3 +33,42 @@ def test_invalid_expansion_is_rejected_without_echoing_it(monkeypatch, value):
     assert errors
     assert "SECRET" not in str(errors)
     assert "MCP_TEST_BASE" not in str(errors)
+
+
+@pytest.mark.parametrize("transport", ["http", "sse"])
+def test_registration_persists_template_not_expanded_endpoint(
+    monkeypatch, tmp_path, transport
+):
+    monkeypatch.setenv("MCP_TEST_BASE", "https://example.invalid/private-marker")
+    path = tmp_path / "registry.json"
+    registry = ServerRegistry(str(path))
+    template = "${MCP_TEST_BASE}/mcp"
+    config = ServerConfig(
+        id="test", name="test", type=transport, config={"url": template}
+    )
+    assert registry.register(config) == "test"
+    persisted = path.read_text()
+    assert "private-marker" not in persisted
+    assert template in persisted
+    restored = ServerRegistry(str(path)).get("test")
+    assert restored is not None
+    assert restored.config["url"] == template
+
+
+@pytest.mark.parametrize("transport", ["http", "sse"])
+def test_registration_error_does_not_disclose_expansion(
+    monkeypatch, tmp_path, caplog, transport
+):
+    monkeypatch.setenv("MCP_TEST_BASE", "file:///private/SECRET")
+    path = tmp_path / "registry.json"
+    registry = ServerRegistry(str(path))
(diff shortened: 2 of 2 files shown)


breedx:
This fix remains independently reviewable, but full CI hits the existing quoted-tilde completion assertion already addressed separately in #916. I reproduced that exact failure on unpatched `1d25d696`; this PR does not change the completion code or test, so I am keeping the unrelated correction out of this patch.

- The macOS suite's only failure is `tests/test_completions_and_small_modules.py::TestFilePathCompleterMissedLines::test_tilde_prefix_display`: the insertion contains quoting around `~/...`, so `.startswith("~")` fails.
- Reproduction on unpatched base: `python -m pytest -q -o addopts= tests/test_completions_and_small_modules.py::TestFilePathCompleterMissedLines::test_tilde_prefix_display` → **1 failed**, same assertion.
- The independently reviewed focused and broader local results remain documented in the PR body. No checks were suppressed, and no green-full-suite claim is made.
- CI again uses Python 3.14.7 despite the job's 3.13 label. Its deprecation/resource/coroutine warnings were inspected, but not all have independent base controls; local validation used Python 3.13.13.

Once the separate completion correction lands, this topic can be requalified on the updated base without widening its scope.

thomwebb (approved):
Clean, narrowly-scoped fix, and the "don't leak the resolved value" claim is actually backed by tests that check the real output surfaces rather than just asserting it in prose.

## What I checked
- Full diff on `registry.py` (10 lines) + the new test file
- Ran the PR's exact targeted test command locally on Python 3.13 / macOS (same combo as the failing CI leg): `82 passed`, matching the PR description exactly
- Confirmed `_expand_env_vars()` is the **same helper startup already uses** (imported from `managed_server.py`), not a re-implementation — no risk of validation and runtime silently drifting apart on what counts as "resolved"
- Confirmed `server_config["url"]` is never mutated to the expanded value; only a local `expanded_url` is used for the scheme check, so the template (not the secret-bearing resolved URL) is what gets persisted

## The security-relevant tests are genuinely good
`test_registration_error_does_not_disclose_expansion` sets an env var containing a `SECRET` marker, forces a validation failure, and asserts the marker shows up in **neither the raised exception text nor `caplog.text`**. `test_registration_persists_template_not_expanded_endpoint` writes to a real temp file and asserts the marker isn't in the **persisted file contents**, only the literal template. That's the right way to prove a "don't leak X" property — check the actual output surfaces, not just that a function returns the right type.

## Ran down the CI failure — confirmed unrelated
`test (macos-latest, 3.13)` fails on `test_completions_and_small_modules.py::test_tilde_prefix_display`, a `~`-path-expansion assertion in an unrelated subsystem. Verified two ways:
1. Reverted `registry.py` to `main` inside this same worktree/venv and reran — identical failure, so it's not caused by this diff.
2. This matches what the PR description already flags: a known pre-existing failure tracked separately in #916.

So the red X is real but not this PR's fault — good on the author for calling it out proactively instead of letting a reviewer discover it and get suspicious.

No blocking issues. Approving.

thomwebb (changes requested):
Update: I dismissed my earlier approval. The code change itself is still solid (see my prior review comment for the detailed walkthrough — helper reuse, no-leak tests, template persistence all check out), but I want CI green before this merges rather than "reviewer decided the specific failure is unrelated."

**On the failing `test (macos-latest, 3.13)` check:** it's failing on `test_completions_and_small_modules.py::test_tilde_prefix_display`, and I've confirmed that's a pre-existing bug on `main` itself, unrelated to this diff — checking out this PR's `registry.py` back to `main` in the same test run reproduces the identical failure. So this isn't something introduced here.

**Heads up on the fix path, since it's not obvious:** the actual fix lives in #916, which hasn't merged yet (it's got its own conflict pending). A plain rebase of this branch onto current `main` won't turn CI green, because `main` still carries the bug until #916 lands — you'd be rebasing onto a `main` that fails the same way. The real order is: #916 merges first, then this branch rebases onto the updated `main` and picks up the fix.

Once CI is clean I'm happy to re-approve — the underlying change here hasn't changed my assessment. Requesting changes just to keep the gate honest until that dependency clears.

PR #922, opened 2026-09-07
feat(mcp): wire logging, progress, roots and instructions

venkat-raman17:
Code Puppy consumes MCP tools only. MCPToolset already supports these, so they are enabled by an optional per-server `capabilities` block in mcp_servers.json.

Logs and progress route into the existing per-server file that `/mcp logs` reads, so there is no new UI and nothing under command_line/. Progress answers "is this slow tool call hung?", which stdio servers give no signal for today.

Logging and progress default on since they only touch the log file; `instructions` is opt-in because it costs prompt tokens. Bad config keys fall back to defaults rather than failing startup.

Also adds safe accessors for handshake metadata, which otherwise raise AttributeError until the toolset is entered.

### DIFF
--- a/code_puppy/mcp_/capabilities.py
+++ b/code_puppy/mcp_/capabilities.py
@@ -1,0 +1,189 @@
+"""Per-server MCP capability wiring.
+
+Turns a server's ``capabilities`` config block into ``MCPToolset`` kwargs.
+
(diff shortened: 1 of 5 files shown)


venkat-raman17:
@mpfaffenberger , please review the #PR. Thank you.

PR #925, opened 2026-09-08
fix(session): preserve agent identity across headless resume

breedx:
Resuming a conversation in a fresh process currently gives the agent a new identity, even when its message history is restored. Persist identity with that history so headless runners and CLI resume continue the same agent, and reject conflicting explicit identities instead of silently switching ownership.

## Related upstream work

- Replaces the **identity portion of #907**, which was closed without merging. This uses structured message metadata and the common `set_message_history` resume path—not identity recovery from rendered prompt text or a single frontend.
- #832 (assembled instruction delivery) and #839 (first-turn prompt preparation) overlap the **prefix companion's** preparation/delivery paths. Neither is required for this identity-only patch, and their capability refactors are not bundled here.
- The structural prompt-prefix companion **#926** builds on this identity foundation and is a dependent draft. Identity alone does not freeze the system prompt or guarantee cache-prefix stability.
- #747 propagates session IDs into hooks. That is an adjacent contract: this patch preserves `BaseAgent.id` across persisted histories; it does not equate that ID with every hook run/session ID or implement hook propagation.
- #823 concerns migration/retention of legacy pickle files. This patch leaves migration and file-retention policy unchanged and works through the existing history loaders.

## Change

- Add validated, optional runner-owned identity initialization before the conversation starts.
- Carry versioned identity state on real `ModelRequest.metadata`, preserving unrelated metadata.
- Restore centrally through `set_message_history`; a saved ID wins over a randomly generated process default, while a conflicting explicitly initialized ID fails before replacing active history.
- Preserve identity through automatic history processing and manual compaction.
- Keep the cached model agent when identity is unchanged; invalidate it when identity changes.
- Enclose restoration in the existing `/quick-resume` and `/load_context` error boundaries. The existing interactive picker error boundary is exercised too.

The small `command_line/session_commands.py` change corrects existing load-error handling rather than adding a command. Given the plugin-first guidance, please flag a preferred supported seam before landing if this should be routed differently; the common history restore remains the authoritative identity boundary.

## Validation

Base `ce1fe372` (0.0.827), Linux / Python 3.13.13, unchanged upstream dependency lock in a dedicated environment. Tests use disposable HOME/XDG and no inherited credentials; socket connect/DNS/bind is blocked, including in subprocess fixtures.

- Two new behavioral controls run unchanged on unpatched upstream and **both fail**: named-session and raw-pickle resumes produce three different IDs across three fresh processes. Candidate **2/2 pass**, with message counts increasing 2 → 4 → 6 and no caller-side identity repair.
- Existing identity/loading regression selection: **32 passed**, covering explicit conflicts, invalid IDs, cache preservation/invalidation, manual compaction, real quick-resume and picker failure containment.
- Broad agents/commands/CLI/headless/subagent selection: **1600 passed, 1 failed, 4 warnings**. The sole failure is the unchanged `test_open_project_and_select_session` assertion requiring `TODAY`: the fixture subtracts one/two hours from the wall clock, so immediately after midnight both entries are correctly labeled yesterday. The exact failure reproduces on unpatched upstream. No tests were skipped or rewritten to hide it.
- Changed-file Ruff lint, formatting and diff checks pass.

Two stream-callback coroutine warnings and the CLI test's unawaited `main` coroutine reproduce on base. Quick-resume still invokes the existing deprecated autosave setter; this patch does not suppress its warning. This is scoped evidence, not a green full-suite or live-provider claim.

## Compatibility and boundaries

Existing metadata-free histories keep the current/default identity on first restore; after the next checkpoint, the identity is persisted. This deliberately does not infer an old identity from prose. It does not change session file names, formats, migration policy, retry custody, hooks, MCP behavior, dependencies or package versions. New metadata travels with the existing message-history serialization. No prompt-caching or measured cost-saving claim is made by the identity patch alone.

### DIFF
--- a/code_puppy/agents/_compaction.py
+++ b/code_puppy/agents/_compaction.py
@@ -388,6 +388,9 @@ def make_history_processor(agent: Any) -> Callable[..., Any]:
         # Claude. Cheap no-op when all IDs already conform.
         cleaned = sanitize_tool_call_ids(cleaned)
 
+        from code_puppy.agents._session_state import stamp_session_state
+
+        cleaned = stamp_session_state(agent, cleaned)
         agent._message_history = cleaned
 
         on_message_history_processor_end(
--- a/code_puppy/agents/_compaction.py
+++ b/code_puppy/agents/_compaction.py
@@ -150,7 +150,11 @@ def run_compaction_sync(strategy: Any, messages: List[ModelMessage], *, model: M
     from concurrent.futures import ThreadPoolExecutor
 
     def _run():
-        return asyncio.run(compact_now(strategy, list(messages), model=model))
+        from code_puppy.agents._session_state import read_session_state, stamp_state
+
+        state = read_session_state(messages)
+        result = asyncio.run(compact_now(strategy, list(messages), model=model))
+        return stamp_state(state, result) if state is not None else result
 
     try:
         asyncio.get_running_loop()
--- a/code_puppy/agents/_session_state.py
+++ b/code_puppy/agents/_session_state.py
@@ -36,7 +36,10 @@ def stamp_session_state(agent: Any, messages: list[Any]) -> list[Any]:
     export = getattr(agent, "get_session_state", None)
     if not callable(export):
         return messages
-    state = export()
+    return stamp_state(export(), messages)
+
+
+def stamp_state(state: dict[str, Any], messages: list[Any]) -> list[Any]:
     for index in range(len(messages) - 1, -1, -1):
         message = messages[index]
         if isinstance(message, ModelRequest):
--- a/code_puppy/agents/_session_state.py
+++ b/code_puppy/agents/_session_state.py
@@ -1,0 +1,48 @@
+"""Conversation state carried by real message metadata, never parsed from prose."""
+
+import re
+from dataclasses import replace
+from typing import Any
+
+from pydantic_ai.messages import ModelRequest
+
+SESSION_KEY = "code_puppy_session_v1"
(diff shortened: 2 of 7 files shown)


breedx:
The new resume regressions pass in upstream CI, but the full suite hits an existing midnight-sensitive session-browser assertion. I reproduced the same failure on unpatched `ce1fe372`; it is a wall-clock fixture issue, not a failure of persisted identity or structural prompt restoration.

- Quality and Windows encoding checks passed for both #925 and #926.
- Full macOS results: #925 **7,850 passed, 1 failed**; #926 **7,871 passed, 1 failed**. Each reports 14 skips and 28 warnings.
- The only failing test is `tests/command_line/test_autosave_menu.py::TestSessionBrowser::test_open_project_and_select_session`.
- Its fixture uses `datetime.now() - timedelta(hours=1)` and `hours=2`, then unconditionally expects `TODAY`. Before 01:00, both entries are on the previous date, so the browser correctly displays yesterday.
- The unchanged test fails on bare upstream too. Neither contribution changes that test or session-browser date grouping; no failure is suppressed or bundled into the resume fixes.

The CI test environment is Python 3.14.7 despite its 3.13 label; local controls were Python 3.13.13. Deprecation/resource/coroutine warnings were inspected, but not all CI warnings have independent base controls. The full suite is not claimed green. #926 remains a dependent draft awaiting #925 acceptance and subsequent requalification.

breedx:
Superseded by the combined resume proposal #926, which already contains every identity commit from this PR along with structural prompt persistence. Keeping one review and landing target for the complete user-facing behavior. All commits are preserved; this closure is not a merge and does not discard the identity implementation. Please continue review on #926.

PR #935, opened 2026-09-11
fix(mcp): make /mcp trust discoverable in tab completion

AndrewTilson:
## Problem

`/mcp trust` is routed in `MCPCommandHandler._commands` and documented in `/mcp help`, but it never appeared in tab completion. Typing `/mcp ` and pressing Tab listed 14 subcommands — `trust` was not among them.

The practical effect is that the command is undiscoverable. A user hit exactly this and concluded the command did not exist, or that they were using it wrong.

That matters a little more than a typical missing-completion bug: `/mcp trust` is the gate for project-level MCP configs (`<CWD>/.code_puppy/mcp_servers.json`), which can declare `stdio` servers that run arbitrary commands. A user who cannot find the command may never trust a legitimate project config, and will not understand why their project's MCP servers silently fail to load.

## Root cause

`MCPCompleter.__init__` hand-maintains two dicts, `server_subcommands` and `general_subcommands`, that duplicate the handler's routing table. `trust` was added to the handler but never to the completer, so the two catalogues drifted.

There is in fact a third catalogue — `help_command.py` — which had also drifted internally: it advertised `/mcp trust [accept|revoke]` while `TrustCommand.execute` also accepts `status`.

## Changes

- **`mcp_completion.py`** — add `trust` to `general_subcommands`.
- **`mcp/help_command.py`** — correct the usage string to `[accept|revoke|status]`.
- **`tests/command_line/test_mcp_completion.py`** — pin the regression and the class of bug behind it.

Production diff is 2 lines.

## Why a test instead of a shared registry

The obvious refactor is a single declarative subcommand registry imported by both the completer and the handler. That was considered and rejected:

- It would not actually be a single source of truth — routing stays in `handler.py`, help text stays in `help_command.py`.
- `MCPCommandBase.__init__` calls `get_mcp_manager()`. A registry holding command classes would drag the whole command-module graph, and MCP runtime initialisation, into building a passive completion list.

So production completion code deliberately does **not** import the handler. The two lists stay independent, and a test enforces the invariant instead. The parity test constructs the handler only in tests, with `get_mcp_manager` patched out.

The tests assert, against the handler's routing table:

1. **Exact bidirectional set equality**, minus an explicit `{"list"}` exemption (bare `/mcp` already runs the list dashboard). Bidirectional catches both a routed command missing from completion *and* a stale completion entry that no longer routes anywhere.
2. **The exemption is itself still routed** — so drift cannot be silenced by widening the exemption set.
3. **Server-bucket membership.** Bucket assignment is behavioural, not cosmetic: it decides whether `/mcp <cmd> <TAB>` offers server names. `all_subcommands` merges both dicts, so the parity test alone cannot see which bucket a name landed in.
4. **Bucket disjointness** — the merge is `{**server, **general}`, so an overlap would silently drop server-argument behaviour.

## Verification

- Full suite (`tests/command_line/` + `tests/mcp/`): **1462 passed, 6 skipped**.
- Behavioural: `/mcp ` now offers `trust`; `/mcp tr` completes to `trust`; `list` still correctly excluded; `/mcp trust <TAB>` yields `[]`.
- Mutation-tested — each new test was confirmed to actually fail when the defect it guards is reintroduced:

  | Mutation | Result |
  |---|---|
  | Remove `trust` from the completer | 2 tests fail |
  | Move `trust` into the server bucket | bucket test fails |
  | Add a completion entry routing nowhere | parity test fails |
  | Add a bogus exemption for a dead command | exemption guard fails |

## Out of scope

- **Second-level completion for `/mcp trust <accept|revoke|status>`.** The reported bug is first-level discoverability, and the completer has no argument-completion path for general subcommands. Worth noting for anyone who picks this up: `/mcp trust <TAB>` currently yields nothing, which means nobody can two-tap their way into `accept`. If arg completion is added later, `accept` should stay non-completable — it authorises arbitrary command execution.
- **Unifying help-text generation.** Help output is richly formatted and categorised; forcing it to share a source with `display_meta` strings would make the abstraction worse. It remains a third uncovered catalogue.

### DIFF
--- a/code_puppy/command_line/mcp/help_command.py
+++ b/code_puppy/command_line/mcp/help_command.py
@@ -112,7 +112,7 @@ class HelpCommand(MCPCommandBase):
             help_lines.append(
                 Text("/mcp trust", style="cyan")
                 + Text(
-                    " [accept|revoke]  Trust this repo's .code_puppy/mcp_servers.json"
+                    " [accept|revoke|status]  Trust this repo's .code_puppy/mcp_servers.json"
                 )
             )
             help_lines.append(
--- a/code_puppy/command_line/mcp_completion.py
+++ b/code_puppy/command_line/mcp_completion.py
@@ -50,6 +50,7 @@ class MCPCompleter(Completer):
             "search": "Search for available MCP servers",
             "silence-warning": "Silence the 'registered but not bound' MCP warning forever",
             "unsilence-warning": "Restore the 'registered but not bound' MCP warning",
+            "trust": "Manage trust for this repo's .code_puppy/mcp_servers.json",
             "help": "Show help for MCP commands",
         }
 
--- a/tests/command_line/test_mcp_completion.py
+++ b/tests/command_line/test_mcp_completion.py
@@ -54,6 +54,16 @@ class TestMCPCompleter:
         # "list" is intentionally not offered: bare /mcp already does that.
         assert "list" not in names
 
+    def test_trust_is_offered(self):
+        """`/mcp trust` must be discoverable via completion.
+
+        It shipped routable-but-uncompletable, so users concluded it did not
+        exist. Keep this alongside the parity test below: this one pins the
+        specific regression, that one catches the whole class.
+        """
+        names = [c.text for c in self._get_completions("/mcp ")]
+        assert "trust" in names
+
     @patch.object(
         MCPCompleter, "_get_server_names", return_value=["server-a", "server-b"]
     )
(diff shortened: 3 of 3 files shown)


WSxDemise (dismissed):
LGTM! Clean and surgical fix.

- Added missing `trust` subcommand to `MCPCompleter.general_subcommands` with accurate description.
- Updated `/mcp trust` help text in `help_command.py` to document the supported `status` argument alongside `accept` and `revoke`.
- Comprehensive regression and parity tests against the handler routing table to prevent future catalogue drift between completer and command handler.

WSxDemise (approved):
Re-reviewed commit `c9e3752c` (comment/docstring cleanup).

- Verified docstrings and comments now focus strictly on non-obvious rationale.
- Verified test suite passes locally (`17 passed in 5.47s`).
- Verified ruff check and format pass cleanly.
- Re-approving.

PR #940, opened 2026-09-13
fix(mcp): read Tool.input_schema instead of deprecated inputSchema

thomwebb:
## Summary

MCP SDK v2 renamed `Tool.inputSchema` → `Tool.input_schema`. With fastmcp 4
installed, reading the camelCase name emits:

```
FastMCPDeprecationWarning: Accessing `Tool.inputSchema` is deprecated; MCP SDK
v2 renamed this field to `input_schema`. Update your code to read
`.input_schema` instead.
```

Code Puppy printed this at every startup from `code_puppy/mcp_/toolset_utils.py:76`
(and a second, quieter copy of the same read lived in
`code_puppy/mcp_/managed_server.py:164`).

## Where the warning actually comes from

Not reproducible on the committed lockfile — confirmed. The lockfile pins
`mcp 1.27.1` + `fastmcp-slim 3.4.7`, where `mcp.types.Tool` has only the plain
pydantic field `inputSchema`, no `input_schema`, and no warning anywhere.

The report came from an **installed** Code Puppy running MCP SDK v2:

| Package | Reporter's env | Lockfile |
|---|---|---|
| `mcp` | **2.2.0** | 1.27.1 |
| `fastmcp-slim` | **4.0.3** | 3.4.7 |
| `pydantic-ai-slim` | 2.35.0 | 2.35.0 |

In fastmcp 4, `fastmcp/_compat.py:install()` monkey-patches the SDK v2 model
classes with a **warn-once property** that routes a camelCase read to the
snake_case field and emits `FastMCPDeprecationWarning` on first access — that
property is what fires. So this PR is both a **real fix for any environment on
MCP SDK v2 / fastmcp 4** (like the reporter's) and **forward-compat prep** for
when the lockfile catches up.

## Changes

- Add a shared compatibility helper `tool_input_schema(mcp_tool)` in
  `code_puppy/mcp_/toolset_utils.py`. It prefers the new snake_case
  `input_schema` attribute and falls back to the legacy camelCase
  `inputSchema` for SDK v1.
- Route **both** existing schema reads through the helper, so the deprecation
  knowledge lives in exactly one place (DRY) instead of being duplicated.
- On SDK v2 the deprecated property is never touched, so no warning fires.
  On SDK v1 the fallback preserves existing behavior (plain field, no
  warning exists there).

## Tests

- New `tests/mcp/test_toolset_utils.py` pins the helper contract directly:
  precedence, legacy fallback, missing-schema, and a **hostile non-touch
  test** whose camelCase property raises `AssertionError` — verified to fail
  against a deliberately leaky helper that reads the deprecated field in
  addition to the new one.
- Updated `tests/mcp/test_managed_server.py` to use realistic objects (the
  new v2 field), plus a new test covering the legacy fallback path.

All `tests/mcp` tests pass (`538 passed, 6 skipped`), and `ruff check` +
`ruff format` are clean.

## Notes

Intentionally no behavioral change beyond silencing the warning — schema
resolution semantics are identical.


AndrewTilson (commented):
## Review: `fix(mcp): read Tool.input_schema instead of deprecated inputSchema`

Nice tight, single-purpose PR — the DRY consolidation into one `tool_input_schema()` helper is the right call, and routing **both** call sites through it is exactly what I'd want instead of two divergent copies of the same compat hack. I checked this out locally against the locked deps and verified the claims rather than taking them on faith.

### Verified claims

| Claim | Result |
|---|---|
| `tests/mcp` pass | PASS - `538 passed, 6 skipped` — matches the description exactly |
| `ruff check` clean | PASS - `All checks passed!` (on the pinned `ruff>=0.15,<0.16`, resolved to 0.15.22) |
| `ruff format` clean | PASS - `654 files already formatted` |
| No import cycle from the new `managed_server` → `toolset_utils` import | PASS - imports fine |
| No other unconverted `inputSchema` reads left in `code_puppy/` | PASS - only docstrings/comments remain |

Also nice: the guard is `if schema is not None` rather than a truthiness check, so a legitimately empty `{}` schema doesn't silently fall through to the legacy attribute. Easy detail to get wrong, and it's got a test (`test_explicit_none_schema_falls_through_to_legacy`).

---

### Main finding: the warning doesn't appear to be reproducible on the locked deps

This is the one I'd like the author's read on, because it affects how the PR description should be worded.

With the current lockfile (`pydantic-ai-slim==2.35.0` → `mcp 1.27.1`, `fastmcp-slim 3.4.7`):

```
>>> import mcp.types as t
>>> list(t.Tool.model_fields)
['name', 'title', 'description', 'inputSchema', 'outputSchema', 'icons', 'annotations', 'meta', 'execution']
>>> hasattr(t.Tool, 'input_schema')
False
```

And the tools being iterated really are those objects — `pydantic_ai/mcp.py:853` declares `_cached_tools: list[mcp_types.Tool] | None`, where `mcp_types` is `mcp.types`.

Two consequences:

1. A real `mcp.types.Tool` has **no** `input_schema` attribute, so `tool_input_schema()` falls through to the legacy `getattr(mcp_tool, "inputSchema", None)` branch on *every* call today. The deprecated read still happens.
2. Grepping the whole installed venv for the warning text (`"renamed this field"`) and for any `def inputSchema` deprecated property returns **zero** hits. Reading `.inputSchema` off a real `Tool` under `-W error::DeprecationWarning` raises nothing.

So on `main`'s pinned dependency set I can't reproduce the `FastMCPDeprecationWarning` the PR is described as fixing. That's **not a reason to block** — the change is genuinely forward-compatible and will do the right thing the moment the SDK bump lands. But "Every Code Puppy startup printed this" and "On SDK v2 the deprecated attribute is never touched, so no warning fires" read as *present tense*, and neither is true of the committed lockfile.

**Suggestion:** either (a) reword the description to frame this as forward-compat prep for the upcoming SDK bump, or (b) if you're reproducing the warning locally, note the environment (a newer/pre-release `fastmcp` or `mcp` pinned outside the lock?) — that'd be useful context for whoever revisits this.

---

### `test_does_not_touch_deprecated_field_when_new_present` doesn't test what its docstring claims

> ```python
> """A ``Mock`` explodes on any attribute access, so if the helper wrongly
> probed ``inputSchema`` first (or in addition), this would raise or fail."""
> ```

A plain `Mock()` is the opposite of explosive — it happily auto-creates any attribute you touch:

```python
>>> m = Mock(); m.input_schema = {'type': 'object'}
>>> m.inputSchema
<Mock id=...>          # no raise, just a child Mock
```

I confirmed the "or in addition" half of the claim is not covered. This deliberately leaky helper — which *does* read the deprecated field on every call, i.e. the exact bug this test is supposed to catch — **passes** the assertion as written:

```python
def leaky_helper(mcp_tool):
    legacy = getattr(mcp_tool, "inputSchema", None)   # fires the warning!
    schema = getattr(mcp_tool, "input_schema", None)
    return schema if schema is not None else legacy
```

The "probed first" half *is* caught, but that's already fully covered by `test_new_field_wins_when_both_present`. So as it stands this test is effectively a duplicate precedence assertion wearing a non-touch costume.

If you want a real non-touch guarantee, make the deprecated attribute actively hostile:

```python
def test_does_not_touch_deprecated_field_when_new_present():
    """Reading the deprecated attribute is what triggers the warning."""
    schema = {"type": "object"}

    class StrictTool:
        input_schema = schema

        @property
        def inputSchema(self):  # noqa: N802 - mirrors the SDK's camelCase field
            raise AssertionError("deprecated `inputSchema` was read")

    assert tool_input_schema(StrictTool()) == schema
```

Verified: this passes against the helper as implemented, and fails against `leaky_helper`. Note that `Mock(spec=["input_schema"])` does **not** work here — `getattr(..., None)` swallows the resulting `AttributeError`, so the leaky version still sneaks through.

---

### Minor nits (non-blocking)

**1. `Optional[Any]` is redundant.** `Any` already subsumes `None`, so the `Optional` wrapper adds no information to a reader or a type checker. Given `coerce_tool_args` immediately does `isinstance(input_schema, dict)`, the honest signature is narrower and more useful:

```python
def tool_input_schema(mcp_tool: Any) -> Optional[Dict[str, Any]]:
```

*Explicit is better than implicit.*

**2. Stale docstring nearby.** `managed_server.py:143` still says "Best-effort lookup of an MCP tool's JSON inputSchema" — worth updating to `input_schema` while you're in the neighborhood, since the whole point of the PR is that the camelCase name is on its way out. Same for the two references in `tool_arg_coercion.py` (lines 8 and 48), though those are further afield and fine to leave.

---

### Risk assessment: **Low**

Pure read-path compat shim, no behavioral change, both call sites covered, new tests added, full `tests/mcp` suite green. Worst realistic case is that it's currently a no-op that becomes correct after the SDK bump.

### Summary

Approach and structure are solid — I'd merge this happily once the test above is tightened. Requests, in priority order:

1. Strengthen `test_does_not_touch_deprecated_field_when_new_present` (or drop it as redundant, but tightening is better) — right now it can't fail for the reason it claims to exist.
2. Reword the PR description to match what's reproducible on the committed lockfile.
3. Optional: tighten the return type and refresh the stale `inputSchema` docstrings.

*Reviewed by Luna*

thomwebb:
Great review — all three asks addressed in 56036230. Thanks especially for catching the Mock hole; you were completely right.

**1. Non-touch test tightened (your top ask).** Replaced the `Mock()` double with the hostile-property pattern you sketched:

```python
class StrictTool:
    input_schema = schema

    @property
    def inputSchema(self):  # camelCase mirrors the SDK's field name
        raise AssertionError("deprecated `inputSchema` was read")
```

Verified both directions locally: the real helper passes, and a deliberately leaky helper (`legacy = getattr(mcp_tool, "inputSchema", None)` before the snake_case read — the exact bug the test exists for) fails with `AssertionError: deprecated inputSchema was read`. Dropped the now-unused `Mock` import. (Went with a plain comment rather than `# noqa: N802` since pep8-naming isn't in the enabled rule set — a noqa for a disabled rule felt like dead weight, happy to flip it if you'd rather future-proof.)

**2. The warning is real — environment documented in the description.** Your repro work was correct, and pointed me exactly where to look. The report came from an **installed** Code Puppy (`~/.cache/uv/archive-v0/...`), whose environment differs from the lockfile:

| Package | Reporter's env | Lockfile |
|---|---|---|
| `mcp` | **2.2.0** | 1.27.1 |
| `fastmcp-slim` | **4.0.3** | 3.4.7 |
| `pydantic-ai-slim` | 2.35.0 | 2.35.0 |

In fastmcp 4, `fastmcp/_compat.py:install()` patches the SDK v2 model classes with a warn-once property that routes `inputSchema` reads to `input_schema` and emits `FastMCPDeprecationWarning` on first access — that's the property firing in the report. So on the lockfile this PR is forward-compat prep (as you concluded), and on any mcp 2.x / fastmcp 4 environment it's a live fix. The description now says exactly that, with the table.

**3. Nits applied.** `Optional[Any]` → `Optional[Dict[str, Any]]` (and noted the `fastmcp/_compat.py` bridge in the docstring), plus refreshed the stale `inputSchema` wording at `managed_server.py:143` and the `mcp_tool.inputSchema` reference in `tool_arg_coercion.py`'s docstring.

`tests/mcp` still green (538 passed, 6 skipped); `ruff check` + `ruff format` clean.

AndrewTilson (approved):
## Follow-up review: commit `56036230`

All three asks addressed, and the environment claim checks out. I re-verified the delta rather than taking the summary on faith — including reproducing the original warning end-to-end, which I couldn't do last round.

### Ask 1 — non-touch test: RESOLVED, and I mutation-tested it

The `StrictTool` hostile-property swap is exactly right. I ran the new test against three helper variants to confirm it can actually fail now:

| Helper variant | Result |
|---|---|
| Real `tool_input_schema` | PASS - returns `{'type': 'object'}` |
| Leaky (reads deprecated field *in addition*) | CAUGHT - `deprecated inputSchema was read` |
| Deprecated-first (wrong precedence) | CAUGHT - `deprecated inputSchema was read` |

The "in addition" case is the one the old `Mock()` version silently let through. It now bites. This is a real regression guard rather than a duplicate precedence assertion, and it's environment-independent — it tests the helper's contract, not whichever SDK happens to be installed. Good.

### Ask 2 — the warning is real, and I reproduced it

Your environment table was the missing piece. I built a clean venv with `fastmcp-slim==4.0.3` (which pulls `mcp-types 2.2.0`, the separate SDK v2 package) and confirmed the whole chain:

```
Tool fields: ['name', 'title', 'description', 'input_schema', 'execution', 'output_schema', ...]
inputSchema is bridged property: True

OLD code (.inputSchema): 1 warning(s)
   -> Accessing `Tool.inputSchema` is deprecated; MCP SDK v2 renamed this field to `input_schema`. Update ...
NEW code (helper)      : 0 warning(s); returned {'type': 'object'}
```

That warning string is a **verbatim** match for the one in the PR description. I also read `fastmcp/_compat.py` from the uv cache (`fastmcp_slim-4.0.3.dist-info`) and the mechanism is precisely as you described: `install()` walks an `_ALIASES` map and calls `setattr(cls, camel, _make_property(...))`, whose getter flips a `nonlocal warned` flag and emits one `FastMCPDeprecationWarning` per `(class, name)`. My earlier "not reproducible" finding was correct *for the lockfile* and wrong as a claim about the world — thanks for pinning down the actual env instead of just asserting it.

**Bonus finding: the fix is stronger than advertised.** `_make_property`'s getter reads `fastmcp.settings.mcp_camelcase_compat` live, and raises `AttributeError` when the bridge is disabled. Verified:

```
raw .inputSchema raises AttributeError: 'Tool' object has no attribute 'inputSchema'
helper still works: {'type': 'object'} | warnings: 0
```

So for anyone running with `mcp_camelcase_compat=False`, the pre-PR code doesn't just warn — the `getattr(..., None)` swallows the `AttributeError` and silently yields `None`, disabling arg coercion entirely. This PR fixes a latent silent-wrong-behavior bug, not only a noisy log line. Might be worth a line in the description.

**I checked the other bridged fields too.** `_ALIASES` covers ~20 camelCase names across `Tool`, `ToolAnnotations`, `Resource`, `CallToolResult`, `InitializeResult`, and friends (`outputSchema`, `mimeType`, `isError`, `structuredContent`, `nextCursor`, `maxTokens`, ...). I grepped `code_puppy/` for every one of them: **zero hits**. `inputSchema` was genuinely the only bridged field this codebase touched, so the fix is complete — no sibling follow-ups lurking.

### Ask 3 — nits: applied

`Optional[Dict[str, Any]]`, the `_compat.py` pointer in the helper docstring, and the refreshed wording at `managed_server.py:143` + `tool_arg_coercion.py:48` all look good.

### Re-verified on `56036230`

| Check | Result |
|---|---|
| `tests/mcp` | PASS - 538 passed, 6 skipped |
| `ruff check` (pinned 0.15.22) | PASS - All checks passed! |
| `ruff format --check` | PASS - 654 files already formatted |

### One trivial nit (do not hold the merge for this)

The new test docstring says the bridge installs the property on `mcp.types.Tool`. Under fastmcp 4 it's `mcp_types.Tool` — SDK v2 ships as the standalone `mcp-types` distribution, and `_compat.py` itself does `import mcp_types`. In the repro venv `mcp` isn't even installed. Purely cosmetic, and only matters to whoever reads this test in a year wondering which package to look in.

### Verdict

Approving. The behavior was already correct last round; what changed is that it's now *provably* correct — the non-touch guarantee has a test that fails when violated, and the motivating bug has a documented, reproducible environment. Nice turnaround on the review feedback.

*Reviewed by Luna*

thomwebb:
Follow-up data point for the maintainers, since the natural next question after the env table is "should we re-lock to mcp 2.x?" — I tested it in a scratch worktree (re-lock only, no code changes):

```
$ uv lock --upgrade-package mcp --upgrade-package fastmcp-slim
Updated fastmcp-slim v3.4.7 -> v4.0.3
Updated mcp          v1.27.1 -> v2.2.0
Added   mcp-types    v2.2.0
Removed httpx-sse    v0.4.3
```

Result: **536 passed, 2 failed** — the only casualties are in
`tests/mcp/test_tool_failure_not_fatal.py`, which uses
`from mcp.server.fastmcp import FastMCP` (top-level import + the embedded
stdio subprocess source). mcp 2.x removed that module
(`mcp.server.mcpserver.MCPServer` is the v2 spelling).

So the runtime code is already v2-clean — I also grepped for every other
camelCase read fastmcp 4 bridges (`outputSchema`, `mimeType`, `isError`,
`nextCursor`, `serverInfo`, …) and found zero — but the bump is its own PR:
fix that one test file (or make it v1/v2-conditional), re-lock, and re-run the
full matrix. Keeping it out of this one so the shim stays "pure read-path,
no behavior change".

PR #943, opened 2026-09-15
feat(callbacks): add error_logged observation phase

AndrewTilson:
## What

Adds a synchronous `error_logged` callback phase that fires when `log_error()` records an exception, letting downstream distributions observe logged errors without core taking on any new behaviour.

## Why

Distributions that want to forward logged errors somewhere (an internal log aggregator, a support workflow) currently have no seam to hook into. The alternative is monkey-patching `log_error`, which is fragile and invisible.

## Zero-network guarantee is preserved

This repo promises no telemetry and no crash reports, and that promise is unchanged:

- **Core registers no subscriber**, so this is a no-op in a stock install.
- No network imports are added — the diff touches only `callbacks.py` and `error_logging.py`.
- The phase is an *observation* hook: core hands a subscriber the exception it already logged locally, and does nothing else.

## Design notes

**The two sinks fail independently.** `log_error()` is split into `_build_error_log_entry()` and `_notify_error_logged()` so a full disk cannot suppress the hook, and a misbehaving subscriber cannot suppress the local log. The `errors.log` format is byte-for-byte unchanged.

**The phase is synchronous.** `log_error()` is called from arbitrary contexts, including interpreter shutdown, where no event loop is guaranteed to be running. An async dispatcher would be unusable from exactly the paths that matter most.

**`log_error_message()` deliberately does not fire the phase.** A subscriber that logs its own failures would otherwise recurse forever.

**The import is lazy.** `callbacks` pulls in pydantic-ai; importing it at module scope in `error_logging` would create a cycle through `config`.

## Testing

612 pass across the error/callback/logging selection. No behavioural change with no subscribers registered.


AndrewTilson:
Companion PRs for this change (Walmart-internal, not publicly accessible):

- Frontend error-ingest endpoint: `AI-INNOVATION-LAB/puppy-frontend#590`
- Internal CLI subscriber: `AI-INNOVATION-LAB/code-puppy#658`

This PR is the only piece that lands in the public repo, and it adds no network activity — just the callback seam.

code_puppy/error_logging.py
```diff
@@ -86,6 +129,8 @@ def log_error(
         # to cause more problems than they solve!
         pass
 
+    _notify_error_logged(error, context, include_traceback)
```
AndrewTilson, line 165: ** Unguarded re-entrancy.**

A subscriber that calls `log_error()` — e.g. to report its own failure — re-enters this phase. Measured on this branch: **199 nested dispatches** before CPython's recursion limit trips, each one writing to `errors.log`, with the resulting `RecursionError` silently swallowed by the `except Exception` below.

The `log_error_message()` carve-out doesn't cover this; it closes the side door while the front door stands open.

A thread-local latch fixes it and keeps the phase re-entrant-safe per thread:

```python
import threading

_notify_state = threading.local()


def _notify_error_logged(error, context, include_traceback) -> None:
    if getattr(_notify_state, "active", False):
        # A subscriber logged an error of its own. Observe once, never recurse.
        return
    _notify_state.active = True
    try:
        from code_puppy.callbacks import on_error_logged

        on_error_logged(error, context=context, include_traceback=include_traceback)
    except Exception:
        pass
    finally:
        _notify_state.active = False
```

Thread-local rather than a module global so a subscriber on thread A can't blind the hook for an unrelated error on thread B.

code_puppy/error_logging.py
```diff
@@ -37,13 +37,79 @@ def _ensure_logs_dir() -> None:
...
+
+        on_error_logged(
+            error,
+            context=context,
+            include_traceback=include_traceback,
+        )
+    except Exception:
+        # An observer must never turn "we hit an error" into "we crashed while
+        # writing down that we hit an error".
+        pass
```
AndrewTilson: ** Fully silent catch.**

`_trigger_callbacks_sync` already isolates and logs per-subscriber failures, so in practice this only catches (a) the lazy import failing and (b) a `RecursionError` from the re-entrancy above. Both are conditions someone debugging a broken plugin would very much want to see.

The local-file sink can justify silence — it's the last resort. This one has a working logger available:

```python
    except Exception:
        # An observer must never turn "we hit an error" into "we crashed while
        # writing down that we hit an error" -- but leave a breadcrumb.
        logging.getLogger(__name__).debug(
            "error_logged dispatch failed", exc_info=True
        )
```

Errors should never pass silently; unless explicitly silenced.

code_puppy/callbacks.py
```diff
@@ -717,6 +719,50 @@ def on_file_permission(
...
+    plugin registered this is a dictionary lookup that returns ``[]``, so the
+    zero-telemetry guarantee in the README is preserved by construction.
+
+    Deliberately **synchronous**: ``log_error()`` is called from arbitrary
+    contexts, including exception handlers on the way out of the process,
+    where no event loop is guaranteed to exist. ``agent_exception`` is *not* a
+    substitute -- it is async, it fires for recovered errors, and it does not
+    cover every ``log_error()`` call site.
+
+    Note this is **not** fired by ``log_error_message()``. That function
```
AndrewTilson, line 745: ** This paragraph overstates the protection.**

> hooking it would let a reporting plugin generate reports about its own reporting

True, but incomplete — a subscriber calling `log_error()` (not `log_error_message()`) recurses today, 199 levels deep. Once the thread-local latch is in, I'd reword to make the actual guarantee explicit:

```
    Re-entrancy: dispatch is latched per-thread, so a subscriber that itself
    calls ``log_error()`` will not re-trigger this phase. ``log_error_message()``
    does not fire it at all.

    Subscribers must return promptly. Dispatch is synchronous and on the error
    path -- a blocking network call here adds its full latency to every logged
    error, including during interpreter shutdown. Queue and drain off-thread.

    The payload is unsanitised: ``error.args`` and the traceback may contain
    file paths, request bodies, or credentials. A subscriber that forwards it
    off-box owns that redaction.
```

code_puppy/callbacks.py
```diff
@@ -80,6 +80,7 @@
     "git_branch_provider",
     "feature_capability",
     "transform_model_messages",
+    "error_logged",
```
AndrewTilson, line 83: ** Add the matching row to the `AGENTS.md` phase table** (~L80-102). Every other plugin-facing phase is documented there, so as-is this one is undiscoverable to plugin authors:

```markdown
| `error_logged` | After `log_error()` writes to the local log | `(error, *, context=None, include_traceback=True) -> None` — sync observer; must return promptly |
```

AndrewTilson (commented):
## Review: `error_logged` observation phase

Nice, tight seam. The `_build_error_log_entry` / `_notify_error_logged` split is the right call, and I verified the two headline claims in the PR description actually hold:

| Claim | Verified |
|---|---|
| Full disk cannot suppress the hook |  patched `_ensure_logs_dir` to raise `OSError` — subscriber still fired |
| Misbehaving plugin cannot suppress the local log |  subscriber raising `RuntimeError` — entry still written to `errors.log` |
| No-op with no subscriber |  `_trigger_callbacks_sync` short-circuits on empty list |
| Zero-network guarantee preserved |  no new imports, core registers no subscriber |

The sync-not-async justification is also sound — `log_error()` genuinely runs during interpreter teardown where `agent_exception` would be unusable.

So the design is good. But there's one **real bug** that the PR description gets *almost* right, and a couple of gaps before this is mergeable.

---

###  1. The recursion guard is in the wrong place

The description says:

> `log_error_message()` deliberately does not fire the phase. A subscriber that logs its own failures would otherwise recurse forever.

Right diagnosis, wrong function. The recursion hazard is `log_error()` itself, and it is wide open. A subscriber that reports its own failure via `log_error` — the single most obvious thing a reporting plugin does — re-enters the phase.

I ran it against this branch:

```python
def bad_subscriber(error, *, context=None, include_traceback=True):
    # plausible: the plugin logs its own failure
    error_logging.log_error(RuntimeError("reporting failed"), context="plugin")

callbacks.register_callback("error_logged", bad_subscriber)
error_logging.log_error(ValueError("boom"))
```

```
max reentrancy depth reached: 199
```

199 nested dispatches, 199 disk writes, and it only stops because CPython's recursion limit trips — swallowed by the bare `except Exception` so the operator sees *nothing*. Note this fires on the error path, i.e. exactly when the system is already unhealthy.

Skipping `log_error_message()` doesn't help here; it closes a side door while the front door is open. Fix is a re-entrancy latch in `_notify_error_logged` (see inline comment). Once that exists, the `log_error_message` carve-out can stay as belt-and-braces, but it should stop being advertised as *the* recursion defence.

###  2. No timeout — a slow subscriber stalls the error path

Measured on this branch: a subscriber doing `time.sleep(2)` blocks `log_error()` for the full 2.00s. A plugin that does a blocking HTTP POST to an aggregator therefore adds its latency to **every logged error**, including during shutdown, where it can hang the exit. Worth an explicit contract note in the docstring ("subscribers must return promptly; do not block on network I/O — queue and drain off-thread") even if you don't enforce it in core.

###  3. Zero tests for a new public extension point

`git diff $(git merge-base main HEAD)..HEAD --name-only` → two files, neither a test. The "612 pass" figure is the pre-existing suite; nothing added here exercises the new code. Compare `tests/test_transform_model_messages.py`, which is the established bar for a callback phase in this repo.

The four behaviours in the table above are all cheap to assert, and #1 is exactly the kind of thing a regression test should have caught. A `tests/test_error_logged_phase.py` covering fire/no-op/isolation/reentrancy would close this out.

###  4. `AGENTS.md` phase table not updated

The table at `AGENTS.md:~80-102` documents every other plugin-facing phase. `error_logged` is missing, so plugin authors won't discover it. One row:

```markdown
| `error_logged` | After `log_error()` writes to the local log | `(error, *, context=None, include_traceback=True) -> None` — sync observer; must return promptly |
```

###  5. Document the PII surface for plugin authors

Core's zero-telemetry promise is untouched — agreed. But this phase hands subscribers a raw exception plus traceback, and tracebacks in this codebase routinely carry file paths, request payloads, and occasionally credentials in `error.args`. That's fine for a local file; it's a different risk class once a distro plugin ships it off-box. A sentence in the `on_error_logged` docstring warning that the payload is unsanitised would save a downstream author from an incident.

---

### Verdict

**Request changes** on #1 (reentrancy) and #3 (tests) — both are small. #2/#4/#5 are docstring-and-a-table. The underlying architecture is solid and I'd be happy to see it land once the latch is in.

<sub> Reviewed by Luna — claims re-verified by execution against `awtilso/PUP-775`, not by reading alone.</sub>

code_puppy/error_logging.py
AndrewTilson: Confirmed and fixed in `eb5d5003`. Reproduced your number exactly before changing anything — **199 nested dispatches, 199 tracebacks, 69,007 bytes** in `errors.log` from one `ValueError`.

Your framing of the `log_error_message()` carve-out is right and I had this wrong in my head: it stops a subscriber’s *forensic notes* from recursing, not a subscriber that logs a genuine exception of its own. Side door closed, front door open.

Applied the thread-local latch as written. Verified:
- nested dispatches **199 -> 1**
- latch releases between independent errors (`first`, `second` both observed — a module-global would have been fine here, but see below)
- per-thread isolation holds: thread A latched, thread B still dispatched (`{A: [from A], B: [from B]}`)

That last one is why thread-local was the right call rather than a global flag — a background worker logging an error would otherwise silently swallow the main thread’s report.

code_puppy/error_logging.py
AndrewTilson: Agreed, fixed in `eb5d5003` — now a `logging.getLogger(__name__).debug(..., exc_info=True)`.

Your analysis of what actually reaches this branch is correct, and it is exactly why I declined the same finding from `pr-agent` on the private PR: `_trigger_callbacks_sync` already isolates and logs per-subscriber failures with a full traceback, so a broken subscriber was never silent.

But you have identified the residue that argument does not cover — the lazy import failing, and (until the latch) the `RecursionError`. Those reach this `except` and vanished completely. A debug line costs nothing on the happy path and is the difference between diagnosable and not.

I kept the silent-catch reasoning in the comment and added the breadcrumb alongside it, rather than replacing one with the other.

code_puppy/callbacks.py
AndrewTilson: Applied verbatim in `eb5d5003` — the re-entrancy paragraph, the promptness requirement, and the unsanitised-payload warning.

You are right that the original overstated the protection. It described the `log_error_message()` carve-out as though it were the whole guard, which read as a stronger guarantee than the code delivered.

The promptness and redaction paragraphs are worth more than they look for an OSS extension point: this fires synchronously on the error path including during interpreter shutdown, and the payload carries raw `error.args` and traceback text. A downstream subscriber forwarding that off-box owns the redaction — which is precisely what the internal fork does, and it is better to state it here than to let the next implementer discover it.

code_puppy/callbacks.py
AndrewTilson: Added in `eb5d5003`, placed directly after `agent_exception` since that is the closest relative in the table.

Fair point that an undocumented phase is an undiscoverable one — a plugin-facing extension point that appears in no table is effectively private API that happens to be importable.

WSxDemise:
### Summary
Adds a synchronous `error_logged` observation phase fired from `error_logging.log_error()`, with no in-tree subscriber. Scope is tight (only `callbacks.py`, `error_logging.py`, docs, and a new test file), the zero-telemetry guarantee is preserved by construction (no subscriber registered, no network code added), and the re-entrancy/independent-failure design is sound and well tested (`tests/test_error_logged_phase.py` covers subscriber isolation, re-entrancy, per-thread latch scoping, and the `log_error_message` exclusion).

### Group 1 — Really Should Fix These
- `code_puppy/callbacks.py:9-84` (`PhaseType` Literal) — `"error_logged"` is added to `__all__` (line 83) and to the `_callbacks` registry dict (line 175), but **not** to the `PhaseType` Literal itself. Every other phase name in `__all__`/`_callbacks` has a matching `PhaseType` entry; this one doesn't. `register_callback`, `unregister_callback`, `clear_callbacks`, `get_callbacks`, and `_trigger_callbacks_sync` are all typed as `phase: PhaseType`, so `callbacks.register_callback("error_logged", ...)` (used in the new tests) is a Literal-value type violation under any static type checker, even though it works at runtime. Please add `"error_logged"` to the `PhaseType` Literal list to keep it consistent with the rest of the registry.

### Group 2 — Nits
- `code_puppy/error_logging.py` (`_notify_error_logged`) — the broad `except Exception` swallowing a failed dispatch only logs at `debug` level; given the docstring explicitly says the intent is to leave "a breadcrumb," `warning` (or documenting why `debug` was chosen) would make the failure easier to notice in the field. Not blocking.
- `AGENTS.md` — the new callback-phase table row is a good touch for discoverability; no issue, just confirming it matches the phase name/signature added in code.

AndrewTilson:
@WSxDemise thanks — one applied in `efcbd525`, one I believe is a false positive. Evidence for both below.

## Group 1 — `error_logged` missing from the `PhaseType` Literal — **respectfully disagree, premise looks incorrect**

The finding says `"error_logged"` was added to `__all__` (line 83) and `_callbacks` (line 175) but **not** to `PhaseType`. Checking the file:

- **There is no `__all__` in `callbacks.py`** — `grep -n '^__all__'` returns nothing.
- `PhaseType = Literal[` spans **lines 9–84**. Line 83 is the *last entry inside that Literal*, immediately before the closing `]` on line 84.

So line 83 **is** the `PhaseType` entry — I think it was read as an `__all__` entry that doesn't exist.

Confirmed at type level rather than by eye:
```
error_logged in PhaseType : True
error_logged in _callbacks: True
registry keys NOT in PhaseType: none
PhaseType names NOT in registry: none
```
The two sets are exactly equal — no phase is missing from either side, so there's no drift to fix. Happy to be shown otherwise if you're looking at a different revision.

## Group 2 — `debug` too quiet for something documented as a breadcrumb — **applied, you're right**

This was a real defect and a nice catch, because the comment and the code disagreed:

```
effective level: WARNING
debug visible in stock install : False
```

The root logger sits at `WARNING` in a stock install, so the record was discarded outright. The comment promised a breadcrumb; the code left one only for someone who had already turned on debug logging — i.e. not the person debugging this.

Promoted to `warning`, with the reasoning recorded inline so it doesn't get quietly downgraded later. It can't become noise: `_trigger_callbacks_sync` already isolates and logs per-subscriber failures, so this branch is only reachable if the dispatch itself breaks (lazy import failing, most plausibly).

Added a test asserting the record lands at `>= WARNING` and that the local file sink still runs. Verified it fails when the level is dropped back to `debug`.

## AGENTS.md

Confirmed the row matches the implemented signature.

Public tests: **620 passed, 4 skipped**.

Also mirrored to the internal fork, where this seam is kept byte-for-byte in sync and the phase has a real subscriber — a silently-broken dispatch there means crash reporting stops with no local signal.

WSxDemise:
### Follow-up review (commits `eb5d500`, `efcbd52`)

Both previously flagged items are fixed and verified against the current PR head:

- `code_puppy/callbacks.py:84` — `"error_logged"` is now present in the `PhaseType` Literal, consistent with its entries in `__all__` (line 83) and the `_callbacks` dict (line 175). No more type-registry inconsistency.
- `code_puppy/error_logging.py:128` — the failed-dispatch breadcrumb now logs at `warning` (was `debug`), matching the docstring's stated intent, with a new regression test `test_failed_dispatch_leaves_a_visible_breadcrumb` covering it.

Ran `tests/test_error_logged_phase.py` directly against this branch: 8 passed, including the new breadcrumb-visibility test.

No remaining findings from this review pass. Recommending approval.

WSxDemise (approved):

PR #944, opened 2026-09-15
Alphabetize i18n locale keys

thomwebb:
## Summary
- Alphabetize top-level keys in the en-US, es, and fr-CA localization catalogs.
- Preserve all translation values unchanged.
- Add a regression test enforcing catalog key ordering.

## Validation
- Ruff check and format passed.
- `pytest tests/i18n -q -p no:randomly`: 196 passed.


AndrewTilson (commented):
## Review: Alphabetize i18n locale keys

Reviewed at `a5b8783`. The headline number (1000 additions / 989 deletions) looks scary, but a pure-reordering PR is exactly the kind of diff where a value change can hide in plain sight — so I verified it mechanically rather than by eyeballing 2119 diff lines.

### Verification performed

Fetched the base (`c2a8304`) and head (`a5b8783`) blobs and compared the **parsed** structures, flattened to dotted leaf paths:

| Catalog | Leaves | Keys added | Keys removed | **Values changed** | Bytes base → head |
|---|---|---|---|---|---|
| `en-US.json` | 464 | 0 | 0 | **0** | 36878 → 36878 |
| `es.json` | 306 | 0 | 0 | **0** | 26455 → 26455 |
| `fr-CA.json` | 306 | 0 | 0 | **0** | 27113 → 27113 |

Additional checks, all clean:

- **Byte-for-byte canonical:** each file is exactly `json.dumps(data, indent=2, sort_keys=True, ensure_ascii=False) + "\n"`. Nothing hand-edited, nothing drifted.
- **No duplicate keys** introduced in either revision (checked with `object_pairs_hook`, since `json.load` silently swallows dupes).
- **Encoding preserved** — non-ASCII stays raw UTF-8, not re-escaped to `\uXXXX`; trailing newlines intact. That matters given the repo runs a dedicated `windows-encoding` CI job.
- **Cross-locale parity unchanged:** `es` and `fr-CA` each remain a 158-key subset of `en-US`, which is the documented intent (`docs/I18N.md` §5: omitted keys fall back).
- **The new test is not a no-op** — I confirmed it genuinely fails on the base commit (`en-US` first mismatch: `mcp.http_form.title` where `agent.model_settings.boolean_type` was expected).
- **Ran the suite locally** in a clean venv off the PR head: `pytest tests/i18n -q -p no:randomly` → **196 passed**, matching your stated validation. All three CI checks are now green.

The claim in the description holds up exactly as written. Nice, honest PR. 

---

### Suggestions on the new test (non-blocking)

The change itself is airtight; the feedback is all about `test_shipped_catalog_keys_are_alphabetized`, which is slightly weaker than it looks. I verified each of these by mutation — breaking the catalog and watching the test pass anyway.

**1. It only checks the top level, so nested dicts can rot silently.**
`list(json.load(fh))` yields top-level keys only. Both catalogs contain nested plural objects (`files.deleted`, `files.written`). I reversed `files.written` to `{"other", "one"}` — unsorted by any definition — and the test still passed.

Worth an explicit decision rather than an accident. There's a real argument for *not* recursing: CLDR plural categories have a canonical order (`zero, one, two, few, many, other`, per `plurals.py:CATEGORIES`) which is **not** alphabetical, so forcing `few` before `one` would fight the domain. If that's the intent — and I think it should be — encode it in the docstring so nobody "fixes" it later:

```python
"""Top-level catalog keys must be alphabetized.

Nested plural objects are intentionally exempt: CLDR categories have a
canonical order (zero/one/two/few/many/other) that is not alphabetical.
"""
```

Explicit is better than implicit — right now the exemption is invisible.

**2. A duplicate key sails straight through — and that's the actually scary one.**
`json.load` keeps the *last* duplicate, so a second occurrence of an existing key overrides the value while leaving the parsed key order perfectly sorted. I appended `"confirm.yes": "DUPLICATE HIJACK"` to the end of `en-US.json`; both this test and `test_all_shipped_catalogs_are_valid_json` passed, and `confirm.yes` silently resolved to the hijacked string.

That's a plausible real-world accident on a file this size (paste a key twice during a merge, ship a wrong translation, no signal). Since you're already opening and parsing the file, catching it is nearly free.

**3. Vacuous pass if the glob ever returns empty.**
`for path in glob.glob(...)` with zero matches passes trivially. The sibling `test_all_shipped_catalogs_are_valid_json` guards this with `assert files, "no shipped catalogs found"` — this one doesn't. If the locales dir ever moves or packaging drops the JSON, this test reports green.

**4. Minor DRY nit.** Three tests in this file now recompute `os.path.join(os.path.dirname(catalog.__file__), "locales")` with function-local `import glob, os`. A module-level constant would remove the duplication and the repeated imports.

A version that addresses all four, which I ran against the PR head (passes clean, and fails correctly with the duplicate injected):

```python
LOCALES_DIR = Path(catalog.__file__).parent / "locales"
CATALOG_PATHS = sorted(LOCALES_DIR.glob("*.json"))


def test_locales_dir_is_not_empty():
    assert CATALOG_PATHS, f"no shipped catalogs found in {LOCALES_DIR}"


@pytest.mark.parametrize("path", CATALOG_PATHS, ids=lambda p: p.name)
def test_shipped_catalog_keys_are_alphabetized(path):
    """Top-level keys must be alphabetized; nested plural dicts are exempt
    (CLDR categories have a canonical, non-alphabetical order)."""
    def _no_dupes(pairs):
        seen = set()
        for key, _ in pairs:
            assert key not in seen, f"{path.name}: duplicate key {key!r}"
            seen.add(key)
        return dict(pairs)

    data = json.loads(path.read_text(encoding="utf-8"), object_pairs_hook=_no_dupes)
    keys = list(data)
    assert keys == sorted(keys), f"{path.name}: top-level keys are not alphabetized"
```

Parametrizing also means a failure names the offending catalog in the test ID instead of aborting the whole loop at the first bad file.

### Two smaller thoughts

- **`docs/I18N.md` §5 (Translator quickstart)** walks through adding keys but never mentions that catalogs must stay sorted. Contributors will now hit a CI failure with no documented rule behind it. One line in that list would close the loop.
- **Enforcement is test-only.** `lefthook.yml` has pre-commit hooks for `*.py` only, so the fix loop is "push → CI fails → re-sort by hand." A tiny `*.json` locale-sorting hook (or a `scripts/` helper alongside `profile_startup.py`) would make the rule self-healing. Pure YAGNI call at 3 catalogs — flagging it for when the locale count grows, not for this PR.

### Assessment

**Risk: LOW.** Data-only reordering with zero semantic delta, proven by structural comparison; the one code change is an additive test. No runtime behavior is affected — `catalog.py` loads these into dicts and looks keys up by name, so file order was never load-bearing. Rollback is trivial.

**Recommendation: approve and merge.** None of the above blocks it — the suggestions harden a *new* test, they don't fix a regression this PR introduces. Worth doing as a quick follow-up commit, with #2 (duplicate keys) the one I'd genuinely prioritize, since it's the failure mode this test's existence implies is covered but isn't.

*Reviewed by Luna  — verified by structural diff + mutation testing, not vibes.*

thomwebb:
Addressed Andrew Tilson's review feedback in commit `9c5c8103`.

- Added shared `Path`-based locale discovery and an explicit non-empty catalog check.
- Parametrized the ordering test by locale.
- Added duplicate-key detection via `object_pairs_hook`.
- Documented that top-level keys are alphabetical while nested CLDR plural categories retain canonical order.
- Merged the latest `upstream/main` and preserved all newer locale keys.

Validation: `pytest tests/i18n -q -p no:randomly` → 199 passed; all GitHub CI checks are green.

AndrewTilson (approved):
## Re-review: `9c5c810` (follow-up to my review of `a5b8783`)

All four suggestions adopted, plus the docs note. I re-verified the parts that actually changed rather than re-litigating what was already proven clean.

### The new risk: the upstream merge

`9c5c810` merges `upstream/main` into a branch whose entire purpose is reordering three JSON files. That is textbook territory for conflict resolution to silently drop a key, so it got the most scrutiny.

Compared the parsed structures against the **new** merge base (`2a72a7a`):

| Catalog | Leaves | Added | Removed | **Values changed** | Dupes | Bytes base → head |
|---|---|---|---|---|---|---|
| `en-US.json` | 493 | 0 | 0 | **0** | none | 39068 → 39068 |
| `es.json` | 335 | 0 | 0 | **0** | none | 28900 → 28900 |
| `fr-CA.json` | 335 | 0 | 0 | **0** | none | 29613 → 29613 |

And the transitive check that actually matters — **original pre-PR base (`c2a8304`) → current head**, which proves the merge didn't eat anything that existed before this branch started:

- **0 keys lost** in all three catalogs.
- **+29 keys gained** per locale (the upstream `add_model_vllm` additions) — arriving, not colliding.
- **0 altered values** among the survivors.

Still byte-for-byte identical to a canonical `json.dumps(..., sort_keys=True, ensure_ascii=False)` dump, cross-locale parity holds at 158 (`es`/`fr-CA` subsets of `en-US`, zero orphan keys), and nested plural dicts correctly retain CLDR `["one", "other"]` rather than being alphabetized. A clean merge on a sorted file, which is the nice side effect of sorting: upstream insertions land deterministically instead of clustering at the bottom.

### The hardened test actually bites now

I didn't take the fix on faith — I re-ran the exact mutations that defeated the previous version:

- **Duplicate key** (the one that mattered): appended a second `"confirm.yes": "DUPLICATE HIJACK"`. Previously passed both tests while silently overriding the translation. Now → `AssertionError: en-US.json: duplicate key 'confirm.yes'`. Caught.
- **Unsorted top-level keys**, swapping first/last in `es.json` only → fails as `test_shipped_catalog_keys_are_alphabetized[es.json]`. The parametrization pays off exactly as hoped: the failure names the offending catalog instead of aborting the loop at the first bad file.
- **Vacuous pass** closed by `test_shipped_catalog_paths_are_not_empty` plus the in-test assert.
- **DRY nit** resolved — `_LOCALES_DIR` / `_CATALOG_PATHS` module constants, function-local `import glob, os` gone from both tests.

The docstring cleanly encodes the CLDR exemption ("nested CLDR plural categories are exempt"), so the non-alphabetical plural order is now a documented decision rather than an accident waiting for someone to "fix" it. `docs/I18N.md` §5 gained the matching translator-facing rule, so contributors learn it from the guide instead of from a CI failure.

### Validation

Fresh venv off `9c5c810`: `pytest tests/i18n -q -p no:randomly` → **199 passed**, matching your stated number. `ruff format --diff` clean on the test file. All three CI checks green, `mergeStateStatus: CLEAN`.

One note, explicitly **not** attributable to this PR: `ruff check` reports one `DTZ001` (naive `datetime` at line 228, `test_format_datetime_iso`). I confirmed it's present identically on the merge base `2a72a7a`, so it's pre-existing and evidently not part of the `quality` gate's ruleset. Mentioning it only so it isn't mistaken for fallout from these commits — leave it alone here.

### Assessment

**Risk: LOW**, unchanged. Data-only reordering with a proven-zero semantic delta across two different bases, a merge that demonstrably preserved every pre-existing translation, and a genuinely stronger regression test verified by mutation.

The one remaining item from my original review is the **pre-commit enforcement** idea, which I explicitly filed as YAGNI-at-three-catalogs. That still stands: test-only enforcement is the right call for now, and a `*.json` sorting hook can wait until the locale count justifies it. Nothing outstanding.

**Recommendation: approve and merge.** Clean follow-through — the duplicate-key fix in particular closed a real silent-translation-hijack hole.

*Re-reviewed by Luna — structural diff across both bases + mutation testing, not vibes.*

PR #861, opened 2026-08-24
feat(i18n): extract tools common approval messages

thomwebb:
Continues the CLI i18n extraction effort by localizing user-facing approval messages in `code_puppy/tools/common.py`.

- Adds `tools.common.approval.*` catalog keys.
- Adds Spanish and Canadian French translations.
- Preserves dynamic placeholders and keeps blank lines/markup out of catalogs.
- Adds focused catalog, pseudolocale, and placeholder tests.

PR #945, opened 2026-09-15
test(i18n): preserve config command contracts

thomwebb:
## Summary
- Keep two focused i18n regression tests for the already-landed `config_commands.py` localization:
  - concrete `cfg.*` interpolation coverage
  - import smoke coverage for the config command module
- Keep the escaped-brace interpolation contract documentation/test clear.

The production extraction and catalog changes are already present on `main`; this PR intentionally contains only the two unique test improvements identified during review. It does not alter runtime code or locale catalogs.

## Validation
- `pytest tests/i18n tests/command_line/test_config_commands_full_coverage.py::TestHandleShowCommand::test_show_command tests/test_command_handler.py::test_show_status -q -p no:randomly`: 203 passed
- Ruff check and formatting passed
- GitHub CI checks are green


AndrewTilson (changes requested):
## Review: `feat(i18n): extract config command messages`

Thanks for pushing the i18n extraction forward — the `cfg.*` catalog work and the new test module are genuinely nice. Unfortunately I have to **request changes**, because the merge in `8e27855` appears to have resurrected ~190 lines of code that `main` deliberately deleted, and two of the resurrected pieces are hard crashes.

I verified everything below against `pr945` (merge-base `52c24c6`) in a clean venv on Python 3.12.

---

### Blocker 1 — `/diff` crashes with `ModuleNotFoundError`

`handle_diff_command` imports a module that does not exist anywhere in the tree:

```python
from code_puppy.command_line.diff_menu import interactive_diff_picker
```

```
$ ls code_puppy/command_line/diff_menu.py
ls: No such file or directory
```

Actually invoking it:

```
/diff -> CRASH: ModuleNotFoundError: No module named 'code_puppy.command_line.diff_menu'
```

Because the import is function-local, this doesn't fail at import time — it fails the moment a user types `/diff`. And since `@register_command` puts it in the `config` category, it *is* advertised in `/help` and in tab-completion. We're shipping a discoverable command that is guaranteed to explode.

### Blocker 2 — `/colors` crashes with `ImportError`

Same shape, different failure:

```python
from code_puppy.command_line.colors_menu import interactive_colors_picker
```

`colors_menu.py` exists, but it has **zero** functions or classes — it's 102 lines of pure dict constants. Its module docstring is, I promise I am not making this up, a tombstone for exactly this code:

> *"The interactive `/colors` TUI that used to live here (and its `/diff` sibling) was removed: banner and diff styling are owned by the theme system now (the `theme` plugin's `/themes` picker drives palettes, banner colors, and diff tints together). This module survives as a pure data module because the theme plugin imports these constants from this path -- keep the module name stable."*

Runtime confirmation:

```
/colors -> CRASH: ImportError: cannot import name 'interactive_colors_picker'
           from 'code_puppy.command_line.colors_menu'
```

### Blocker 3 — the merge re-added deleted code, it wasn't authored here

This is the root cause of 1 and 2, and it's worth calling out explicitly so the fix is a revert rather than a patch.

```
$ git log --oneline -S 'name="diff"' pr945 -- code_puppy/command_line/config_commands.py
36fb522 chore: bump version [ci skip]        # ancient — predates the removal

$ git diff 8e27855^2 8e27855 -- code_puppy/command_line/config_commands.py | grep -c '^+'
191
```

`8e27855^2` is `52c24c6` (current `main`). The merge commit re-introduces 191 lines relative to the main side, including `handle_diff_command`, `handle_colors_command`, and `_show_color_options`. None of these exist on `main`:

```
$ git grep -n 'name="diff"\|name="colors"\|_show_color_options' main -- '*.py'
main:tests/command_line/test_config_commands_extended.py:44: def _show_color_options(diff_type):   # a local test stub, unrelated
```

So this is a conflict resolution that picked "ours" on a block where `main` correctly said "delete." A PR titled *extract config command messages* should not be reviving two removed features as a side effect. The whole `/diff` + `/colors` + `_show_color_options` block wants to go.

---

### Major — `_show_color_options` is dead code, and the merge mangled it

Zero callers in `code_puppy/`:

```
$ grep -rn "_show_color_options" --include=*.py code_puppy/
code_puppy/command_line/config_commands.py:552:def _show_color_options(color_type: str):
```

It's ~105 lines of hardcoded Rich color tables that duplicate data the theme system already owns — a DRY violation on top of being unreachable. It also carries a visible merge scar:

```python
# ============================================================================
# UTILITY FUNCTIONS
# ============================================================================

def _show_color_options(color_type: str):
    # ============================================================================
    # UTILITY FUNCTIONS
    # ============================================================================

    """Show available Rich color options organized by category."""
```

The banner got duplicated *and* one copy landed inside the function body, above the docstring. The docstring still technically registers (comments aren't statements, so `ast.get_docstring` still returns it — I checked), but this is unambiguously a botched conflict resolution rather than intent.

### Major — 12 of the 19 new catalog keys are orphans

The `cli.help.*` block added to `en-US.json` is referenced by nothing:

```
ORPHAN cli.help.autosave_load      ORPHAN cli.help.exit
ORPHAN cli.help.cancel_key         ORPHAN cli.help.macos_paste
ORPHAN cli.help.clear              ORPHAN cli.help.paste_images
ORPHAN cli.help.commands           ORPHAN cli.help.shell_passthrough
ORPHAN cli.help.completion         ORPHAN cli.help.tutorial
ORPHAN cli.help.diff               ORPHAN cli.help.editor_shortcuts
```

(`cli.help.press_tab` is the only `cli.help.*` key `main` uses, from `cli_runner.py:795`.) These were presumably extracted against an older `cli_runner.py` help block that has since changed. They're pure translator burden — every string here is real work for the `es` and `fr-CA` translators, for text no user will ever see. YAGNI says drop them.

And the kicker: once blockers 1–3 are fixed, the remaining 7 `cfg.colors.*` / `cfg.diff.*` keys become orphans too, since their only call sites are the code being removed. **All 19 new keys are effectively unreachable.**

### Minor — `cfg.colors.usage` documents an interface that doesn't exist

```json
"cfg.colors.usage": "Usage: /diff {{color_type}} <color_name>"
```

Three things wrong in one line: it's a `cfg.colors.*` key describing `/diff`; `/diff` as written takes no arguments at all (it's a zero-arg interactive picker); and the `{{color_type}}` escaping exists purely to make a stale CLI signature render. `test_colors_usage_renders_as_literal_placeholder` is carefully locking in the correctness of a sentence that is factually wrong. Delete the key and the test goes with it.

### Minor — locale parity

All 19 new keys are missing from `es.json` and `fr-CA.json` (both sit at 333 keys vs `en-US` at 510). That gap is pre-existing so I won't hold the PR on it, but note the repo *does* enforce parity elsewhere — `test_agent_manager_i18n.py` and `test_common_i18n.py` both assert across `("es", "fr-CA")`. If `cfg.*` ever gets the same treatment, these keys become a test failure. Another argument for not adding keys you don't need.

### Minor — file size

`config_commands.py` is now **658 lines**, up from ~478. Removing the resurrected block puts it back comfortably under the 600-line guideline without any artificial splitting. Nice when the right fix is also the smaller one.

### Nit — unrelated change in `fe11f08`

The `get_effective_model_settings` swap in `handle_show_command` is a sensible fix, but it's a behavior change riding along in an i18n-extraction PR with no test. Worth its own PR (and a test asserting the `"(model default)"` fallback renders).

---

### What's good

- The `cfg.*` catalog extraction itself is clean and consistent.
- `test_no_leftover_placeholder_for_supplied_params` is a genuinely smart invariant — a catalog-wide guard beats per-key assertions, and the `{{` skip is correctly reasoned in the comment.
- Docstrings explain the *why* (the reload/restart-notice contract comment is good context for a non-obvious behavior).
- The test module docstring is honest about what it does *not* cover. Respect.

### Verification I ran

| Check | Result |
|---|---|
| `pytest tests/i18n -q -p no:randomly` |  206 passed (matches your claim) |
| `pytest tests/command_line -q` |  1036 passed, 1 skipped |
| `ruff format --check` on changed files |  clean |
| `ruff check` on changed files | 12 findings vs 10 on `main` — pre-existing `BLE001` style, not a blocker (repo baseline is 4011 repo-wide) |
| `handle_diff_command("/diff")` |  `ModuleNotFoundError` |
| `handle_colors_command("/colors")` |  `ImportError` |

Note the green test suite is exactly the problem: nothing exercises the two new command handlers, so 206 passing tests sail right past two guaranteed crashes. The PR description's "206 passed" is accurate and also not evidence that this works.

### Suggested path forward

1. Delete `handle_diff_command`, `handle_colors_command`, and `_show_color_options` — redo the merge taking `main`'s side on that block.
2. Drop all 12 `cli.help.*` keys and the now-orphaned 7 `cfg.colors.*` / `cfg.diff.*` keys.
3. Drop `test_colors_usage_renders_as_literal_placeholder` along with its key.
4. Move the `fe11f08` model-settings fix to its own PR, with a test.
5. Re-run `ruff format` and confirm `config_commands.py` lands back under 600 lines.

What's left is the actual `cfg.*` extraction plus two solid new tests — a clean, mergeable change.

---
*Reviewed by Luna — Code Puppy*

thomwebb:
Addressed Andrew Tilson's requested changes in commit `d2ca5597`.

- Restored current `main`'s `config_commands.py` implementation.
- Removed the resurrected `/diff`, `/colors`, and `_show_color_options` code.
- Removed orphaned `cfg.colors.*`, `cfg.diff.*`, and stale `cli.help.*` catalog keys.
- Removed the obsolete colors-specific test while retaining the generic escaped-brace contract test.
- Dropped the unrelated model-settings compatibility change from this i18n PR.

Validation: local targeted/i18n tests passed (**207 passed**), Ruff passed, and all GitHub CI checks are green.

AndrewTilson (commented):
## Re-review: `d2ca5597` — all blockers cleared, but the PR now has no production change left

Thanks for the fast turnaround on `d2ca5597`. Every blocker from my previous review is genuinely fixed, and I re-verified each one rather than taking the commit message's word for it. Full results below.

### Previously-blocking findings: all resolved

| # | Finding | Status | Evidence |
|---|---|---|---|
| 1 | `/diff` -> `ModuleNotFoundError` | **Fixed** | handler deleted; no `diff_menu` reference remains |
| 2 | `/colors` -> `ImportError` | **Fixed** | handler deleted; no `interactive_colors_picker` reference remains |
| 3 | Merge resurrected 191 deleted lines | **Fixed** | `config_commands.py` is now byte-identical to `main` |
| 4 | `_show_color_options` dead code + mangled banner | **Fixed** | gone entirely |
| 5 | 19 orphan catalog keys | **Fixed** | `en-US.json` byte-identical to `main`; `cli.help.*` back to just `press_tab` |
| 6 | `cfg.colors.usage` documenting a non-existent CLI | **Fixed** | key and its test both removed |
| 7 | File over 600 lines | **Fixed** | 658 -> **477** lines |
| 8 | `fe11f08` model-settings fix unrelated/untested | **Moot** | `main` landed the identical change independently |

Confirmation that the two production files are now no-ops:

```
$ git diff main pr945 -- code_puppy/command_line/config_commands.py
$ git diff main pr945 -- code_puppy/i18n/locales/en-US.json
(both empty)
```

Re-ran the suite green: `tests/i18n` 205 passed, `tests/i18n + tests/command_line` **1241 passed, 1 skipped**. `ruff check` and `ruff format --check` on both changed files: clean, zero findings (an improvement — the old file carried 12).

---

### The new problem: what's left is a test-only PR duplicating tests that already exist

With the production changes reverted to match `main`, the PR's entire remaining contribution is `tests/i18n/test_config_commands_i18n.py` plus a two-line docstring edit. And the extraction this file was written to protect **already landed on `main` independently**:

```
$ git log --oneline -S 'cfg.set.success' main -- code_puppy/command_line/config_commands.py
ca3708e chore: bump core plugins to 0.0.34 (#891)

$ git merge-base --is-ancestor e91724b main
NO - main got it independently
```

`main` already has 21 `cfg.*` keys and 29 `t("cfg.*")` call sites. The feature shipped via another PR.

Worse, `main` also already has `tests/i18n/test_catalog_namespaces.py` (landed in #861), which sweeps the `cfg.` namespace with a floor of 20 — and **four of the six tests here are verbatim copies of it**, differing only in the key-filter expression. Side by side:

```python
# PR: test_every_cfg_key_resolves_to_real_text
offenders = [k for k in _cfg_keys()      if not translate.t(k) or translate.t(k) == k]
# main: test_every_key_resolves_to_real_text
offenders = [k for k in _keys(prefixes)  if not translate.t(k) or translate.t(k) == k]
```

`test_no_leftover_placeholder_for_supplied_params` is a line-for-line duplicate, **including the `{{`-skip comment**. That's the test I praised last round as "a genuinely smart invariant" — I owe you a correction: it's smart, and it's already in the codebase. I credited it as new because I reviewed the diff without checking whether `main` had beaten you to it. My miss.

This is the round-1 pattern repeating, and `test_catalog_namespaces.py`'s own docstring calls it out by name:

> *"These checks used to be duplicated verbatim in each per-module i18n file; they are identical apart from the key-prefix filter and the 'namespace is populated' floor. The per-module files keep their concrete key/interpolation contracts, **this module owns the generic sweep**."*

Someone already did this consolidation and left a note asking not to undo it. Deleting the whole file loses nothing — `pytest -k cfg` still passes 5 tests from the consolidated sweep with it removed.

**Only two tests here are actually unique**, and both are worth keeping:

- `test_parametrized_cfg_keys_interpolate` — concrete per-key interpolation assertions. This is exactly the "per-module files keep their concrete contracts" role the docstring carves out.
- `test_config_commands_imports_cleanly` — the only test that would have caught blockers 1 and 2. Genuinely valuable, ironically.

**Suggestion:** cut the file down to those two tests and drop the four duplicates. ~80 lines becomes ~25, and it slots into the existing convention instead of fighting it.

### Minor — the docstring edit left a broken sentence

`test_i18n_contracts.py` now reads:

```
Tests the grammar directly: the last catalog key using this escape
The contract must hold for future translators regardless of which
catalog key uses the escape.
```

Line 1 is a sentence fragment that dead-ends — the old clause `(cfg.colors.usage) left with the /colors TUI, but` was removed and the remnant never rejoined. Suggested fix:

```
Tests the grammar directly rather than through a catalog key. The
contract must hold for future translators regardless of which key
uses the escape.
```

Worth noting the substance of this edit is correct: **zero** keys in `en-US.json` currently contain `{{`, so testing `_interpolate` directly is the right call. (Side effect: the `if "{{" in text: continue` branch in both sweep files is currently unreachable — harmless future-proofing, just don't expect it to be exercised.)

### Nit — PR metadata is now stale

Title says *"extract config command messages"* and the description lists *"Localize user-facing messages in `config_commands.py`"* and *"Add the current config/diff/color catalog keys to `en-US.json`."* Both files are now identical to `main`. If this lands as a test-only change, retitle to something like `test(i18n): add cfg.* interpolation and import-smoke coverage` so the history doesn't claim an extraction that happened elsewhere.

---

### Verification summary

| Check | Result |
|---|---|
| `pytest tests/i18n` | 205 passed |
| `pytest tests/i18n tests/command_line` | 1241 passed, 1 skipped |
| `ruff check` (changed files) | All checks passed |
| `ruff format --check` (changed files) | 2 files already formatted |
| `config_commands.py` vs `main` | identical |
| `en-US.json` vs `main` | identical |
| `config_commands.py` length | 477 lines (was 658) |
| `pytest -k cfg` with PR file deleted | 5 passed (coverage retained by existing sweep) |

Note: `models_dev_parser.py`, `pyproject.toml`, `uv.lock`, and `test_models_dev_registry_cache.py` also show in a `main...PR` diff — that's just `main` moving 2 commits ahead (`4b6f74b`, `556229d`), not your work. A rebase will clear it.

### Recommendation

The dangerous parts are gone and the cleanup was done properly — thank you for that. I'm not re-blocking on correctness; nothing here can crash. Remaining asks are both small:

1. Trim the test file to `test_parametrized_cfg_keys_interpolate` + `test_config_commands_imports_cleanly`; drop the four duplicated sweep tests.
2. Fix the fragment sentence in the `test_i18n_contracts.py` docstring.
3. Retitle/update the description to reflect a test-only change.

Do those and this is a clean, focused little PR. Happy to re-review quickly.

---
*Reviewed by Luna - Code Puppy*

thomwebb:
Updated the PR metadata to reflect the final scope.

The production i18n extraction is already present on `main`; this PR now intentionally keeps only the two unique test improvements from review:

- concrete `cfg.*` interpolation coverage
- `config_commands` import smoke coverage

Also fixed the escaped-brace contract docstring. No runtime or catalog changes remain in this PR. Local validation is green with 203 passed, and GitHub CI is green.

AndrewTilson (approved):
## Re-review: `4b652d63` — LGTM, approving

All three round-2 asks are done, and I verified each one rather than trusting the commit message. The PR is now 23 added lines across two test files. Nice shrink from the original 294.

### Round-2 asks: all resolved

| Ask | Status | Evidence |
|---|---|---|
| Trim the four duplicated sweep tests | **Done** | file is 20 lines, exactly the two unique tests |
| Fix the fragment sentence in `test_i18n_contracts.py` | **Done** | reads cleanly now |
| Retitle / update description | **Done** | `test(i18n): preserve config command contracts`, body accurately says production changes already landed on `main` |

The true PR diff against merge-base is now exactly:

```
tests/i18n/test_config_commands_i18n.py | 20 ++++++++++++++++++++
tests/i18n/test_i18n_contracts.py       |  7 +++----
```

Also good: dropping the local `_reset_locale` fixture was the right call, not an isolation regression. `tests/i18n/conftest.py` already provides an autouse `_i18n_isolate` that resets the catalog, forces `en-US`, **and** clears the five locale env vars — strictly stronger than what was removed. Its docstring explicitly says it "replaces all nine per-file copies," so the trim moved *toward* the convention this repo already set. Exactly the right instinct.

### I verified the tests actually bite

Rubber-stamping "tests pass" is worthless if the tests can't fail, so I mutated the code under them:

**Mutation 1** — stripped the `{key}`/`{value}` placeholders out of `cfg.set.success`:
```
FAILED test_parametrized_cfg_keys_interpolate
```

**Mutation 2** — injected a module-level bad import:
```
FAILED test_config_commands_imports_cleanly  (ModuleNotFoundError)
```

Both genuinely catch regressions. Working tree restored and re-verified clean afterward.

### Correcting myself: the import test would NOT have caught the round-1 blockers

In my last review I called `test_config_commands_imports_cleanly` "the only test that would have caught blockers 1 and 2." **That was wrong**, and I want it on the record rather than quietly dropped.

I reconstructed the exact round-1 shape — a registered command whose bad import sits *inside* the function body:

```python
@register_command(name="diff", description="x", usage="/diff", category="config")
def handle_diff_command(command: str) -> bool:
    from code_puppy.command_line.diff_menu import interactive_diff_picker
    return True
```

Result: **2 passed.** Sails right through. Function-local imports don't execute at import time, which is precisely why those two crashes hid behind 206 green tests in the first place. I asserted that claim from reasoning instead of testing it; the mutation run is what caught me.

This isn't a blocker and I'm not asking for a change — the test still has real value (it catches module-level breakage, as mutation 2 shows) and closing the gap properly means invoking handlers, which needs the full CLI stack. Your module docstring was always honest that registration side-effects aren't exercised here. Just don't let anyone cite this test as crash-proofing for slash commands; it isn't. Worth a future ticket for a smoke test that actually calls each registered handler.

### Verification summary

| Check | Result |
|---|---|
| `pytest tests/i18n` | 201 passed |
| `pytest tests/i18n tests/command_line tests/test_command_handler.py` | 1326 passed, 1 skipped |
| Merge `main` + PR, then `pytest tests/i18n` | clean merge, 201 passed |
| `ruff check` (changed files) | All checks passed |
| `ruff format --check` | 2 files already formatted |
| Production code vs `main` | unchanged (no runtime or catalog delta) |
| All 4 asserted `cfg.*` keys present on current `main` | yes |
| `handle_unpin_command` present on current `main` | yes |
| GitHub CI | 3/3 SUCCESS, `mergeable: MERGEABLE`, `CLEAN` |

Note `main` has moved 39 commits ahead since the merge-base, so a GitHub file-diff will show unrelated churn (`gemini_*`, `model_factory`, `callbacks`, etc.). I tested an actual merge of current `main` with this branch — no conflicts, everything green. Nothing to do.

### Verdict

Approving. Two small, focused, non-redundant tests that fail when they should, slotted into the existing i18n convention instead of duplicating it. Thanks for working through three rounds of this — the end state is genuinely tidy.

---
*Reviewed by Luna - Code Puppy*

PR #946, opened 2026-09-15
Retry transient raw TLS stream failures

thomwebb:
## Summary

- classify raw `ssl.SSLError` failures with reason `DECRYPTION_FAILED_OR_BAD_RECORD_MAC` as transient
- retry the interrupted provider stream from the last completed step
- keep certificate-verification and TLS protocol/configuration failures non-retryable
- cover direct, `ExceptionGroup`, and retry-loop behavior

## Root cause

AnyIO can surface a raw TLS exception while reading an established SSE response instead of wrapping it as an `httpx` or `httpcore` network error. The streaming classifier did not recognize that raw exception, so Code Puppy skipped automatic recovery and rendered the full traceback.

The classifier uses a narrow TLS-reason allowlist rather than treating every `ssl.SSLError` as transient. Retrying certificate or protocol failures would hide actionable configuration problems.

## Validation

- `pytest -q -o addopts='' tests/agents/test_runtime_streaming_retry_classifier.py tests/agents/test_agent_exception_recovery.py tests/agents/test_retry_checkpoint.py tests/agents/test_runtime_pause_integration.py tests/agents/test_runtime_pause_leakage.py` — 32 passed
- `ruff check code_puppy/agents/_runtime.py tests/agents/test_runtime_streaming_retry_classifier.py`
- `ruff format --check code_puppy/agents/_runtime.py tests/agents/test_runtime_streaming_retry_classifier.py`
- `git diff --check`

## Reported failure

```text
ssl.SSLError: [SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC]
decryption failed or bad record mac (_ssl.c:2648)
```

## Internal tracking

- [PUP-790](https://jira.walmart.com/browse/PUP-790)


AndrewTilson (commented):
## Review: Retry transient raw TLS stream failures

Nice, tightly-scoped fix. The diagnosis is convincing, the allowlist-over-blanket-`ssl.SSLError` instinct is exactly right, and I like that certificate/protocol failures got explicit negative tests instead of being left implicit. CI is green on all three jobs and `ruff format --check` is clean.

I checked this out locally (py3.12) and ran the suite — 10/10 pass. But one finding below is worth fixing before merge.

---

### The `.reason` branch is never exercised by the tests

`ssl.SSLError.reason` is populated by CPython's `_ssl` C layer when OpenSSL raises the error. It is **not** set by the Python constructor, so a hand-built exception always has `reason is None`:

```python
>>> e = ssl.SSLError(ssl.SSL_ERROR_SSL, "[SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC] ...")
>>> e.reason
None          # same on 3.9 and 3.12
```

`_tls_record_error()` builds exactly that, so every new test falls through to the message-parsing fallback. The primary branch is untested.

Mutation test to confirm — I deleted the branch outright:

```python
-    reason = getattr(exc, "reason", None)
-    if isinstance(reason, str) and reason.upper() in _RETRYABLE_TLS_REASONS:
-        return True
+    # MUTATED: .reason branch disabled
```

```
10 passed in 7.19s
```

All green with the load-bearing code removed. That branch is the one that fires in production (real errors *do* carry `.reason` — I confirmed against `expired.badssl.com`, which yields `reason='CERTIFICATE_VERIFY_FAILED'`), so right now the path that matters in prod has zero regression protection while the backup path has all of it.

A one-liner closes it:

```python
def test_is_retryable_one_uses_structured_reason_attribute() -> None:
    """Real OpenSSL-raised errors populate .reason; constructed ones don't."""
    exc = ssl.SSLError(ssl.SSL_ERROR_SSL, "decryption failed or bad record mac")
    exc.reason = "DECRYPTION_FAILED_OR_BAD_RECORD_MAC"

    assert _is_retryable_one(exc) is True
```

Worth a brief comment in `_tls_record_error()` too, noting it deliberately models the `.reason`-less shape — otherwise the next reader assumes it covers both paths.

---

### The fallback can false-positive on echoed text

```python
message = str(exc).upper()
return any(f"[SSL: {reason}]" in message for reason in _RETRYABLE_TLS_REASONS)
```

Upcasing the whole message means any casing matches, including a string merely *quoting* the reason rather than being it:

```python
>>> _is_retryable_one(ssl.SSLError(1, "config error mentioning [ssl: decryption_failed_or_bad_record_mac]"))
True
```

Contrived, and the blast radius is a couple of wasted retries — but the PR description's own premise is *"keep this allowlist narrow,"* and matching on upcased free text is the loosest tool available. Since the fallback exists only for exceptions missing `.reason`, gating it on that would keep it from second-guessing the structured path:

```python
if reason is None:
    message = str(exc).upper()
    return any(f"[SSL: {r}]" in message for r in _RETRYABLE_TLS_REASONS)
return False
```

### Shadowed loop variable

The comprehension rebinds `reason`, which is already a local from the lookup six lines up:

```python
reason = getattr(exc, "reason", None)
...
return any(f"[SSL: {reason}]" in message for reason in _RETRYABLE_TLS_REASONS)
```

Generator scoping means this is harmless at runtime, but it reads like a bug on first pass. `for r in ...` (or `for tls_reason in ...`) costs nothing. Readability counts.

---

### Questions on allowlist scope

Both are plausibly deliberate — just want them to be decisions rather than oversights:

1. **`SSLV3_ALERT_BAD_RECORD_MAC`** — the peer-sent alert for the same underlying condition. Currently not retried. If a corrupted record is transient when *we* detect it, is it not equally transient when the server tells us it saw one?
2. **`ssl.SSLEOFError`** mid-stream — an abrupt TLS teardown on an established SSE stream looks a lot like the dropped-socket case `_RETRYABLE_EXCEPTIONS` already covers via `httpx.NetworkError`. Does it reliably arrive pre-wrapped, or can it escape raw the same way this one did?

A one-line comment on `_RETRYABLE_TLS_REASONS` recording *why* the set is a singleton would save the next person this exact archaeology.

---

### What's good

- Placement inside `_is_retryable_one` is the correct seam — it inherits `should_retry_streaming`'s `ExceptionGroup` descent and cause-chain walking for free, no duplicated traversal. Nicely DRY.
- `_is_retryable_tls_stream_error` is a standalone predicate, independently testable, consistent with the existing helper style in this module.
- Negative tests for cert-verify and `WRONG_VERSION_NUMBER` are the right guard — they pin the *narrowness*, which is the part most likely to erode under future pressure.
- Comments explain the *why* (AnyIO leaking a raw exception past the httpx wrapper), not the *what*.
- Single atomic commit, conventional format, no scope creep.

### Assessment

**Risk: LOW–MEDIUM.** Additive change to retry classification on a well-tested seam; worst realistic outcome is a handful of unnecessary retries, not a crash or a hang. Also worth noting the failure mode this fixes is user-hostile (full traceback, lost turn), so shipping it has real upside.

**Recommendation:** approve once the `.reason` regression test lands. The other three items are polish and fine to fold in or defer — your call.

<sub>Reviewed by Luna — verified locally on py3.12: 10/10 tests pass, mutation-tested the classifier branches, confirmed the ruff findings on this file are pre-existing on `main` and not introduced here.</sub>

thomwebb:
Thanks, Andrew — verified the premise on Python 3.13 before changing it: constructed `ssl.SSLError` instances have no `.reason`, and assigning the structured reason works for a direct production-path regression test.

Addressed in `971e1ed8`:

- added direct coverage for `reason='DECRYPTION_FAILED_OR_BAD_RECORD_MAC'`
- restricted message parsing to exceptions where `.reason` is absent
- added a negative test proving a structured non-retryable reason cannot be overridden by matching message text
- renamed the generator variable to avoid shadowing
- documented why the allowlist remains a singleton
- clarified that the original helper intentionally models the reason-less fallback shape

Focused recovery suite is now 32/32 passing; Ruff and format checks are clean.

I’m deliberately deferring `SSLV3_ALERT_BAD_RECORD_MAC` and raw `SSLEOFError` until we observe or reproduce either bypassing the httpx/httpcore wrappers. Their normal wrapped forms are already covered, and broadening the raw allowlist speculatively would work against the narrow policy this PR is trying to preserve.

AndrewTilson (approved):
## Re-review: `971e1ed8` — all findings resolved

Verified the new commit locally on py3.12 against the same harness I used last round. All four items are genuinely closed, and I re-ran the mutation tests rather than taking the checklist at face value.

### The coverage gap is actually closed

Last round I deleted the `.reason` branch and all 10 tests still passed. Same mutation against `971e1ed8`:

```python
     reason = getattr(exc, "reason", None)
-    if isinstance(reason, str) and reason.upper() in _RETRYABLE_TLS_REASONS:
-        return True
+    # MUTATED: structured .reason branch disabled
```

```
FAILED ...::test_is_retryable_one_uses_structured_tls_reason
1 failed, 11 passed
```

The production path now fails loudly when broken. That's the difference between a test that exists and a test that holds.

### The new precedence guard is load-bearing too

I didn't want the `reason is not None` early-return landing as untested code and recreating the original problem one layer down, so I mutated it as well:

```python
-    if reason is not None:
-        return False
+    # MUTATED: precedence guard removed
```

```
FAILED ...::test_structured_tls_reason_is_not_overridden_by_message_fallback
1 failed, 11 passed
```

Pinned. `test_structured_tls_reason_is_not_overridden_by_message_fallback` is my favourite test in the PR — a structured `WRONG_VERSION_NUMBER` carrying retryable-looking message text is exactly the adversarial case, and it now returns `False`. Structured data authoritatively beats string matching. Explicit is better than implicit.

Behavioural spot-checks on the clean tree:

| case | result |
|---|---|
| `.reason='DECRYPTION_FAILED_OR_BAD_RECORD_MAC'` | `True` |
| `.reason='WRONG_VERSION_NUMBER'` + matching message | `False` |
| `SSLCertVerificationError` | `False` |

### Verification

- focused suite: **32/32 passing** — matches your report
- `ruff format --check`: clean
- `ruff check`: still 54 on this file, unchanged from `main` — pre-existing, not yours (restating so it doesn't get misread as a regression later)
- CI: all three jobs green on `971e1ed8`

### On the deferred items

Deferring `SSLV3_ALERT_BAD_RECORD_MAC` and raw `SSLEOFError` until one is actually observed escaping the wrappers is the right call, and better than what I asked for. I raised them as questions precisely because I couldn't demonstrate either bypassing httpx/httpcore — speculatively widening a security-adjacent allowlist on my hunch would trade a real narrow fix for a theoretical broad one. "Add another reason only after confirming it can bypass the wrappers" in the code is worth more than either symbol would have been. YAGNI, correctly applied.

### Residual nit (non-blocking, no action needed)

The echoed-text case still classifies as retryable when `.reason` is absent:

```python
>>> _is_retryable_one(ssl.SSLError(1, "config error mentioning [ssl: decryption_failed_or_bad_record_mac]"))
True
```

But the guard shrank this to *only* exceptions with no structured reason, and real OpenSSL-raised errors always populate `.reason` — so reaching this now requires a synthetic or re-wrapped exception whose message happens to quote the marker. Worst case is a couple of wasted retries. Not worth more code; the interesting half got fixed.

---

**Approving.** Clean root-cause fix, allowlist stayed narrow under review pressure, both branches now mutation-proof, and the deferrals are documented in the code where the next person will find them. Thanks for verifying the `.reason` premise independently on 3.13 instead of just taking my word for it — that's the right instinct.

<sub>Reviewed by Luna — re-verified on py3.12: 32/32 pass, both new branches mutation-tested, ruff delta confirmed zero against `main`.</sub>

PR #947, opened 2026-09-15
Defer background agent rebuilds to the main loop

thomwebb:
## Summary
- add a thread-safe deferred agent-reload queue for background integrations
- apply queued `refresh_config()` and `reload_code_generation_agent()` calls on the live CLI event loop
- drain pending reloads after input resolution and before each interactive/headless agent run
- use the existing model-pin flow as a real producer of deferred reload requests
- retain requests for inactive agents; log and retry failed rebuilds up to three times
- expose `refresh_config()` on `BaseAgent` as a documented no-op interface for static agents

This keeps pydantic/MCP agent construction off transient worker `asyncio.run()` loops while providing a concrete, reusable seam for plugins and other background integrations that update agent configuration.

Related internal tracking: [PUP-42](https://jira.walmart.com/browse/PUP-42).

## Test plan
- changed-file Ruff: passed
- focused OSS tests: 19 passed, 98 deselected
- related command/agent suite: 239 passed
- The worktree-local `uv` environment could not bootstrap because the public PyPI certificate for `code-puppy-core-plugins==0.0.50` failed with `UnknownIssuer`; validation used an existing OSS worktree environment instead.


AndrewTilson (changes requested):
## Review: Defer background agent rebuilds to the main loop

Nice, tight, well-documented change — the core idea (keep pydantic/MCP construction off transient worker `asyncio.run()` loops) is correct and the module docstring actually explains *why*, which I appreciate. The retain-on-failure and retain-for-inactive-agent semantics are thoughtful.

That said, I think the drain is wired into the wrong point in the REPL loop, and the feature currently has no producers. Details below.

---

###  Blocking

**1. The drain happens a turn too early — it runs *before* the blocking prompt read, not before the run.**

The PR description says "drain pending reloads ... before each interactive turn," but that's not what the placement does:

```
cli_runner.py:914   apply_pending_agent_reloads()     # drain
cli_runner.py:918   current_agent = get_current_agent()
cli_runner.py:950   task = await wait_for_idle_submission()   # <-- blocks for minutes
cli_runner.py:1163  await run_prompt_with_attachments(current_agent, ...)   # <-- runs
```

The loop drains, then **parks on the idle prompt indefinitely**, then runs the agent. The idle wait is precisely the window where a background worker is most likely to finish and call `request_agent_reload()` — and any request arriving in that window is *not* applied to the turn the user is about to submit. It gets applied at the top of the *next* iteration, i.e. one full prompt too late. The user types a message, it runs against the stale agent, and the rebuild lands afterward.

Fix is small: move the call to after `task` is resolved and before `run_prompt_with_attachments`. `current_agent` is captured at line 918, but that's fine — `reload_code_generation_agent()` mutates `agent._code_generation_agent` in place, so the existing local reference stays valid.

While you're in there: the same reasoning applies to the `initial_command` branch at line 821, which *is* correctly placed (drain immediately precedes the run). Worth making the two consistent so the intent is obvious.

**2. Zero producers — nothing in the repo ever calls `request_agent_reload()`.**

```
$ grep -rn "request_agent_reload" --include=*.py .
(only the definition and its tests)
```

This is a consumer, a queue, and a test suite with no producer. The description frames it as "a small generic seam for plugins and other background integrations," which is the textbook YAGNI smell — we're merging speculative infrastructure and paying the maintenance cost now for a caller that doesn't exist yet. *"Special cases aren't special enough to break the rules"* cuts both ways: if there's a concrete in-flight caller (the PUP-42 work?), land them together so the seam is validated by a real user. If there isn't one yet, I'd hold this.

If it must land standalone, please at least land the first real caller in the same PR or reference the follow-up PR that consumes it.

**3. Headless `--prompt` mode never drains.**

Both `apply_pending_agent_reloads()` call sites are inside `interactive_mode`. `execute_single_prompt` (line 1478) — the `-p` path — never drains, so a background integration's config update is silently dropped in headless runs. Either add a drain there or explicitly document that this is interactive-only (the module docstring currently implies it's general).

---

###  Should fix

**4. `except Exception: pass` is fully silent, and the retry is unbounded.**

```python
except Exception:
    # Keep the request queued so a later prompt can retry safely.
    pass
```

Two problems compounding each other:

- Nothing is logged or emitted. A permanently-failing rebuild (bad model config, dead MCP server) is completely invisible — the user gets a silently stale agent forever with no signal that anything is wrong. *"Errors should never pass silently."* At minimum `emit_warning` / a debug log on the failure path.
- There's no cap or backoff, so a poison entry re-attempts a (potentially slow, MCP-touching) rebuild on **every single turn, forever**. Consider a retry counter that evicts + loudly warns after N attempts.

Also note the `try` wraps `get_current_agent()` too, so a failure *there* silently no-ops the entire drain. Worth narrowing the `try` to just the reload calls so the two failure modes are distinguishable.

**5. `hasattr(current, "refresh_config")` is LBYL duck-sniffing where LSP would do.**

`BaseAgent` doesn't define `refresh_config`; only `JSONAgent` does (`json_agent.py:315`). So the callsite compensates for an incomplete base interface by inspecting for the method at runtime. The SOLID-correct fix is a no-op `refresh_config()` on `BaseAgent`:

```python
def refresh_config(self) -> None:
    """Re-read configuration from its source. No-op for static agents."""
```

Then the caller is just `current.refresh_config()` — subtypes are substitutable, the branch disappears, and the behavior is documented on the interface instead of implied at one callsite. *"Explicit is better than implicit."*

**6. Test asserts on `refresh_config` prove nothing (bare `MagicMock` auto-creates every attribute).**

```python
>>> hasattr(MagicMock(), "refresh_config")
True
```

So `test_pending_reload_rebuilds_active_agent_on_main_loop` always takes the `hasattr` → True branch regardless of what real agents look like, and `agent.refresh_config.assert_called_once_with()` passes vacuously. The `False` branch — the one that actually matters, since it's the `BaseAgent` case, i.e. **most agents in the repo** — has zero coverage.

Use `MagicMock(spec=...)` (or `spec_set`) against the real classes, and add a case for an agent *without* `refresh_config` asserting that `reload_code_generation_agent()` is still called. This is the bug class where a green suite hides a live `AttributeError`.

**7. No concurrency test in a module whose entire reason for existing is thread safety.**

`grep -c "threading\|Thread" tests/agents/test_deferred_reload.py` → `0`. Every test drives the queue single-threaded. Please add one that spawns N threads hammering `request_agent_reload()` while the main thread drains, asserting no lost updates and no exceptions — otherwise the lock is untested decoration.

---

###  Nits / polish

**8. Two different import paths to the same `get_current_agent`.** `deferred_reload.py` imports from `code_puppy.agents`, `cli_runner.py:912` imports from `code_puppy.agents.agent_manager`. Same function, two aliases — which also makes the tests' `patch("code_puppy.agents.get_current_agent")` quietly path-dependent. Pick one and stick to it.

**9. Import placement in the loop is genuinely ugly.**

```python
while True:
    from code_puppy.agents.deferred_reload import apply_pending_agent_reloads
    from code_puppy.agents.agent_manager import get_current_agent

    apply_pending_agent_reloads()
    from code_puppy.messaging import emit_info      # <-- stranded below the call
```

Sandwiching an import *after* a statement mid-block is hard to read, and these are re-executed every iteration. Group the imports. (`cli_runner.py` has an `E402` per-file ignore, so a module-level import is also on the table.)

**10. Module-level mutable globals make the tests poke at privates.** `setup_function`/`teardown_function` reach into `deferred_reload._pending_lock` and `._pending_agent_reloads` to reset state. A small `DeferredReloadQueue` class with a module-level default instance would keep the public API identical while making the thing injectable and resettable without underscore-spelunking. Also removes the cross-test-leakage footgun if someone forgets the teardown.

**11. API naming asymmetry:** `request_agent_reload` (singular) vs `apply_pending_agent_reloads` (plural). `request_agent_reload` / `apply_agent_reloads` reads better as a pair.

**12. Not exported from `code_puppy/agents/__init__.py`.** Every other agent-package symbol is re-exported there with an `__all__` entry. If this is meant to be the public plugin seam, it should be reachable as `from code_puppy.agents import request_agent_reload` rather than requiring a deep-module import.

---

###  Verified locally

- `pytest tests/agents/test_deferred_reload.py` → **3 passed**
- Confirmed `BaseAgent` has no `refresh_config` (so the `hasattr` branch is load-bearing, and untested)
- Confirmed `request_agent_reload` has no callers outside its own tests
- Confirmed both drain sites are interactive-only
- FWIW I hit the same `UnknownIssuer` PyPI cert wall mentioned in the test plan; installing against an internal mirror index got a clean env if that's useful to you

### Risk assessment

**Medium.** Blast radius is small and the failure mode is fail-safe (silently does nothing), so it won't break existing flows. But as written the feature doesn't reliably deliver on its own contract — issue #1 means updates land a turn late, and #2 means nothing exercises it yet. The silent-forever-retry in #4 is the one I'd least like to debug in the field.

Happy to re-review once the drain moves and a real caller shows up.

thomwebb:
Addressed Andrew Tilson's review in d67f9829:

- moved the interactive drain to after input resolution and immediately before the agent run;
- added the same drain to headless `-p` execution;
- converted the existing model-pin deferred path into the first real producer;
- added bounded retry/logging (three attempts) instead of silent infinite retries;
- added `BaseAgent.refresh_config()` as a documented no-op interface;
- exported the queue API and replaced private test-state manipulation with a queue class;
- added concurrent request coverage and retained inactive-agent requests.

Validation: 239 related tests passed; changed-file Ruff passes. The public PyPI certificate bootstrap limitation remains documented in the PR description.

WSxDemise (commented):
## Review: Defer background agent rebuilds to the main loop

**Summary of the fix**: Replaces the ad-hoc `_PENDING_PIN_RELOADS` list (drained only after the agent picker) with a general-purpose, thread-safe `DeferredReloadQueue` (`code_puppy/agents/deferred_reload.py`) that any background integration can use via `request_agent_reload()` / `apply_agent_reloads()`. The queue is drained at three points on the live event loop in `cli_runner.py` (initial command, top of the main loop, and headless `-p` execution), which correctly avoids scheduling MCP autostart tasks on the picker's transient `asyncio.run` loop — the root cause of the described deadlock.

### Strengths
- `DeferredReloadQueue` is small, well-isolated, and covered by good tests, including a real concurrency test (`test_concurrent_requests_are_not_lost`) exercising 12 threads calling `request()` simultaneously.
- Retry-with-eviction (`MAX_RELOAD_ATTEMPTS = 3`) prevents an unrecoverable agent from spinning forever on every loop iteration.
- `refresh_config()` added as a no-op default on `BaseAgent` (`base_agent.py`) keeps the queue generic without forcing every agent subclass to implement it.
- Removal of the bespoke `agent_menu._PENDING_PIN_RELOADS` / `core_commands.py` drain wiring is a nice simplification — one shared mechanism instead of a single-purpose one.

### Issue worth addressing (non-blocking)
`_reload_agent_if_current()` (removed) used to call `emit_info`/`emit_warning` so the user saw "Active agent reloaded with pinned model 'x'" or a failure warning in the CLI. The new `DeferredReloadQueue.apply()` only calls `logger.exception(...)` on failure and emits nothing on success — a plain Python logger, not code_puppy's `messaging.emit_*` surface the user actually watches. Net effect: after pinning a model, the user gets zero CLI feedback that the reload happened, and if the reload fails 3 times and gets evicted, that failure is silent to the interactive user (only visible in logs). Recommend wiring `emit_success`/`emit_warning` back in at the call sites in `cli_runner.py` (or inside `apply()`, since it already knows success/failure) as a fast follow.

### Minor observation
`request_agent_reload()`'s `setdefault(agent_name, 0)` intentionally preserves an existing retry count (per the docstring). That means if a reload already failed once or twice and the user re-triggers a pin/config change for that same agent, the fresh request inherits the stale attempt count instead of resetting it, so it may get evicted sooner than the "3 attempts" a user would expect for their new request. Worth a quick look, not a blocker.

### Testing
`quality` and `windows-encoding` checks are green; `test (macos-latest, 3.13)` was still pending at review time — worth confirming it finishes green before merge.

### Recommendation
**Approve** — the deadlock fix is correct, well-tested, and a real improvement over the previous single-purpose queue. The missing user-facing success/failure messaging is a legitimate regression but not a correctness blocker; suggest addressing it in this PR or a fast follow-up.

PR #950, opened 2026-09-18
feat(i18n): localize rich renderer messages

thomwebb:
## Summary
- extract user-facing Rich renderer labels, status text, and fallback messages into the i18n catalog
- add reviewed Spanish and Canadian French translations for every new `renderer.*` key
- add focused catalog and interpolation/pseudolocale contract tests

## Validation
- `python3 -m py_compile code_puppy/messaging/rich_renderer.py tests/i18n/test_rich_renderer_i18n.py`
- JSON/catalog key coverage checks pass for `en-US`, `es`, and `fr-CA`
- `git diff --check` passes
- `pytest` could not run in this environment because `uv` and `pytest` are not installed


AndrewTilson (changes requested):
## Review: `feat(i18n): localize rich renderer messages`

Thanks for pushing the i18n epic into the renderer -- this is exactly the right next namespace to extract, and the es / fr-CA translations read well. That said, I reproduced **three user-visible regressions** and a **red CI** locally, so I'm requesting changes.

Everything below was verified by running the repo's own interpolator (`code_puppy/i18n/translate._interpolate`) and `ruff` against the branch head (`8077a7a7`) -- not eyeballed.

---

### Blockers

#### 1. `renderer.exit_duration` silently ships a raw placeholder to users

The catalog entry uses a **format spec**:

```json
"renderer.exit_duration": "[dim]  exit=[/dim][{exit_style}]{exit_code}[/{exit_style}][dim]  {duration:.1f}s[/dim]"
```

But `_interpolate` is deliberately **not** `str.format`. Per the docstring in `translate.py`, the grammar is `\{\{|\}\}|\{(\w+)\}` -- "no dots, indexing, specs, or conversions" -- and anything that isn't a bare `{identifier}` is "left untouched as a literal." So `{duration:.1f}` never matches, and never substitutes.

Reproduced against the branch:

```
template : [dim]  exit=[/dim][{exit_style}]{exit_code}[/{exit_style}][dim]  {duration:.1f}s[/dim]
rendered : [dim]  exit=[/dim][green]0[/green][dim]  {duration:.1f}s[/dim]
```

Every user on `output_level=high` now sees the literal text `{duration:.1f}s` after every shell command. Fix: round at the call site and pass a pre-formatted string.

```python
t("renderer.exit_duration", ..., duration=f"{msg.duration_seconds:.1f}")
```

...with the catalog using a bare `{duration}`. (This also needs doing in the es and fr-CA copies, which inherited the same spec.)

#### 2. Three banner emoji/glyphs were dropped during extraction

The catalog values lost their leading glyphs -- verified by codepoint:

| key | value on branch | should be |
|---|---|---|
| `renderer.fork` | `" FORK"` (5 chars, leading `U+0020`) | `U+1F374` + `" FORK"` |
| `renderer.invoke_agent` | `" INVOKE AGENT"` (13 chars) | `U+1F916` + `" INVOKE AGENT"` |
| `renderer.subagent_response` | `" AGENT RESPONSE"` (15 chars) | `U+2713` + `" AGENT RESPONSE"` |

Compare the removed lines in the diff -- `"\U0001f374 FORK"`, `"\U0001f916 INVOKE AGENT"`, `" AGENT RESPONSE"`. These now render as banners with an orphan leading space. This is a visual regression on the single most-used banner in the sub-agent flow.

Side note: glyphs like these are *not* translatable content and arguably shouldn't live in the catalog at all -- keep the icon at the call site and localize only the word. That also stops a translator from accidentally deleting it again.

#### 3. CI `quality` is failing -- `ruff format --check` rejects both files

The job is already red. I reproduced it with the pinned `ruff>=0.15,<0.16` from `.github/workflows/ci.yml`:

```
2 files would be reformatted
```

Offenders include the `renderer.planned_next_steps` f-string, the `_format_banner("subagent_response", ...)` call, `_format_banner("universal_constructor", ...)`, and the `assert data["renderer.directory_listing"] != ...` block in the new test. Just run `ruff format .` before the next push. (Several of the new call sites are also well past 100 chars -- e.g. the `renderer.found_matches` line -- which the formatter will fix for you.)

---

### Design issues

#### 4. The new test file re-implements the shared sweep -- and is strictly weaker

`tests/i18n/test_catalog_namespaces.py` exists precisely so namespaces don't each hand-roll "is populated / resolves / pseudolocalizes / no leftover placeholder." Its own docstring says so: *"These checks used to be duplicated verbatim in each per-module i18n file."* PUP-480 deduplicated nine copies of this. This PR adds a tenth.

This isn't just a DRY nit -- the hand-rolled version **misses the bug in #1**. Two differences:

- The new file uses `re.compile(r"\{(\w+)")` -- no closing brace -- so it happily extracts `duration` from `{duration:.1f}` and supplies it, unlike the shared `r"\{(\w+)\}"`.
- The assertion is hardcoded to two keys: `assert "{error}" not in rendered` / `assert "{cwd}" not in rendered`, instead of the shared sweep's generic "no `{` survives."

I ran both against this branch:

```
--- EXISTING shared sweep applied to renderer.* ---
  FAIL: renderer.exit_duration -> [dim]  exit=[/dim][X]X[/X][dim]  {duration:.1f}s[/dim]
  total failures: 1

--- PR's NEW test ---
  total failures: 0   <-- bug sails straight through
```

The existing infrastructure catches the regression; the new bespoke test does not. Please delete `tests/i18n/test_rich_renderer_i18n.py` and add one row to `_NAMESPACES`:

```python
"renderer": (("renderer.",), 26),
```

That gets you populated + resolves + pseudolocalizes + placeholder coverage for free, in four parametrized tests you didn't have to write. If there are renderer-specific *concrete* contracts worth asserting, those belong as a topic in `test_i18n_contracts.py`, which is the designated home for them.

(Also: `assert len(keys) >= 20` when there are 26 is a floor with no teeth, and `_keys()` / `load_catalog("en-US")` get re-invoked inside loops.)

#### 5. `match_word` / `file_word` defeat the point of localizing `found_matches`

```python
match_word = "match" if msg.total_matches == 1 else "matches"
file_word = "file" if len(by_file) == 1 else "files"
...
t('renderer.found_matches', matches=..., match_word=match_word, files=..., file_word=file_word)
```

The template is translated but the nouns are hardcoded English injected as parameters. A Spanish user gets `Se encontraron 3 matches en 2 files`. Worse, it bakes in a two-form plural rule -- the repo already ships `plurals.py` and `ngettext()` for exactly this, and `renderer.*` currently uses `ngettext` zero times.

This should be a plural-dict catalog entry driven by `ngettext`, which also lets Slavic/Arabic locales add `few`/`many` later without touching the renderer. Same applies to the two other `match_word` sites at lines ~786 and ~820 that were left as raw English.

#### 6. Rich markup inside catalog values is inconsistent, and the pseudolocale mangles it

Nine of the 26 new keys embed markup, against only two in the entire pre-existing catalog -- so this PR quadruples the pattern. Two concrete problems:

**a) It's applied inconsistently.** Some keys own their markup, others get wrapped at the call site:

```
renderer.render_error        catalog_has_markup=True   AND call site wraps in [dim red]...[/dim red]
renderer.no_matches          catalog_has_markup=False  call site wraps in [dim]
renderer.working_directory   catalog_has_markup=False  call site wraps in [dim]
renderer.timeout             catalog_has_markup=False  call site wraps in [dim]
```

`renderer.render_error` therefore emits `[dim red][dim red]Render error: ...[/dim red][/dim red]`. Rich tolerates the nesting, but pick one convention -- I'd suggest markup stays at the call site and the catalog holds plain text.

**b) The pseudolocale corrupts the tags.** `pseudolocalize()` preserves `{placeholders}` but accents everything else, tags included:

```
en-US : [dim red]Render error: {error}[/dim red]
en-XA : [đîḿ ŕéđ]Ŕéñđéŕ éŕŕöŕ: {error}[/đîḿ ŕéđ]
```

`[đîḿ ŕéđ]` is not a valid style, so the pseudolocale run -- the mechanism whose stated job is catching exactly this class of bug -- produces broken output for these nine keys. Moving markup out of the catalog resolves this too.

#### 7. `renderer.summary` moved the colon outside the markup

```diff
- self._console.print("\n[bold cyan]Summary:[/bold cyan]")
+ self._console.print(f"\n[bold cyan]{t('renderer.summary')}[/bold cyan]:")
```

The colon is now unstyled, which is a (small) visual change. More importantly it's now unlocalizable punctuation -- fr-CA requires a narrow no-break space before `:`, and the rest of the fr-CA catalog in this very PR correctly writes `Répertoire de travail :`. Fold the colon into the catalog value.

---

### Nits

- `renderer.grep` is `"GREP"` in all three locales. Fine to keep for consistency, just noting it buys nothing.
- PR description says `pytest` could not run locally. Understandable, but findings #1 and #4 would both have surfaced from a single `pytest tests/i18n/` run -- worth getting a working env before the next push.

---

### Assessment

**Risk: MEDIUM-HIGH.** Touches the rendering path for every message type in the TUI, CI is red, and there are three confirmed user-visible regressions.

The translation quality is genuinely good and the key naming is clean -- the problems are all mechanical. Rough order of work:

- [ ] Fix `{duration:.1f}` in all three locale files + call site (#1)
- [ ] Restore the three dropped glyphs, ideally at the call site (#2)
- [ ] `ruff format .` to green up CI (#3)
- [ ] Delete the bespoke test file; register `renderer` in `_NAMESPACES` (#4)
- [ ] Convert `found_matches` to `ngettext` with a plural dict (#5)
- [ ] Settle the markup-location convention and strip tags from catalog values (#6)
- [ ] Move the `Summary` colon into the catalog (#7)

Happy to re-review once CI is green.

WSxDemise (commented):
### Summary
- Extracted user-facing Rich renderer strings into the `renderer.*` i18n namespace across `code_puppy/messaging/rich_renderer.py`.
- Added corresponding catalog entries and plural forms to `en-US.json`, `es.json`, and `fr-CA.json`.
- Updated catalog namespace counts in `tests/i18n/test_catalog_namespaces.py`.

### Group 1 — Really Should Fix These
None. Scope is surgical and minimal, pluralization correctly uses `ngettext`, and catalog contracts match across all locales.

### Group 2 — Nits
None.

[RECOMMEND APPROVAL]

WSxDemise (approved):

thomwebb:
Andrew — thanks for the thorough review. I addressed all seven findings in follow-up commit `3f706f36`:

1. **Duration placeholder:** formatted the duration at the call site and changed all three catalogs to use the bare `{duration}` placeholder.
2. **Banner glyphs:** restored the fork, invoke-agent, and sub-agent response glyphs at the call site so translators cannot accidentally remove them.
3. **Ruff:** ran the pinned `ruff>=0.15,<0.16`; both `ruff check .` and `ruff format --check .` pass, and CI is green.
4. **Shared namespace coverage:** removed the bespoke renderer test and added `renderer` to `tests/i18n/test_catalog_namespaces.py` with the shared population, resolution, pseudolocale, and placeholder checks.
5. **Pluralization:** replaced all three raw match/file plural sites with `ngettext`-backed `renderer.match_count` and `renderer.file_count` catalog entries.
6. **Rich markup:** removed Rich markup from all `renderer.*` catalog values and kept styling at the call sites. The catalogs now contain zero renderer markup tags.
7. **Summary punctuation:** moved the colon into the localized `renderer.summary` value, including the fr-CA spacing.

All three catalogs (`en-US`, `es`, `fr-CA`) contain the complete renderer namespace, and the upstream CI checks are passing.

PR #951, opened 2026-09-18
feat(i18n): localize model factory warnings

thomwebb:
## Summary
- extract the 15 user-facing model factory warnings into the `model_factory.*` catalog namespace
- add Spanish and Canadian French translations for every new key
- register the namespace with the shared i18n coverage, pseudolocale, and placeholder checks

## Validation
- pinned Ruff (`ruff>=0.15,<0.16`) `ruff check .` passes
- pinned Ruff `ruff format --check .` passes
- Python compilation passes for changed modules
- all `model_factory.*` keys exist in `en-US`, `es`, and `fr-CA`
- `git diff --check` passes

PR #949, opened 2026-09-18
feat(i18n): extract ask-user renderer messages

thomwebb:
## Summary
- Localize ask-user-question renderer chrome: help, navigation, input, timeout, and render-error messages.
- Keep user-provided question/option text dynamic and escaped.
- Add the English catalog keys without changing runtime behavior.

## Validation
- `pytest tests/tools/test_ask_user_question/test_renderers.py tests/i18n -q -p no:randomly`: 230 passed
- Ruff check and formatting passed.

PR #952, opened 2026-09-18
feat(i18n): localize agent menu messages

thomwebb:
## Summary
- extract the 10 user-facing agent menu messages into the `agent_menu.*` catalog namespace
- add Spanish and Canadian French translations for every new key

## Validation
- all `agent_menu.*` keys exist in `en-US`, `es`, and `fr-CA`
- `git diff --check` passes


AndrewTilson (changes requested):
## Review: `feat(i18n): localize agent menu messages`

Nice, tight, on-pattern continuation of the #949/#950/#951 namespace-extraction series. Catalog hygiene is genuinely good — I verified independently:

- All 10 `agent_menu.*` keys present in `en-US`, `es`, and `fr-CA` (0 missing, 0 extras)
- Placeholder sets match **exactly** across all three locales per key (`{model}`, `{agent}`, `{error}`)
- All 10 keys resolve to real text, pseudolocalize (`⟦…⟧`), and leave no un-substituted `{field}`
- Keys stay alphabetically sorted in every catalog (respects #944)
- Every message rendered cleanly in all 3 locales; Spanish and French read naturally, tú-form imperative in `es` matches the existing `agent_manager.delete.active_agent` register

So the translations themselves are solid. My concerns are about **completeness** and the **regression net**, not correctness of what's here.

---

### Blocking: the extraction is incomplete — 2 of 12 sites were missed

`_apply_pinned_model` has two code paths. The JSON-agent branch got localized; the **built-in Python agent branch did not**:

```python
# code_puppy/command_line/agent_menu.py:236-243
else:
    # Handle built-in Python agent - use config functions
    if model_choice == "(unpin)":
        clear_agent_pinned_model(agent_name)
        emit_success(f"Model pin cleared for '{agent_name}'")   # ← line 238, raw f-string
        pinned_model = None
    else:
        set_agent_pinned_model(agent_name, model_choice)
        emit_success(f"Pinned '{model_choice}' to '{agent_name}'")  # ← line 242, raw f-string
        pinned_model = model_choice
```

The repo's own audit tool agrees:

```
$ python -m code_puppy.i18n.audit code_puppy/command_line/agent_menu.py --json
{
  "extracted": 10,
  "raw": 2,
  "dynamic": 0,
  "coverage": 83.33,
  "raw_sites": [
    {"path": "code_puppy/command_line/agent_menu.py", "line": 238, "call": "emit_success"},
    {"path": "code_puppy/command_line/agent_menu.py", "line": 242, "call": "emit_success"}
  ]
}
```

What makes this particularly worth fixing before merge:

1. **The keys already exist.** `agent_menu.pin_cleared` and `agent_menu.pin_set` were added in this very PR and are used 8 lines above. This is a pure two-line change — no new catalog entries, no new translations, no test-floor bump:

   ```python
   emit_success(t("agent_menu.pin_cleared", agent=agent_name))
   emit_success(t("agent_menu.pin_set", model=model_choice, agent=agent_name))
   ```

2. **This is the *more common* branch.** There are ~29 modules under `code_puppy/agents/` shipping built-in Python agents. A user pinning a model to `code-puppy`, `helios`, `planning-agent`, etc. lands on the untranslated path. A user pinning to a JSON agent gets Spanish. Same menu, same action, two different languages — worse UX than consistently-English, because it reads like a bug.

3. It hard-violates DRY in the ugliest way: identical user-facing text maintained in two representations (catalog entry + inline f-string) in the same function. Change the English wording later and the two silently diverge.

The PR description says "extract the 10 user-facing agent menu messages." There are 12.

---

### The test floor can't catch this class of bug

```python
"agent_menu": (("agent_menu.", ), 10),
```

`test_namespace_is_populated` asserts `len(keys) >= 10`. That's a **catalog-side** assertion: it proves ten keys exist in JSON. It proves nothing about whether call sites actually use them. The two raw f-strings above sail through green CI — and indeed all three checks are passing on this PR right now.

This isn't a flaw introduced by this PR (the same shape exists for `renderer`, `model_factory`, etc.), but it's why this miss slipped by, and it's worth naming. Two suggestions, take either or neither:

**(a) Cheap and targeted** — add a module-level audit assertion so the source-side is actually guarded:

```python
def test_agent_menu_has_no_raw_emit_sites():
    from code_puppy.i18n import audit
    report = audit.audit_tree("code_puppy/command_line/agent_menu.py")
    assert not report.raw, [f"{s.path}:{s.line}" for s in report.raw]
```

That test fails today and passes after the two-line fix — exactly the signal you want.

**(b) Strategic** — `docs/I18N.md` already documents `python -m code_puppy.i18n.audit --fail-under N`, but nothing in `.github/workflows/ci.yml` or `lefthook.yml` invokes it. Every PR in this series is nudging coverage upward with zero ratchet enforcing it. Wiring a `--fail-under` gate into CI would make the whole `#949 → #950 → #951 → #952` campaign self-defending. Out of scope here, but worth a follow-up ticket.

---

### Nit (non-blocking): the floor has zero slack

```
agent_menu.   actual=10  floor=10  slack=0
```

Matches the house convention (`mcp.wizard`, `model_factory`, `tools.common.approval` are all at slack=0), so this is consistent and I won't argue it. Just flagging that an exact-match floor is a tripwire, not a floor — it fires on *removal* but is silent on the "did we actually finish" question, which is the one that matters here.

---

### Things done right, credit where due

- Exceptions passed as `error=exc` rather than `error=str(exc)` — matches the dominant convention (`cli_runner.py:574`, `agent_manager.py:383`) and is safe because `_interpolate` in `translate.py` calls `str()` on every substituted value anyway.
- Used `t()` eagerly rather than `lazy()` — correct call, since all these fire from an interactive menu long after locale detection has settled.
- Left `f"{DIM_ON}Name:{RESET}"` and friends in `_render_agent_details` alone. Those are ANSI-decorated preview-pane internals, not the message-bus surface the audit tool covers — reasonable scoping, not an oversight.
- French uses proper guillemets (`« »`) and narrow-space punctuation; Spanish uses straight quotes consistent with the other 54 `es` entries. Somebody actually thought about typography per-locale instead of copy-pasting.
- Diff is minimal and surgical; no drive-by refactors smuggled in.

---

### Verdict

**Request changes** — but this is a ~2-line fix, not a redesign. Localize lines 238 and 242 with the keys this PR already added, and it's a clean merge. Optionally drop in the audit-based test so the next module in this series can't repeat the miss.

Great series overall — the per-namespace cadence with matching `es`/`fr-CA` coverage is genuinely disciplined work.

<sub>Reviewed by Luna — verified locally against `a76b6a8`: catalog parity, placeholder symmetry, pseudolocale bracketing, alphabetical ordering, and `code_puppy.i18n.audit` output.</sub>

thomwebb:
Thanks for catching the incomplete built-in-agent path. Fixed in `af26d27e`:

- localized the two missed `emit_success` calls in `_apply_pinned_model` using the existing `agent_menu.pin_cleared` and `agent_menu.pin_set` keys
- added `tests/i18n/test_agent_menu_i18n.py`, which audits the module source and fails on any raw emit-site literals

The module now has 12 extracted sites and no raw audit sites. Pinned Ruff format/lint, compilation, and diff checks pass locally; CI is rerunning.

AndrewTilson (approved):
## Re-review: `af26d27e` — blocker resolved

Both requested changes landed, and I re-verified them against `af26d27e` rather than taking the commit message on faith.

### Blocker: resolved

The two built-in-agent-path `emit_success` calls are now localized with the keys this PR already shipped. The repo's own audit tool confirms:

```
$ python -m code_puppy.i18n.audit code_puppy/command_line/agent_menu.py --json
{"extracted": 12, "raw": 0, "dynamic": 0, "coverage": 100.0, "raw_sites": []}
```

12/12, up from 10/12. All 12 `emit_*` call sites in the module now route through `t()`. Catalog parity re-checked and still clean — 10 keys, zero missing in `es`/`fr-CA`, `en-US` still alphabetically sorted. No new keys were needed, exactly as predicted.

### The new test is a real guard, not a rubber stamp

I checked that `tests/i18n/test_agent_menu_i18n.py` actually discriminates, by running it against the pre-fix source:

```
OLD (a76b6a8) raw sites -> ['agent_menu.py:238', 'agent_menu.py:242']   test FAILS
NEW (af26d27) raw sites -> []                                           test PASSES
```

That's the property that matters: it fails on the bug it was written to catch. Source-side coverage now complements the catalog-side floor in `test_catalog_namespaces.py`, which closes the exact gap that let the miss through. Nicely done.

All three CI checks pass on the new head.

---

### Nit (non-blocking): `read_text()` without an encoding

```python
sites = audit.audit_source(source_path.read_text(), str(source_path))
```

`Path.read_text()` with no `encoding=` uses `locale.getpreferredencoding()`. That's UTF-8 on the macOS CI runner and on my box, but **cp1252 on Windows** for every Python this project supports (`requires-python = ">=3.11,<3.15"`; PEP 686's UTF-8 default doesn't arrive until 3.15).

Today that's harmless-but-wrong: `agent_menu.py` contains six non-ASCII bytes, all em-dashes (U+2014), which cp1252 turns into mojibake (`—`) without raising, and `ast.parse` happily accepts it. So the audit still returns the right answer by luck.

The trap is latent, though. If anyone later puts a character whose UTF-8 encoding contains `0x81`/`0x8D`/`0x8F`/`0x90`/`0x9D` into this module, the test dies with `UnicodeDecodeError` on Windows instead of reporting a finding. Those bytes are not exotic here — 43 `.py` files under `code_puppy/` already contain them (e.g. the folder glyph U+1F4C1 encodes to `f0 9f 93 81`). And because `windows-encoding` only runs `tests/test_windows_encoding.py`, CI would never catch it; it'd land on a Windows contributor's local run.

Worth noting the repo already went to some trouble over exactly this failure mode — there's a whole CI job and a long comment in `ci.yml` about Windows stdio defaulting to the legacy codepage.

The tidy fix is to not hand-roll it at all. `audit.audit_tree()` already accepts a single `.py` path, opens with explicit `encoding="utf-8"`, and raises `FileNotFoundError` loudly on a bad path:

```python
def test_agent_menu_has_no_raw_emit_sites():
    """Every literal message-bus string in agent_menu must use the catalog."""
    root = Path(__file__).resolve().parents[2]
    report = audit.audit_tree(str(root / "code_puppy" / "command_line" / "agent_menu.py"))
    assert not report.raw, [f"{s.path}:{s.line}" for s in report.raw]
```

Verified this returns `sites: 12, raw: 0, coverage: 100.0` — identical signal, three hand-rolled behaviours deleted (file reading, encoding choice, path validation), and it reuses the function `test_i18n_audit.py::test_single_file_path_is_accepted` already pins as a contract. DRY, and it inherits the fix if `audit_tree` ever changes how it reads files.

Alternatively just add `encoding="utf-8"` and move on. Either is fine; I won't hold the PR for it.

---

### Still open as a follow-up (out of scope here)

`docs/I18N.md` documents `python -m code_puppy.i18n.audit --fail-under N`, but nothing in `ci.yml` or `lefthook.yml` invokes it. This PR's per-module test is the right local fix; a repo-wide ratchet would stop the next module in the series from needing one. Worth a ticket, not a change request.

---

### Verdict

**Approving.** The completeness blocker is genuinely fixed and independently verified, and the new test proves it stays fixed. The `read_text()` encoding nit is a one-line polish item you can take or leave — grab the `audit_tree()` version if you're pushing another commit anyway.

Good, fast turnaround on the feedback, and thanks for adding the guard rather than just patching the two lines.

<sub>Reviewed by Luna — re-verified against `af26d27e`: audit coverage 12/12, guard-test discrimination on old vs new source, catalog parity and ordering, CI status, and cp1252 decode behaviour.</sub>

AndrewTilson (approved):

PR #953, opened 2026-09-18
feat(i18n): localize mcp install messages

thomwebb:
## Summary
- extract user-facing MCP install command messages into the `mcp.install.*` catalog namespace
- add Spanish and Canadian French translations for all 19 new keys

## Validation
- all `mcp.install.*` keys exist in `en-US`, `es`, and `fr-CA`
- `git diff --check` passes


code_puppy/command_line/mcp/install_command.py
```diff
@@ -138,10 +137,10 @@ def _install_from_catalog(self, server_name_or_id: str, group_id: str) -> bool:
             existing_server = find_server_id_by_name(self.manager, server_name)
             if existing_server:
                 override = emit_prompt(
-                    f"Server '{server_name}' already exists. Override it? [y/N]: "
+                    t("mcp.install.override_prompt", server=server_name)
                 )
                 if not override.lower().startswith("y"):
```
AndrewTilson, line 144: **Blocking.** This parser still only accepts `y`, but `mcp.install.override_prompt` now advertises a localized accept-key: `[s/N]` in es and `[o/N]` in fr-CA.

Verified against the real catalogs: a Spanish user who follows our own prompt and presses `s` gets `Instalación cancelada` with no explanation. Same for `o` in French. The translated flow ends up strictly worse than the untranslated one.

The sibling `mcp.install_wizard.override_prompt` deliberately keeps `[y/N]` in all three locales for exactly this reason (`wizard_utils.py:240` has the identical parser). Either match that convention, or introduce a `common.yes_key` catalog entry and parse `override.lower().startswith(t("common.yes_key"))` -- the latter fixes `wizard_utils.py` too.

code_puppy/command_line/mcp/install_command.py
```diff
@@ -194,7 +199,9 @@ def _install_from_catalog(self, server_name_or_id: str, group_id: str) -> bool:
                         if not required:
                             arg_prompt += " (optional)"
```
AndrewTilson: This hardcoded English `(optional)` gets concatenated into `arg_prompt`, which is then fed into the `{prompt}` slot of the freshly-translated `mcp.install.argument_prompt`. Net result for a French user: `Chemin (optional) :` -- half-translated, which defeats the purpose of localizing this prompt.

`mcp.install_wizard.optional_suffix` already exists and is already translated in all three locales. Worth pulling into this PR rather than deferring.

(The repo's own audit agrees -- `audit_source` on this file still returns `Counter({'extracted': 15, 'raw': 3, 'dynamic': 3})`; lines 104 and 108 are the other two, and those are fair to defer.)

tests/i18n/test_catalog_namespaces.py
```diff
@@ -59,6 +59,7 @@
     "tools.common.approval": (("tools.common.approval.",), 12),
     "renderer": (("renderer.",), 26),
     "model_factory": (("model_factory.",), 15),
+    "mcp.install": (("mcp.install.",), 19),
```
AndrewTilson: This registers `mcp.install` with the *generic* sweep only, which is the weakest tier available -- it derives interpolation params from the string under test, so it is self-referential.

I typo'd only the Spanish `no_server_found` to `'{servidor}'` and re-ran all four generic checks:

```
populated>=19 : True
all resolve   : True
pseudolocalize: True
no leftovers  : True

ACTUAL es render: No se encontró ningún servidor que coincida con '{servidor}'
```

All four green, raw placeholder on the user's screen. `tests/i18n/test_wizard_utils_i18n.py` exists precisely to close this hole -- its `_REAL_PARAMS` table supplies the *actual call-site kwarg names* per locale, and its docstring notes that self-referential checks "structurally cannot detect" a diverged placeholder name.

Suggest a `test_install_command_i18n.py` mirroring it, with `_REAL_PARAMS` for the 11 parameterized keys (`server`, `name`, `variable`, `prompt`, `description`, `error`). Bonus: that would also catch the `{server}`/`{name}`/`{variable}` naming drift from the sibling namespace.

code_puppy/i18n/locales/es.json
```diff
@@ -155,6 +155,25 @@
...
+  "mcp.install.description": "Descripción: {description}",
+  "mcp.install.environment_prompt": "  Ingresa el valor de {variable}:",
+  "mcp.install.error": "Error de instalación: {error}",
+  "mcp.install.exact_id_hint": "Usa el ID exacto del servidor: '/mcp install <server_id>'",
+  "mcp.install.failed": "No se pudo instalar: {error}",
+  "mcp.install.installing": "Instalando: {server}",
+  "mcp.install.multiple_servers": "Se encontraron varios servidores que coinciden con '{server}':",
+  "mcp.install.no_description": "No hay descripción disponible",
+  "mcp.install.no_server_found": "No se encontró ningún servidor que coincida con '{server}'",
+  "mcp.install.override_prompt": "El servidor '{server}' ya existe. ¿Sobrescribirlo? [s/N]: ",
```
AndrewTilson: `[s/N]` is the correct Spanish affordance, but `install_command.py:142` only accepts `y`, so pressing `s` silently cancels the install.

Either revert to `[y/N]` (matching `mcp.install_wizard.override_prompt`, which kept `[y/N]` in es for this exact reason) or make the parser locale-aware.

code_puppy/i18n/locales/fr-CA.json
```diff
@@ -155,6 +155,25 @@
...
+  "mcp.install.description": "Description : {description}",
+  "mcp.install.environment_prompt": "  Entrez la valeur de {variable} :",
+  "mcp.install.error": "Erreur d’installation : {error}",
+  "mcp.install.exact_id_hint": "Utilisez l’identifiant exact du serveur : « /mcp install <server_id> »",
+  "mcp.install.failed": "Échec de l’installation : {error}",
+  "mcp.install.installing": "Installation : {server}",
+  "mcp.install.multiple_servers": "Plusieurs serveurs correspondent à « {server} » :",
+  "mcp.install.no_description": "Aucune description disponible",
+  "mcp.install.no_server_found": "Aucun serveur ne correspond à « {server} »",
+  "mcp.install.override_prompt": "Le serveur « {server} » existe déjà. Le remplacer? [o/N] : ",
```
AndrewTilson: Same as the es case: `[o/N]` is advertised but `install_command.py:142` only accepts `y`, so `o` silently cancels. Note `mcp.install_wizard.override_prompt` in this same file keeps `[y/N]` deliberately.

code_puppy/i18n/locales/en-US.json
```diff
@@ -280,6 +280,25 @@
...
+  "mcp.install.description": "Description: {description}",
+  "mcp.install.environment_prompt": "  Enter value for {variable}:",
+  "mcp.install.error": "Installation error: {error}",
+  "mcp.install.exact_id_hint": "Please use the exact server ID: '/mcp install <server_id>'",
+  "mcp.install.failed": "Installation failed: {error}",
+  "mcp.install.installing": "Installing: {server}",
+  "mcp.install.multiple_servers": "Multiple servers found matching '{server}':",
+  "mcp.install.no_description": "No description available",
+  "mcp.install.no_server_found": "No server found matching '{server}'",
+  "mcp.install.override_prompt": "Server '{server}' already exists. Override it? [y/N]: ",
```
AndrewTilson: Nit on whitespace consistency within this block: this key kept its trailing space, but three sibling prompts added in the same diff dropped theirs --

```
mcp.install.custom_name_prompt  'Enter custom name for this server [{name}]:'  -> no trailing space
mcp.install.environment_prompt  '  Enter value for {variable}:'                -> no trailing space
mcp.install.argument_prompt     '{prompt}:'                                    -> no trailing space
mcp.install.override_prompt     '... Override it? [y/N]: '                     -> has trailing space
```

`main` had `f"  Enter value for {var}: "`, and every `mcp.install_wizard.*` prompt keeps the trailing space. Low impact since `emit_prompt` renders via `emit_info` and reads from a separate `>>> ` line, but it's an unintended change and inconsistent with itself. (`git diff --check` won't flag it -- it lives inside JSON string values.)

Separately: 6 of the 19 keys here are verbatim duplicates of `mcp.install_wizard.*` (`already_set`, `cancelled`, `catalog_unavailable`, `command_line_arguments`/`cmd_args_header`, `failed`/`install_failed`, `required_environment_variables`/`env_vars_header`) -- 18 translations across 3 locales to hand-sync forever. Worth hoisting to a shared `mcp.common.*` namespace, or a comment explaining why they're intentionally separate.

AndrewTilson (changes requested):
## Review: `feat(i18n): localize mcp install messages`

Solid, well-scoped extraction that follows the established pattern from #950/#951 — catalog keys are alphabetized, all 19 keys exist in all three locales, `ruff check` / `ruff format --check` pass, and the namespace is wired into the shared sweep. I verified the PR's stated validation locally and it all holds.

That said, I found **one functional bug that breaks the translated flow**, plus a test gap that structurally cannot catch it. Requesting changes on the first item; the rest are suggestions.

---

### Blocking

**1. `override_prompt` translates the accept-key, but the parser still only accepts `y` — es/fr users literally cannot say yes**

`install_command.py:142` is unchanged:

```python
if not override.lower().startswith("y"):
```

but the prompt now advertises a *localized* accept-key:

| locale | rendered prompt | user types | accepted? |
|---|---|---|---|
| en-US | `... Override it? [y/N]: ` | `y` | yes |
| es | `... ¿Sobrescribirlo? [s/N]: ` | `s` |  **silently cancels** |
| fr-CA | `... Le remplacer? [o/N] : ` | `o` |  **silently cancels** |

Verified by running the real `t()` against the real catalogs. This is worse than leaving it in English: we're instructing the user to press a key and then ignoring it. A Spanish user following our own prompt gets `Instalación cancelada` and no explanation.

Note the pre-existing `mcp.install_wizard.override_prompt` keeps `[y/N]` in **all three** locales — precisely because `wizard_utils.py:240` has the same `startswith("y")` parser. This PR diverges from that convention without changing the parser.

Two ways out, either is fine:
- **Cheap:** make es/fr use `[y/N]` to match the existing `install_wizard` convention and the actual parser.
- **Better:** add a `common.yes_key` catalog key (`y`/`s`/`o`) and parse `override.lower().startswith(t("common.yes_key"))`. Then localize the hint honestly. Worth doing once, centrally, since `wizard_utils.py` has the identical bug waiting.

Please also add a regression test asserting the affirmative token in each locale's `override_prompt` is actually accepted by the parser — otherwise this reintroduces itself the next time someone "improves" a translation.

---

### Non-blocking

**2. The new namespace registration is the weakest check in the file — I broke it and nothing failed**

`test_catalog_namespaces.py:62` only adds `mcp.install` to the generic sweep. That sweep derives interpolation params *from the string under test*, so it is self-referential and cannot detect a locale whose placeholder name diverges from the call site.

Proof — I typo'd only the Spanish translation to `'{servidor}'` and re-ran all four generic checks:

```
populated>=19 : True
all resolve   : True
pseudolocalize: True
no leftovers  : True

ACTUAL es render: No se encontró ningún servidor que coincida con '{servidor}'
```

Four green checks, raw `{servidor}` on the user's screen. This is exactly the failure mode `tests/i18n/test_wizard_utils_i18n.py` was written to prevent — its `_REAL_PARAMS` table supplies the *actual call-site kwarg names* per locale, and its docstring calls out that self-referential checks "structurally cannot detect" this. The sibling namespaces (`mcp.install_wizard`) get both layers; `mcp.install` gets only the weak one.

Suggest a `tests/i18n/test_install_command_i18n.py` mirroring the wizard file, with a `_REAL_PARAMS` map for the 11 parameterized keys (`server`, `name`, `variable`, `prompt`, `description`, `error`), asserting per-locale that the value lands and no `{` survives. That also covers finding #3 for free.

**3. Placeholder names drift from the sibling namespace for no reason**

| new key | new param | equivalent existing key | existing param |
|---|---|---|---|
| `mcp.install.installing` | `{server}` | `mcp.install_wizard.installing` | `{name}` |
| `mcp.install.override_prompt` | `{server}` | `mcp.install_wizard.override_prompt` | `{server_name}` |
| `mcp.install.environment_prompt` | `{variable}` | `mcp.install_wizard.env_var_prompt` | `{var}` |

Same concept, three different names across two namespaces. Translators work key-by-key and will mix these up; the generic sweep (see #2) won't catch it when they do. Please align on the existing names.

**4. Six of nineteen keys are verbatim duplicates of `mcp.install_wizard.*` — DRY**

Whitespace-insensitive comparison of the en-US catalog:

```
mcp.install.already_set                    == mcp.install_wizard.already_set          'Already set'
mcp.install.cancelled                      == mcp.install_wizard.cancelled            'Installation cancelled'
mcp.install.catalog_unavailable            == mcp.install_wizard.catalog_unavailable  'Server catalog not available'
mcp.install.command_line_arguments         == mcp.install_wizard.cmd_args_header      'Command Line Arguments:'
mcp.install.failed                         == mcp.install_wizard.install_failed       'Installation failed: {error}'
mcp.install.required_environment_variables == mcp.install_wizard.env_vars_header      'Required Environment Variables:'
```

That's 6 strings × 3 locales = 18 duplicated translations that must now be kept in sync by hand, forever. `install_command.py` and `wizard_utils.py` are two entry points into *the same* catalog-install flow, so these will drift. Either reuse the `mcp.install_wizard.*` keys directly, or hoist the shared six into a neutral `mcp.common.*` namespace and have both sites reference it. (If they're kept separate deliberately, a one-line comment saying why would save the next reader the same archaeology.)

**5. Trailing space stripped from three interactive prompts — cosmetic regression**

```
NEW mcp.install.custom_name_prompt  'Enter custom name for this server [{name}]:'   trailing_space=False
NEW mcp.install.environment_prompt  '  Enter value for {variable}:'                 trailing_space=False
NEW mcp.install.argument_prompt     '{prompt}:'                                     trailing_space=False
OLD mcp.install_wizard.name_prompt  'Enter name for this server [{default_name}]: ' trailing_space=True
OLD mcp.install_wizard.env_var_prompt '  Enter value for {var}: '                   trailing_space=True
```

`main` had `f"  Enter value for {var}: "` — the space is gone now. Interestingly `mcp.install.override_prompt` *kept* its trailing space, so the treatment is inconsistent within this very PR. Minor, since `emit_prompt` renders the prompt via `emit_info` and then reads from a separate `>>> ` line, but it's an unintentional visual change and it diverges from the sibling namespace. (`git diff --check` won't flag these — they're inside JSON string values, not line endings.)

**6. Three raw strings left behind in the same function**

The repo's own audit tool still reports `Counter({'extracted': 15, 'raw': 3, 'dynamic': 3})` for this file:

- `install_command.py:104` — `f"  {i + 1}. {server.display_name}{indicator_str}"` (`mcp.install_wizard.server_list_item` already exists for exactly this)
- `install_command.py:108` — `f"     ID: {server.id}"`
- `install_command.py:200` — `arg_prompt += " (optional)"` → **this one is user-visible English injected into the `{prompt}` slot of a translated string**, so a French user sees `Chemin (optional) :`. `mcp.install_wizard.optional_suffix` already exists and is already translated.

`:200` is the one I'd actually fix in this PR — it defeats the purpose of localizing `argument_prompt` at all. The other two are fair to defer.

---

### Summary

| | |
|---|---|
| **Risk** | Medium — #1 is a real functional break in the localized path this PR exists to enable |
| **Scope** | Clean, focused, conventions mostly followed |
| **Tests** | Present but the weakest available tier; demonstrably passes with a broken catalog |
| **Must fix** | #1 (override parser), and I'd argue #6's `(optional)` leak |
| **Should fix** | #2 (real-param test), #3 (param naming) |
| **Nice to have** | #4 (DRY), #5 (trailing space) |

Happy to re-review quickly once #1 is sorted. Nice work on the catalog hygiene otherwise — alphabetization, locale parity, and pseudolocale coverage were all clean on first check.

thomwebb:
Addressed the requested changes in `5a985180`:

- reused the existing `mcp.install_wizard.*` keys for shared install-flow copy instead of duplicating catalog entries
- kept the existing `[y/N]` affordance aligned with the parser
- restored trailing spaces on interactive prompts
- reused the localized `optional_suffix`
- aligned placeholder names with the wizard namespace
- added real call-site interpolation coverage for the install messages

Pinned Ruff format/lint, compilation, and diff checks pass locally.

PR #954, opened 2026-09-18
feat(i18n): localize mcp edit messages

thomwebb:
## Summary
- extract user-facing MCP edit command messages into the `mcp.edit.*` catalog namespace
- add Spanish and Canadian French translations for all 10 new keys

## Validation
- all `mcp.edit.*` keys exist in `en-US`, `es`, and `fr-CA`
- `git diff --check` passes


thomwebb:
Fixed the CI failure in follow-up commit `5e10944c`.

The shared namespace test accidentally included the `mcp.install` floor from the separate PR #953, so this branch had no matching keys and failed with `0 >= 19`. Removed that unrelated row; PR #954 now registers only its own `mcp.edit` namespace.

Pinned Ruff format/lint, compilation, and diff checks pass locally. CI is rerunning.

AndrewTilson (commented):
## Review: `feat(i18n): localize mcp edit messages`

Reviewed at `2ce9db07`..`5e10944c` (+44/-0 across 5 files). I checked this out locally, ran the namespace suite, ruff, and a few adversarial probes against the i18n runtime.

**Verdict: LGTM with suggestions.** Clean, mechanical extraction that follows the established `PUP-480` pattern exactly. One latent hazard I'd fix before merge, plus a few nits. Risk: **low**.

### What I verified

- All 10 `mcp.edit.*` keys exist in `en-US`, `es`, and `fr-CA` — zero drift, and key insertion preserves the sorted order of all three catalogs.
- `ruff check .` → `All checks passed!`; `ruff format --check .` → `666 files already formatted`.
- Re-ran the four `test_catalog_namespaces.py` checks against the `mcp.edit.` prefix standalone (the full suite needs `pydantic_ai`, which I didn't install): **populated / resolves / pseudolocalizes / no-leftover-placeholder all PASS**.
- AST-walked `edit_command.py`: all 10 `t()` call sites resolve to real catalog keys, and there are **zero orphan keys** in the catalog. No typos, no dead entries. Nice.
- Placeholder names (`{error}`, `{server}`, `{name}`) are identical across all three locales, so the pseudolocale placeholder-preservation path holds.
- Translations read correctly — `« {server} » introuvable` uses guillemets consistent with the 38 existing fr-CA strings that already do, and the Spanish is idiomatic. Good localization instincts, not just machine output.

---

### The one thing I'd fix: catalog text now flows into a Rich markup parser

`edit_command.py:45`

```python
Text.from_markup(f"[yellow]{t('mcp.edit.usage')}[/yellow]")
```

Before this PR the string inside `from_markup` was a hardcoded literal a reviewer could eyeball. Now it's **translator-supplied data being concatenated into a markup grammar** — the i18n equivalent of building SQL with an f-string. `[` is a metacharacter in Rich markup, and translators have no idea.

I reproduced both failure modes against real Rich:

```
'Uso: /mcp edit [nombre_del_servidor]'  ->  'Uso: /mcp edit '        # text silently eaten
'Uso: [/mcp edit]'                      ->  MarkupError: closing tag doesn't match any open tag
```

The current `en-US` string uses `<server_name>` so nothing breaks *today*. But the whole point of `<...>` vs `[...]` here is a convention only the English string happens to follow, and a future translator (or a `fr-CA` reviewer who prefers `[nom_du_serveur]`) turns a cosmetic edit into a silent text-drop or a crash in the error path. Nothing in the suite catches it — `test_every_key_pseudolocalizes` runs the raw string, never the `from_markup` round-trip.

Styling isn't language, so it shouldn't live in the same string as language. Suggest keeping the two concerns separated:

```python
emit_info(Text(t("mcp.edit.usage"), style="yellow"), message_group=group_id)
```

Same rendering, markup grammar no longer reachable from the catalog, and it drops an f-string. (This pattern exists elsewhere in `code_puppy/command_line/mcp/` too — `restart_command.py:98`, `handler.py:109`, `search_command.py:121` — so it's a pre-existing seam, not something you invented. Worth not *extending* it though, since this PR is the first to route catalog data through it.)

---

### Nits (non-blocking)

**1. `mcp.edit.error` is `"Error: {error}"` — the string carries no information.**

It's the only key in all 562 whose entire value is the bare word "Error". Neighbouring namespaces all went descriptive: `cli.command.error` → `"Command error: {error}"`, `cmd.compact.error` → `"/compact error: {error}"`, `oauth.callback.error` → `"OAuth callback error: {error}"`. A translator seeing `"Error: {error}"` with no context can't pick the right gender/register, and the user sees an error with no idea which subsystem produced it. Suggest `"Error editing server: {error}"` — it also matches the `logger.error(f"Error editing server: {e}")` on the line directly above it, which currently says something *more useful* than the message the user actually gets. Explicit is better than implicit.

**2. `mcp.edit.server_name` = `"  • {name}"` is formatting, not language.**

Two leading spaces and a bullet aren't translatable content — every locale got the byte-identical value, which is the tell. It's now a catalog round-trip and a translator's chance to typo an indent. `mcp.wizard.stdio.example_node` already does this, so there's precedent and I won't die on this hill — just flagging that it inflates the key count without localizing anything. (It also nudges the namespace to exactly 10, see below.)

**3. The `("mcp.edit.", 10)` floor has zero slack, so `test_namespace_is_populated` is currently a tautology.**

`len(keys) >= 10` where the catalog has exactly 10 — it can only ever fail on deletion, never catch a *partial* extraction. Consistent with `mcp.wizard` (39/39), `mcp.install_wizard` (31/31) and `model_factory` (15/15), which all sit at zero slack too, so this is the house style and I'm fine with it. Noting it so nobody mistakes that assertion for coverage of "did we get all the strings in this file."

**4. Optional hardening: `tests/command_line/mcp/test_edit_command.py` now asserts on translated output.**

Those assertions (`"No MCP servers configured" in ...`, `"not found" in ...`) used to check hardcoded literals; they now check whatever the active locale renders. The translator singleton defaults to `en-US` and is only seeded explicitly, so **this passes today** — but `tests/i18n/conftest.py` scrubs `CODE_PUPPY_LOCALE`/`LC_ALL`/`LANG` precisely because ambient locale leaking into assertions is a known hazard, and `tests/command_line/mcp/` has no equivalent guard. Confirmed the failure shape is real if the locale ever does get set:

```
CODE_PUPPY_LOCALE=es  ->  "No se encontró el servidor 'x'."  ->  'not found' in ... == False
```

Cheapest fix is asserting on the key rather than the prose (`mock_err.call_args` contains `t("mcp.edit.not_found", server="myserver")`), which also makes the test survive future copy edits. Genuinely optional.

---

### Summary

| | |
|---|---|
| **Correctness** | Pass - keys/placeholders/locales all consistent; no orphans or typos |
| **Tests** | Registered with the shared sweep, but no test covers the `from_markup` round-trip |
| **Risk** | Low — behaviour-preserving for `en-US` today |
| **Blocking?** | No. I'd take the two-line `Text(..., style=)` change and the `mcp.edit.error` wording, then ship. |

Commits are conventional and atomic, and the PR body's validation claims all check out against what I ran locally — appreciated, that made this fast to verify.

<sub> Reviewed by Luna (Code Puppy)</sub>

thomwebb:
Addressed the review suggestions in `3b590b82`:

- separated translated usage text from Rich styling with `Text(..., style="yellow")`
- made the exception message descriptive with `mcp.edit.config_error`
- updated namespace coverage for the new catalog key

Pinned Ruff format/lint, compilation, and diff checks pass locally.

AndrewTilson (commented):
## Re-review: `3b590b82` (follow-up)

Thanks for the quick turnaround. Re-checked out at `3b590b82`, diffed only `5e10944c..HEAD` (+6/-3 across the same 5 files) and re-ran the gates. Skipping everything I already verified last round unless the delta touched it.

### Resolved

**1. Rich markup injection - fixed properly.** `Text(t("mcp.edit.usage"), style="yellow")` is exactly the right shape: styling is now a parameter, not string-concatenated grammar. Re-ran both hostile payloads from my last review against the new call:

```
'Uso: /mcp edit [nombre_del_servidor]'  ->  'Uso: /mcp edit [nombre_del_servidor]'   (was: silently truncated)
'Uso: [/mcp edit]'                      ->  'Uso: [/mcp edit]'                        (was: MarkupError)
```

Both render verbatim with the style still applied. Translators can no longer break the usage line, and the last `from_markup` in this file is gone.

**2. Gates still green.** `ruff check .` -> `All checks passed!`, `ruff format --check .` -> `666 files already formatted`. All four namespace checks pass at the new floor of 11 (populated / resolves / pseudolocalizes / no-leftover-placeholder). Catalogs remain sorted, `mcp.edit.*` parity across `en-US`/`es`/`fr-CA` holds, and placeholder-name parity is clean - I diffed the `{...}` sets per key across all three locales, zero drift.

---

### New finding: `mcp.edit.error` is now a dead key, and the floor bump pins it there

The fix added `mcp.edit.config_error` instead of rewording `mcp.edit.error`, so the old key is now unreferenced. AST-walked the module to confirm:

```
catalog mcp.edit.* keys     : 11
used in source              : 10
ORPHAN (in catalog, unused) : ['mcp.edit.error']
```

`grep -rn "mcp.edit.error"` across the repo returns hits in the three locale files and **nowhere else** - no Python reference at all. Last round this namespace had zero orphans, which I called out as a strength; this reintroduces one, tripled across locales.

The part worth pausing on: bumping the floor `10 -> 11` makes `test_namespace_is_populated` assert `len(keys) >= 11`, so the suite now *requires* the dead key to stay. Delete `mcp.edit.error` and CI goes red for the wrong reason. That inverts the test's purpose - it was meant to catch a partial extraction, and it's now load-bearing for a string nothing renders. Please drop the three `mcp.edit.error` entries and leave the floor at **10** (10 real keys, 10 call sites, back to zero orphans and no net change to the count).

---

### The rename swapped one mislabel for a different one

Appreciated the intent, but I don't think `config_error` is the right name for that handler. Mapping every handler in the file:

| Scope | Handler | Key | en-US |
|---|---|---|---|
| `_load_server_config` | `json.JSONDecodeError` | `config_read_error` | `Error reading config file: {error}` |
| `_load_server_config` | `Exception` | `config_load_error` | `Error loading server config: {error}` |
| `execute` | `Exception` | **`config_error`** | **`Config error: {error}`** |

The `execute` handler is the outermost catch-all, and config loading is *already fully handled* one level down - `_load_server_config` catches its own failures and returns `None`, and the caller bails on `None` before going further. So by the time control reaches `execute`'s `except`, the realistic causes are `run_custom_server_form` blowing up or `reload_mcp_servers()` failing. Those are TUI/reload faults, not config faults.

Net effect: a form crash now tells the user `Config error: ...` and sends them to inspect a config file that is fine. That's arguably worse than the original vague `Error:` - it's confidently wrong rather than merely unhelpful. It also collides namewise with the two genuine config keys directly above it, which is a trap for the next reader.

My original suggestion was `"Error editing server: {error}"`, matching the `logger.error(f"Error editing server: {e}")` on the line immediately above - same event, same words, and it stays true whatever the cause. Suggest reusing the existing `mcp.edit.error` key with that wording rather than adding a new one, which resolves the orphan in the same stroke:

```json
"mcp.edit.error": "Error editing server: {error}"
```

```python
emit_error(t("mcp.edit.error", error=e), message_group=group_id)
```

One key, no orphan, floor stays at 10, and the label matches what actually failed. (`es`: `Error al editar el servidor: {error}` / `fr-CA`: `Erreur lors de la modification du serveur : {error}`, if useful.)

---

### Still open from last round (both optional, both unchanged)

- **`mcp.edit.server_name` = `"  • {name}"`** - still byte-identical across all three locales, which remains the tell that it's formatting rather than language. Precedent exists (`mcp.wizard.stdio.example_node`), so still not blocking.
- **`tests/command_line/mcp/test_edit_command.py` locale guard** - confirmed still absent. Those English prose assertions pass today because the translator defaults to `en-US`, but there's no `monkeypatch.delenv` equivalent to `tests/i18n/conftest.py`. Unchanged risk, unchanged low priority.

---

### Summary

| | |
|---|---|
| **Blocking fix** | Resolved - markup injection closed correctly |
| **Gates** | ruff clean, namespace suite passes at floor 11, catalogs sorted, parity intact |
| **New** | `mcp.edit.error` orphaned, and the floor bump makes CI enforce keeping it |
| **Carried** | `config_error` mislabels a non-config handler; two prior nits unchanged |
| **Risk** | Still low - cosmetic/maintainability, nothing user-breaking |

Close to done. Folding the orphan and the rename into one change - reuse `mcp.edit.error` with `"Error editing server: {error}"` and revert the floor to 10 - clears the rest of this in about four lines.

<sub>Reviewed by Luna (Code Puppy)</sub>

PR #956, opened 2026-09-19
Add plugin-provided i18n catalogs

thomwebb:
## Summary

- add an optional public `register_plugin_catalog` API for plugin-provided locale JSON catalogs
- support installed package resources through `importlib.resources` Traversable streams and loose user/project plugin filesystem directories
- derive ownership from the existing plugin loader and namespace plugin message keys by canonical plugin identity
- require an `en-US` source catalog while allowing partial translated catalogs to use the existing locale fallback chain
- validate and register each provider atomically, including deterministic collision handling, rollback, cache invalidation, and nested/concurrent plugin-load lifecycle safety
- preserve the existing `add_catalog_dir` API, precedence, and legacy language-pack/core override behavior
- keep new locales limited to message catalogs; plural rules and number/date formatting retain their existing fallbacks
- add no runtime dependencies

The lifecycle coverage includes failed-load rollback, trusted project-plugin hot-loads, nested and concurrent loading, and catalog visibility across plugin enable/disable changes.

## Validation

- full test suite: **8167 passed, 32 skipped**
- Ruff check and format check
- Python `compileall`
- source distribution and wheel builds
- isolated installed-wheel smoke test using a separate entry-point plugin wheel with packaged `en-US` and translated catalogs after fixture source removal

## Documentation

Plugin-author documentation is included with package-data examples, namespacing and source-catalog requirements, fallback behavior, diagnostics, lifecycle behavior, compatibility guidance, legacy override guidance, and the message-catalog-only limitation.

## Tracking

[PUP-843](https://jira.walmart.com/browse/PUP-843)


AndrewTilson (commented):
## Review: Add plugin-provided i18n catalogs (PUP-843)

**Verdict: Comment / approve-with-nits.** This is genuinely high-quality work — the transactional staging model, the "never hold a lock across plugin-supplied I/O" discipline, the immutable `MappingProxyType` providers, and the fail-soft public boundary are all the right calls. Test coverage is excellent (~1250 test lines for ~1000 source lines, covering rollback, nesting, concurrency, and enable/disable visibility). Docs are thorough.

**Verified locally** on `feat/plugin-i18n-catalogs` (Python 3.12):
- `tests/i18n` + `tests/test_installed_plugins.py` + `tests/test_project_plugin_failure_cleanup.py` → **273 passed**
- `ruff check` (project rule set) → clean; `ruff format --check` on all changed files → clean
- CI: `quality`, `test (macos-latest, 3.13)`, `windows-encoding` all pass

**Risk: MEDIUM-HIGH.** It touches the plugin loader lifecycle and swaps a process-global for a `ContextVar` — blast radius is every plugin, not just i18n-aware ones. The test suite mitigates this well, but the loader changes deserve the closest look from a second reviewer.

Everything below is a nit or a design question. None of it is a merge blocker.

---

### 1. DRY: collision logic is implemented twice, verbatim

`catalog.py` has two copies of "is this owner already registered / does this namespace collide," including byte-identical warning strings:

- `register_plugin_catalog` (~L196–L228)
- `_merge_pending_provider` (~L236–L258)

The only real difference is the candidate set (`_plugin_catalogs` + lineage pending vs. `_plugin_catalogs` + target). That's a parameter, not a reason to fork the function:

```python
def _accept_provider(
    provider: _PluginCatalogProvider,
    existing: Iterable[_PluginCatalogProvider],
) -> bool:
    """Return whether *provider* may join *existing* without collision."""
```

Then `register_plugin_catalog` calls it with the lineage view and `_merge_pending_provider` calls it with the target view. One copy of the rules, one copy of the diagnostics, and future rule changes can't drift between the two paths — which is exactly the failure mode that makes duplicated validation dangerous.

### 2. Layering: the plugins package pokes i18n privates through `sys.modules`

Two places do this:

```python
catalog_module = sys.modules.get("code_puppy.i18n.catalog")
if catalog_module is not None:
    catalog_module._finish_plugin_catalog_load(...)   # plugins/__init__.py
    catalog_module._plugin_state_changed()            # plugins/config.py
```

I understand the motive (don't import i18n for plugins that never touch it), and I agree with the goal. But the implementation gives the plugin loader a stringly-typed, private-attribute dependency on an i18n internal, with zero type checking and zero import-time verification. Rename `_finish_plugin_catalog_load` and nothing fails until runtime, on the failure path, silently.

Dependency inversion would be cheaper here than it looks: have `catalog.py` register a finalizer callback with the loader at import time, and let the loader call an empty list of hooks when i18n was never imported. The loader depends on "a list of lifecycle hooks," not on "a module whose name I hardcoded." Also makes this testable without `sys.modules` surgery.

If the indirection is judged not worth it, at minimum please centralize the two lookups into one `_catalog_hook()` helper so there's a single place to break.

### 3. `register_plugin_catalog` is 101 lines

It currently does: transaction lookup → resource parsing → lineage recheck → dedup → collision scan → staging. That's six responsibilities and it's double the checklist's 50-line guideline. Fixing #1 above removes the biggest chunk; extracting the staging block into `_stage_provider(transaction, provider)` would get the public entry point down to something a reader can hold in their head.

Same note, lighter, for `_load_one_project_plugin` — the nesting is now `with` → `try` → `try` → `if` → `with`, and the two `finally` blocks are far enough from their `try` that the control flow takes real effort to verify.

### 4. `_finish_plugin_catalog_load` has conditionally-bound locals

`changed` / `next_generation` are only assigned in the `parent_transaction_id is None` branch; `prospective_parent` only in the other. The later `elif changed and prospective_registry is not None` is only reachable when they *are* bound, so it's correct — but it's correct by non-local reasoning, which is the fragile kind. Static analyzers will flag it, and a future edit that changes the branch structure breaks it silently. Initializing both up front, or splitting into `_transfer_to_parent()` / `_publish_to_registry()`, would make the invariant local and obvious. *Readability counts.*

### 5. Magic number: `for _attempt in range(3)`

Why 3? The comment explains the *strategy* well but not the *constant*. Please hoist it:

```python
_MAX_OPTIMISTIC_CATALOG_BUILDS = 3
```

Self-documenting, greppable, tunable. (The repo already leans on named constants elsewhere — `_PROJECT_PLUGIN_PYCACHE`, `_WIN_RESERVED`.)

### 6. `except MemoryError: raise` appears 13 times

The intent is right and the ordering is correct (`MemoryError` *is* an `Exception`, so it genuinely needs to precede the broad catch). But 13 hand-rolled copies of a safety invariant is 13 chances to forget the 14th. Consider a single small helper — a `@fail_soft(...)` decorator or a `_soft_errors()` context manager — so the "what must always propagate" policy lives in exactly one place. *Special cases aren't special enough to break the rules* applies to the repetition, not the rule itself.

Related: `catalog._error_text()` and `plugin_catalog.resource_name()` are both "defensively stringify an untrusted object," both with the same `except MemoryError: raise` dance. They could share a home.

### 7. `_canonical_plugin_namespace` alias exists only for tests

```python
# Kept private for existing tests and diagnostics that inspect canonical owners.
_canonical_plugin_namespace = canonical_plugin_namespace
```

Nothing in `code_puppy/` uses it — only the three new test modules do, and they're new code that could just as easily `from code_puppy.i18n.plugin_catalog import canonical_plugin_namespace`. Production code shouldn't carry an alias whose sole consumer is a test suite written in the same PR. YAGNI.

### 8. `_snapshot_project_modules()` wraps one expression

```python
def _snapshot_project_modules() -> dict[str, object]:
    return dict(sys.modules)
```

One call site, zero logic, and the name promises "project modules" while returning *all* modules. Either inline it or make the name honest. (The docstring is doing real work explaining *why* — that comment could live at the call site.)

### 9. Minor

- **Typo**, `catalog.py` L59–60: `"This keeps a"` / `"a plugin that registers..."` — duplicated `a` across the line break.
- **Docs grammar**, `I18N.md`: *"Ordinary malformed/provider and filesystem/archive resource errors"* — the slashes garble the sentence. Suggest: *"Ordinary malformed-provider errors and filesystem/archive resource errors."*
- **Unrelated churn**: alphabetizing `__all__` in `i18n/__init__.py` is a fine change, but it's a 50-line reorder mixed into a feature diff. Easier to review as its own commit.
- **Commit hygiene**: 2,564 additions in a single squashed commit titled `Add plugin-provided i18n catalogs`. The atomic seams are already visible in the diff (extract `interpolation.py`; extract `plugin_catalog.py`; `ContextVar` loader lifecycle; project-plugin module cleanup; plugin catalog registration; docs). Splitting would make `git bisect` and future archaeology much kinder — the loader-lifecycle change in particular deserves to be independently revertable, since it's the piece most likely to surface a regression later.

---

### Design question (not a blocker)

The nested-transaction savepoint machinery — lineage chains, shared `RLock` inheritance, parent/child merge-on-transfer — is the most complex part of this PR, and it exists to handle a plugin importing another plugin mid-load. Is that a real scenario today, or defensive design for a hypothetical?

If it's hypothetical, *simple is better than complex* argues for rejecting nested registration outright with a clear warning, and adding savepoints when something actually needs them. If it's real (project plugins importing siblings, presumably), a one-line comment naming the concrete case would help the next reader understand why this machinery earns its keep. Either way the code is correct — this is about what we're signing up to maintain.

Also worth confirming: the `ContextVar` swap means a plugin that spawns a thread during import no longer inherits loading context (threads don't copy context), whereas the old module-global would have leaked it in. That's almost certainly the *better* behavior, but it is a silent behavior change for any existing plugin doing threaded init — worth a line in the docs' compatibility section.

---

**Bottom line:** merge-ready on correctness. Findings #1 (duplicated collision logic) and #2 (`sys.modules` private-attribute coupling) are the two I'd genuinely like addressed before merge — both are maintenance hazards rather than bugs, but both are cheap to fix now and annoying to fix later. The rest is polish. Nice work on the test suite; the rollback and concurrency coverage is what makes this reviewable at all at 2.5k lines. 

<sub>Reviewed by Luna (Code Puppy) — tests, lint, and format independently re-run against the PR branch.</sub>

thomwebb:
Thanks, Andrew, for the thorough review.

I addressed the feedback in commit [`5445e64a`](https://github.com/mpfaffenberger/code_puppy/commit/5445e64a59a743e86b2007787c95a7bb3decc386):

- deduplicated provider collision handling;
- replaced the stringly `sys.modules` lifecycle coupling with a typed, lazy bridge;
- split the long registration and project-loader paths into focused helpers;
- localized finalization invariants and added a named retry constant;
- removed the test-only production alias and minimized `__all__` churn;
- fixed wording/docs and documented nested loading plus runtime-configurable thread `ContextVar` inheritance.

I intentionally retained the explicit `MemoryError` boundaries because each protects a different fail-soft boundary, and kept `_snapshot_project_modules` as the deliberate fault-injection seam for restoration tests.

Validation is green: focused lifecycle/i18n tests **269 passed**, the full suite **8172 passed / 32 skipped**, and all current PR CI checks pass.

Would you please take another look and formally approve if you're satisfied with the cleanup? Thanks again.

AndrewTilson (approved):
## Follow-up review of `5445e64a` — all prior findings resolved

Thanks for the fast and unusually thorough turnaround, TJ. I re-verified every item rather than taking the summary at face value, and the cleanup holds up. **Approving.**

### Independent validation on `5445e64a`

- Focused lifecycle/i18n set (`tests/i18n`, `test_installed_plugins`, `test_project_plugin_failure_cleanup`, `test_plugin_i18n_lifecycle`) → **278 passed**
- Full suite → **8169 passed, 33 skipped, 2 failed**
- `ruff check` (project rule set) and `ruff format --check` on all changed files → clean
- CI: `quality`, `test (macos-latest, 3.13)`, `windows-encoding` → all green

On the 2 failures: they are `tests/mcp/test_tool_failure_not_fatal.py`, and I confirmed they **fail identically on `main` (5ac1de6)** — `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, local mcp 2.x env drift. Unrelated to this PR, no action needed here. Flagging only so the number doesn't look like a regression against your 8172.

### Finding-by-finding

| # | Finding | Status |
|---|---|---|
| 1 | Duplicated collision logic | **Fixed** — single `_accept_provider` |
| 2 | `sys.modules` private coupling | **Fixed** — typed `Protocol` bridge |
| 3 | 101-line `register_plugin_catalog` | **Fixed** — now ~28 lines |
| 4 | Conditionally-bound locals | **Fixed** — early `return` |
| 5 | `range(3)` magic number | **Fixed** — named constant |
| 6 | 13x `except MemoryError` | **Declined, accepted** |
| 7 | Test-only production alias | **Fixed** — removed |
| 8 | `_snapshot_project_modules` | **Declined, accepted** |
| 9 | Typo / docs / `__all__` churn | **Fixed** |

**#1** — `_accept_provider` is now the single source of truth for both owner-dedup and namespace-collision rules, including the warning strings. Both call sites route through it. This was my main concern and it's properly resolved.

**#2** — `_plugin_i18n_lifecycle.py` is a genuinely nice piece of design. A `Protocol`-typed `_CatalogLoadFinalizer`, a frozen `_I18nLifecycleHooks` dataclass, and no-op adapters when i18n was never imported. The loader now depends on an interface rather than a hardcoded module string, and the laziness goal is preserved. Better than what I suggested — and `test_plugin_without_i18n_does_not_import_i18n_stack` asserts the laziness property in a subprocess, which is exactly the right way to test "this module must not be imported." Nice touch.

**#4** — The parent-transfer branch now returns early instead of falling through to a shared tail that depended on conditionally-bound `changed` / `next_generation`. The invariant is local and obvious now; the `except BaseException` savepoint-discard semantics are unchanged. Correct.

### On the two you pushed back on — you're right, and I verified why

**#6 (`MemoryError` boundaries).** I checked the call sites and withdraw the suggestion. They guard genuinely different boundaries — `_error_text`, `resource_name`, `read_plugin_catalogs`, `_module_origin_is_under`, the loader paths — and several sit inside `except` clauses where a decorator or context manager would have to re-raise through an active exception context. Collapsing them would have traded 13 explicit, locally-auditable guards for one clever abstraction with subtler semantics. *Explicit is better than implicit* wins here. The new comment in `plugins/__init__.py` documenting the policy is the right resolution.

**#8 (`_snapshot_project_modules`).** Also correct, and I should have grepped before flagging it. `tests/test_project_plugin_failure_cleanup.py:107` monkeypatches it via `raising=False` to inject an exhausted-snapshot fault — the seam is load-bearing, and inlining it would delete a test's only injection point. The updated docstring ("intentional failure-injection seam") now says so, which is what I actually wanted: the *why* written down. Same reasoning applies to the `_finish_plugin_catalog_load_hook` / `_plugin_state_changed_hook` wrappers — I checked, and tests at `test_plugin_catalog_hardening.py:193` and `test_plugin_catalog_transactions.py:108` monkeypatch the underlying functions by name, so the late-binding indirection is required for that fault injection to reach the loader. Worth the one-line docstrings you gave them.

### Bonus: the threading docs

I flagged the `ContextVar` thread-inheritance behavior change as a passing note and you turned it into a properly precise docs paragraph — naming Python 3.14's `thread_inherit_context`, the free-threaded default, and the key invariant that an inheriting thread *still* gets rejected if it completes after loader exit. That last part is the subtle bit most people would have missed. The nested-loading doc now also names the concrete scenario (plugin code synchronously re-entering the loader), which answers my design question: the savepoint machinery earns its keep. Good.

### One optional nit for a future PR (not blocking)

`_accept_provider`'s third parameter is a little awkward:

```python
accepted, already_present = _accept_provider(
    provider,
    (*target.values(), *_plugin_catalogs.values()),   # owner_candidates
    (*_plugin_catalogs.values(), *target.values()),   # namespace_candidates
)
```

Same two sets, different order, and the ordering is load-bearing — `owner_candidates` prefers `target` so a child savepoint shadows the published registry, while `namespace_candidates` prefers `_plugin_catalogs` so collision diagnostics name the already-published plugin. That's a real distinction and the code is correct, but it's currently only discoverable by reading `_accept_provider` and reasoning backwards. A one-line comment at that call site explaining why the orders differ would save the next reader the trip. The `Iterable | None` defaulting is fine as-is.

---

Everything I raised is either fixed or has a reasoned justification I agree with. CI is green, the full suite is clean modulo a pre-existing environment failure, and the refactor didn't change behavior — the same 278 focused tests pass before and after. Thanks for engaging with the pedantry rather than just pattern-matching it away; pushing back on #6 and #8 with actual reasons was the right call.

Ship it.

<sub>Reviewed by Luna (Code Puppy) — tests, lint, and format independently re-run against `5445e64a`; the 2 MCP failures independently reproduced on `main`.</sub>

